PR #1, opened 2020-07-19
Printer power plugin

jordanruthe:
Adding plugin to control printer power via shell scripts. This would allow moonraker to turn on/off the printer rather than manually doing it through the cli. Implementing this into moonraker rather than klipper since if a MCU is powered down, due to a power off, klipper will refuse to issues any commands such as M80/M81 which would allow the MCU to turn back on if controlled by hardware.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -174,8 +174,9 @@ batch subscription updates together.
 
 ## Plugin Configuration
 The core plugins are configured via the primary configuration above.  Optional
-plugins each need their own configuration.  Currently the only optional plugin
-available is the `paneldue` plugin, which can be configured as follows:
+plugins each need their own configuration as outlined below.
+
+### PanelDue Plugin
 
 ```
 [moonraker_plugin paneldue]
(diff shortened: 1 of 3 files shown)


lixxbox:
Hi!
Could you later add an "auto power off" function?  A timeout that turns off the relay when the klipper is idle and no interaction via moonraker takes place?

I am looking forward to the plugin and the possibility to use it via mainsail.

Arksine:
Hi.  I think the concept of managing "printer power" from Moonraker makes sense.  I do have some high level comments.

I haven't yet had a chance to add the contributing guidelines to the repo, I'll get to that as soon as I am able.   The tentative plan for Moonraker is to eventually merge it into the Klipper  repositiory, after which this repo will be archived.  Thus contributions to Moonraker will have the same guidelines as Klipper:
- The start of the source code needs a copyright notice
- Each commit needs to be signed off
We can move the configuration documentation to the bottom of installation.md until I have a better place for it.

As for the code itself, it looks good, however I can't merge the configurable shell scripts.  The scripts that toggle GPIOs will need to be hardcoded.  We can do that in the source itself by writing to a script to a temporary file, or we can add scripts to the "scripts" folder that toggle/query a supplied GPIO.

Thanks.

jordanruthe:
Thanks for the contribution guidelines.

I was trying to avoid tying it to RPi GPIO for people who may not run this setup on a RPi and keep the flexibility to do more than just toggle a pin (such as restart klipper, turn on/off a display), but just swapped it an have it working on my system.

As for the timeout feature, I'll look into that as that's also something I would like.

Arksine:
Thanks.  I understand the flexibility point of view.   I'm okay with using RPi.GPIO for now, however in the future I can write a small plugin that toggles GPIOs using sysfs.   That way GPIO functionality would not be tied to Raspberry Pi hardware.

If we need to add more customization aside from toggling GPIOs its possible to add support for Jinja2 templates.  We would just need to define what kind of functionality we need to make available.  I think that is something to hold off on unless it is needed though.

When you are ready for me to merge this let me know, I just need you to sign off on your last commit.  It is up to you if you want me to do a squash merge or if you want to rebase yourself.

jordanruthe:
Good point, I just modified it so RPi.GPIO isn't a dependency. I'm good with it if you are, go ahead and squash merge it.

Arksine:
Thanks.

lixxbox:
would it be a lot of work to implement mqtt? so that you can turn sonoff/tasmota devices on and off, for example.
this would give us back some flexibility without allowing direct access to the system.

PR #27, opened 2020-10-17
Adding timeout feature for power devices

jordanruthe:
Signed-off-by: Jordan Ruthe <jordan.ruthe@gmail.com>

### DIFF
--- a/docs/api_changes.md
+++ b/docs/api_changes.md
@@ -1,6 +1,21 @@
 This document keeps a record of all changes to Moonraker's remote
 facing APIs.
 
+### October 25th 2020
+- The `modified` field reported for files and directories is no
+  longer represented as a string.  It is now a floating point
+  value representing unix time (in seconds).  This can be used
+  to display the "last modified date" based on the client's
+  timezone.
+
+### October 21st 2020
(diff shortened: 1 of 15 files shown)


Arksine:
Hi Jordan, sorry for the delay.  Before I commented I wanted to complete a PR for Klipper that I think may be useful in implementing the functionality you seek.

With regard to the idle timeout, I think it would be better to use Klipper's [idle_timeout] functionality to do this.  That way users can customize it however they please.  I just PRed the ability to register client methods with Klipper's template context.  If that PR is merged it will be possible to toggle a power gpio from a `gcode_macro` (or from the idle timeout gcode).  If that PR is rejected then we can implement the idle timeout in moonraker.

The last two commits look good to me.  I just need you to add your [email address](https://github.com/Arksine/moonraker/blob/master/docs/contributing.md) to the "Signed-off-by" line.  For example, my signed off line looks like this:\
`Signed-off-by:  Eric Callahan <arksine.code@gmail.com>`

If you would prefer you can just edit the pull request message instead of each individual commit.

Arksine:
Since the timeout functionality can now be handled via gcode I'm going to go ahead and close this PR.  Thanks.

PR #33, opened 2020-11-01
Code style

expeditioneer:
Hi these are minor improvements, would be glad when they are accepted.

### DIFF
--- a/moonraker/app.py
+++ b/moonraker/app.py
@@ -12,6 +12,7 @@ from inspect import isclass
 import tornado
 from tornado.escape import url_unescape
 from tornado.routing import Rule, PathMatches, AnyMatches
+from tornado.web import HTTPError
 
 from authorization import Authorization
 from authorization import AuthorizedRequestHandler, AuthorizedFileHandler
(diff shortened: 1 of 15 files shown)


Arksine:
Thanks.  While I appreciate the effort, I won't merge changes to code style.  As programmers have different style preferences, committing such changes inevitably leads to code churn. As long as code is readable all I require is that each line be limited to 80 characters and do not contain trailing whitespace.

Aulos:
Maybe black?

trevjonez:
@Arksine you might consider adding in an editorconfig file to help keep some of the basic concerns automated.
https://editorconfig.org/

```.editorconfig
root = true

[*]
indent_style = space
indent_size = 4
insert_final_newline = true
trim_trailing_whitespace = true
end_of_line = lf
charset = utf-8

[*.py]
max_line_length = 80
```

Arksine:
Thanks @trevjonez.  The editorconfig file is a good idea, I will add one to the docs directory.

Arksine:
I went ahead and added the file to the root directory in the repo.  I removed the `end_of_line` option,as I think it would be preferable to use native line endings and allow git to handle the conversion.

PR #37, opened 2020-11-19
power:  use libgpiod, add support for tplink-smartplug

Arksine:
This is a significant refactor of the power plugin two accomplish two goals:
- Switch from sysfs to libgpiod gpio management
- Implement native support for tplink smartplug devices

The first solves several potential issues that lingered with sysfs managment.  Using libgpoid gives moonraker exclusive access to a gpio and it automatically releases the line if moonraker restarts or stops.  When a gpio is set using libgpoid it is no additional query is necessary to determine if the operation was successful.  This makes behavior consistent, however users should be aware that restarting moonraker will likely result in connected devices powering off.

This change has introduced additional dependencies that require Moonraker be reinstalled and the virtualenv be rebuilt.  As such, new options have been added to `install-moonraker.sh`. This may be of interest to you @th33xitus:
- `-r`   Rebuild the python virtual env\
- `-f`   Force an overwrite of `/etc/default/moonraker` during installation\
- `-c /path/to/moonraker.conf`    Allows user to specify the path to moonraker.conf during configuration.  Using this in conjunction with `-f` will update the defaults file wih the new path.

Client developers should be aware that the path for the APIs has changed from `gpio_power` to `device_power`.  The behavior of the APIs has changed slightly, as documented in [api_changes.md](https://github.com/Arksine/moonraker/blob/7adedf64597311e45f8554d0dd2498fd68ce3502/docs/api_changes.md).

@jordanruthe @Aulos If you have any custom functionality you may want to take a look at this as well.  Some specific things to keep an eye on:
- Custom Device Implementations should be modeled after the GpioDevice class, the should specifically with the following methods.
  -  `get_name()` - must return the device name.  Cannot be async.
  -  `get_device_info()`  - must return a dict with `device`, `status`, and `type` fields.  Cannot be async
  - `initialize()` - If not necessary pass off the function. If this call fails the device status should be set to "error".
  - `refresh_status()` - If not necessary pass off the function. If this call fails the device status should be set to "error".
  - `set_power()` - must toggle device power.  If this call fails the device status should be set to "error".
- The `initialize()`, `refesh_status()` , and `set_power()` methods may be defined as async, but it is not required.
- The  `refresh_status()` method is not called if `set_power()` has been called.  The implementation of `set_power()` must update the device state after toggling power.

I plan to merge this rather soon, however I tagged interested parties so they could see this and be aware of it.  If anyone finds an issue after merging please feel free to create a PR or let me know about it.  Thanks.

Signed-off-by: Eric Callahan <arksine.code@gmail.com>

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -47,18 +47,24 @@ cd ~
 git clone https://github.com/Arksine/moonraker.git
 ```
 
-If you have an experimental verison of moonraker that pre-dates this repo,
-it must be uninstalled:
-```
-cd ~/moonraker/scripts
-./uninstall-moonraker.sh
-```
-
 Finally, run moonraker's install script:
 ```
 cd ~/moonraker/scripts
 ./install-moonraker.sh
 ```
+
+The install script has a few command line options that may be useful,
+particularly for those upgrading:
+- -r\
+  This will rebuild the virtual environment for existing installations.
+  Sometimes this is necessary when a dependency has been added.
+- -f\
+  This will tell the script to overwrite Moonraker's "defaults" file.
+  By default the script will not modify the "defaults" file if it is
+  detected as present.
+- -c /path/to/moonraker.conf\
+  This allows the user to specify the path to Moonraker's config file.
+  The default location is "/home/<user>/moonraker.conf".
 
 When the script completes it should start both Moonraker and Klipper. In
 `klippy.log` you should find the following entry:\
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -262,33 +262,44 @@ gcode:
 Power Plugin Configuration.  One may use this module to toggle the
 state of a relay using a linux GPIO, enabling the ability to power
 a printer on/off regardless of Klippy's state.  GPIOs are toggled
-using linux sysfs.
+using libgpiod.  A configuration section should be added for each
+device as shown below:
 ```
-[power]
-devices: printer, led
-#   A comma separated list of devices you wish to control. Device names may not
-#   contain whitespace.  This parameter must be provided.
-#
-# Each device specified in "devices" should define its own set of the below
-# options:
-{dev}_name: Friendly Name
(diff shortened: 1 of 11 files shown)


moonraker/plugins/power.py
```diff
@@ -6,94 +6,84 @@
...
+        prefix_sections = config.get_prefix_sections("power")
+        logging.info(f"Power plugin loading devices: {prefix_sections}")
+        for section in prefix_sections:
+            cfg = config[section]
+            dev_type = cfg.get("type")
+            if dev_type == "gpio":
+                dev = GpioDevice(cfg, self.chip_factory)
+            elif dev_type == "tplink_smartplug":
+                dev = TPLinkSmartPlug(cfg)
+            else:
```
Aulos, line 48: I'd suggest
```
elif dev_type is None:
  continue
```

or  few lines above:

```
if "type" not in cfg:
  continue
```

I need to load power plugin, to later add my devices, but I don't want to add dummy gpio device, just so plagin will initialize and won't raise an error.

Arksine: I don't believe that will be necessary.  Presumably you will be calling `server.load_plugin('power', config)` when you init your module.  If you aren't adding a [power xxx] section to your config, then `config.get_prefix_sections()` will evaluate to an empty list and the for loop wont be entered.

docs/installation.md
```diff
@@ -256,33 +262,44 @@ gcode:
...
+```
+Below are some potential examples:
+```
+[power printer]
+type: gpio
+pin: gpio26
+
+[power printer_led]
+type: gpio
+pin !gpiochip0/gpio16
```
jordanruthe, line 297: I believe you're missing a : here.

Arksine: Good catch, thanks.

moonraker/plugins/power.py
Aulos: ok, that worked

Issue #23, opened 2020-10-05
Hostreboot don't work


anjoku:
Hello, (please excuse my bad English)

I installed Mainsail on an Orangepi Lite. I used the script from https://github.com/th33xitus/kiauh.
It works so far, only the HOSTREBOOT buttons don't work for me.

The Button HOSTSHUTDOWN work fine.

Where can I set what happens when these buttons are pressed?
Example: HOSTREBOOT should execute "sudo shutdown -h now".

Thank you Andre

[moonraker.log](https://github.com/Arksine/moonraker/files/5327951/moonraker.log)
[klippy.log](https://github.com/Arksine/moonraker/files/5327957/klippy.log)

Arksine:
Thanks.  It sounds like your distro does not support the `reboot` command.  I went ahead and pushed an update that executes `sudo shutdown -r now`, which should hopefully work with more distros.

anjoku:
Hello, thank you.
I just tried it.

"sudo reboot "

works on the orangePi

Greetings
Andre

KwadFan:
Hello @anjoku , 
I discovered this Issue like you.
It seems that the configuration of choosen Modules of MainsailOS and even FluiddPI create this Issue

`export MODULES="base(network, raspicam, klipper_moonraker, canserial, fluidd, mjpgstreamer, password-for-sudo)"`

The Module "password-for-sudo" parses the /etc/sudoers and /etc/sudoers.d to comment out the lines that user pi can "sudo" without any password.
That interferes with the "simple" sudo shutdown -h now or that the API in the back uses.

I am working on a fix for that, till then simply use:
`echo 'pi ALL=(ALL) NOPASSWD: /sbin/shutdown, /sbin/reboot'| sudo EDITOR='tee -a' visudo` 

This will allow pi to use shutdown and reboot command without a need for password. 

Regards
Stephan aka Kwadfan :)

Arksine:
Given that this is a disto problem and not a moonraker issue I'm going to close this.  The sudo_fix.sh script is there for CustomPiOS based distros.  Folks using other distros will need to make sure that Moonraker has the proper permissions to execute sudo commands.

PR #38, opened 2020-11-19
Added a reboot/shutdown Workaround

KwadFan:
This should prevent Issues with shutdown and reboot, even with the "password-for-sudo" Module of CustomPiOS.
Related to #23

### DIFF
--- a/scripts/install-moonraker.sh
+++ b/scripts/install-moonraker.sh
@@ -92,6 +92,12 @@ verify_ready()
     fi
 }
 
+# Allow user 'pi' to reboot and shutdown without password
+shutdown_sudo_fix()
+{
+    echo 'pi ALL=(ALL) NOPASSWD: /sbin/shutdown, /sbin/reboot'| sudo EDITOR='tee -a' visudo
+}
+
 # Force script to exit if an error occurs
 set -e
 
@@ -105,4 +111,5 @@ install_packages
 create_virtualenv
 install_script
 install_config
+shutdown_sudo_fix
 start_software


cadriel:
Is this why current CustomPiOS images with `password-for-sudo` currently fail when issue'ing the host reboot or shutdown commands?

KwadFan:
Yes, exactly. But, if you dont use the "password-for-sudo" you can sudo any command without a password. In my Opinion a huge Security Flaw. So, better nopasswd Option for reboot. Looking for a way better Solution like "power" group in full blown Distros.

Arksine:
Thanks.  I'm not adverse to adding this workaround to the repo, however I think that it should be a separate script rather than part of the installation script.  I think that its possible that this issue could affect the forthcoming `update_manager` plugin I am working on as well.  It needs to be able to run `sudo systemctl` and `sudo apt-get` commands.   Also I wonder if we can detect the user rather than hardcode it.  

Separately, each commit needs a signed-off line as detailed in [contributing.md](https://github.com/Arksine/moonraker/blob/master/docs/contributing.md).

KwadFan:
This is what I also thought the last couple of days. the update handling og Mainsail and FluiddPi need both some attention in view of Linux administration. The Problem is, that those who want to design 3D Parts and print them arent really happy dealing with
`apt-get` or `systemctl` commands to type in to an Interface from like coming from 1984. This whole process should be integrated in the Web Interface, instead of fideling around with - for the average user - messy Terminal based commands. Like a Network router for example. So, bring the latest and greatest, I want and i will contribute. By the Way, coming from Marlin getting frustrated by compiling every Crap a new Version, seen 1 Day Klipper Ecosystem and was blown through the roof. Great Job, Thanks ;)

PR #39, opened 2020-11-21
Tasmota power plugin

0x4d4e:
Based on the TPLinkSmartPlug, I added a power plugin for Tasmota switches. Switching the plug from fluidd works with the power macros from the documentation and it works in fluidd's Power menu to toggle the device.

Since this is the first time I'm working with Tornado/async, I'm happy to get feedback. Also, there is a bit of code duplication, this could probably be improved if additional power plugins are added in the future as well.

Signed-off-by:  Matthias Neumayr matt.neumayr@gmail.com

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -292,6 +292,9 @@ output_id:
 #   Provide a password if configured in Tasmota (default is empty). 
 #   Provided an output_id (relay id) if the Tasmota device supports
 #   more than one (default is 1).
+#   If your single-relay Tasmota device switches on/off successfully,
+#   but fails to report its state, ensure that 'SetOption26' is set in 
+#   Tasmota.
 
 ```
 Below are some potential examples:
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -265,10 +265,10 @@ a printer on/off regardless of Klippy's state.  GPIOs are toggled
 using libgpiod.  A configuration section should be added for each
 device as shown below:
 ```
-[power device name]
+[power device_name]
 type: gpio
-#   The type of device.  Can be either gpio or tplink_smartplug.  This
-#   parameter must be provided.
+#   The type of device.  Can be either gpio, tplink_smartplug or tasmota.
+#   This parameter must be provided.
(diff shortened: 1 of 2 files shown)


Arksine:
Thanks.  The implementation looks good to me.  Could you add your signed-off line to each commit as well?  Thanks.

0x4d4e:
Do you have a pointer on how to change past commit messages? Never had to do that :) 

One more thing: I noticed, that another Tasmota device returned a slightly different response ({"POWER": ...} instead of the expected {"POWER1": ...}). (see also here: https://github.com/arendst/Tasmota/issues/1055)

The fix is to simply run "SetOption26 1" in the Tasmota web console. Should I add this to the moonraker documentation or catch this case in the code? Apparently, it only appears in this legacy response mode (SetOption26 is unset/off) and for single-relay devices (so only if output_id == 1).

Arksine:
You can do an interactive rebase to reword the commits.  I'm not sure if you are on Windows or Linux, but the general idea should be the same.  Navigate to the repo and make sure you are on your `power_tasmota` branch.  In a console, enter the following command:
```
git rebase -i HEAD~3
```
This should open up a text editor with something that looks like this:
image
Change "pick" to "r" for the commits you want to reword, save the file and close it:
image

Now the text editor will open for the first commit you selected.  You can add your signed-off line, save the file, and close it:
image
A new editor will open for each commit.
Interactive rebase is a powerful tool.  It can drop commits, squash commits together, move them around, edit them, etc.  If you notice that you accidentally messed something up you can reset the local repo to the remote branch and redo it.

When everything looks good you need to force push it to the remote.  Once this is done there is no going back:
```
git push --force
```

With regard to legacy mode, I think that documenting the potential issue is ok.

0x4d4e:
Oh wow, thanks for the description! I have known about rebase, but never used it. Worked like a charm!

I have edited all commits to include the signoff and added a hint regarding the legacy mode in the installation documentation.

Arksine:
Excellent.  Thanks.

PR #40, opened 2020-11-28
moonraker-sudo (mnrkrsudo)

KwadFan:
module name: moonraker-sudo (mnrkrsudo)

 Provides a specified Group that is intended to elevate user privileges
 to help moonraker with sudo actions, if in CustomPIOS Images with
 Module password-for-sudo.

Signed-off-by: KwadFan <me@stephanwe.de>

Related to #23


Arksine:
Thanks, this looks good to me and I will merge shortly.  With regard to #41, since its related to this, I think it would be easy enough to simply update this script if more commands are added which require sudo.

Arksine:
Ah, I need one thing before I can merge.  The signed off line needs to contain your real name.  Thanks.

Arksine:
Thanks.

Issue #46, opened 2020-12-05
API: /machine/update/system


KwadFan:
After testing the new [update_manager] feature ( Which was for me unclear if it has to set in Moonraker.conf or printer.cfg ),
I discovered that "DEBIAN_FRONTEND=noninteractive" isn't allowed to set.
There for, i will update the sudo_fix.sh, but it has a major issue. It HAVE to be a part of the installation
becaus you have to to update the /etc/environment file, to set DEBIAN_FRONTEND as a fixed System Variable.

Tested on latest FluiddPI Release.
After manual changes it works perfectly.

Who struggels with that Issue can do the following till update:

`echo DEBIAN_FRONTEND=noninteractive >> /etc/environment`

Update your 020-sudo-for-moonraker ( or create file according to your machine ) and append line:

`Defaults!/usr/bin/apt-get env_keep +="DEBIAN_FRONTEND"`

Please make sure you issue an Reboot!

Arksine:
Thanks.  I'll clarify the documentation regarding the config.  I'm unsure at the moment as to why `/etc/environment` needs to be written in the install script, however we can discuss it when you submit a PR.  I appreciate your work with this, as I expect in the future most users will prefer to use prebuilt distros rather than vanilla Raspberry Pi OS.

KwadFan:
I am working on that too, I am currently fiddling aroun with customPiOS to generate a Distro tailored to Klipper/Moonraker with some comfy Things to choose a Frontend. Indeed this Distro will not need sudo_fix :)
Further more I want completly get rid off "ssh-ing" into the Pi. I only have to understand how to abuse "shell_command.py" and writing a tornado based backend :)
Time will tell...

KwadFan:
Fix is merged, so, lets close this :)

PR #47, opened 2020-12-05
According to #46

KwadFan:
That is not the best solution to force users to use some sudo commands without a password,
but keeps working moonraker flawless with that changes.


KwadFan:
So, echoing into `/etc/environment` is the simplest solution to serve the Variable across all Distros and user combinations. A better way would modifying the `/etc/skel` related files but they can be differ in many ways, so that it would take much more patience through functions and parsing existing configurations. In that prospective, applying system wide is K.I.S.S prinziple

Arksine:
Thanks.  I'm okay with the changes to sudo_fix.sh and will merge them.  My opinion on the install script remains the same, the fix only applies to CustomPiOS based distros so the fix shouldn't be part of the generic install script.   Users who are running CustomPiOS derived distros should run this script separately.

KwadFan:
Ok, i will change that. Unfortunatly due lack of time, will be tomorrow...

Arksine:
Thanks.  I merged the change to sudo_fix.sh, so you don't need to take any additional action unless you need to further change the sudo_fix script.

cadriel:
@KwadFan

How would you expect the custompios build to introduce the sudo change? Should the moonraker module simply run the new install script, or would a specific new module need to be introduced for this?

KwadFan:
Sorry for being late with the answer. Its a busy time in my main Job so close bevor Christmas :) I think, you could just run sudo_fix.sh in the installation module or in fluiddpi module. Second Method would be my prferred. Put the "cat part" directly into a file and copy over into filesystem section of fluiddpi module, add pi to needed mnrkrsudo group via start_script. If you want I can add a pull request for FluiddPi at Weekend with the changes. I am also sorry for not using the right tags and terms its a mess with german keyboard on Mobilephone :)

KwadFan:
I think we can close this, due your merge. Have nice Christmas Holidays :)

PR #48, opened 2020-12-06
Changed cura thumbnail parsing to read from gcode

Gravedigger7789:
I added a post processing script in cura to create thumbnails similar to prusaslicer (https://github.com/Ultimaker/Cura/pull/8890)

I wasn't sure if you wanted to keep backwards compatibility with reading a *.png file. If you do let me know and I can adjust the code accordingly

I will mark this as a draft PR until the cura PR is merged

Screenshot 2020-12-06 134342

### DIFF
--- a/scripts/extract_metadata.py
+++ b/scripts/extract_metadata.py
@@ -302,34 +302,52 @@ class Cura(BaseSlicer):
             os.path.basename(self.path))[0] + ".png"
         thumbPath = os.path.join(
             os.path.dirname(self.path), "thumbs", thumbName)
-        if not os.path.isfile(thumbPath):
-            return None
-        # read file
+        thumb_matches = re.findall(
+            r"; thumbnail begin[;/\+=\w\s]+?; thumbnail end", self.header_data)
         thumbs = []
-        try:
-            with open(thumbPath, 'rb') as thumbFile:
-                fbytes = thumbFile.read()
-                with Image.open(io.BytesIO(fbytes)) as im:
-                    thumbFull = base64.b64encode(fbytes).decode()
-                    thumbs.append({
-                        'width': im.width, 'height': im.height,
-                        'size': len(thumbFull), 'data': thumbFull
-                    })
-                    # Create 32x32 thumbnail
-                    im.thumbnail((32, 32), Image.ANTIALIAS)
-                    tmpThumb = io.BytesIO()
-                    im.save(tmpThumb, format="PNG")
-                    thumbSmall = base64.b64encode(
-                        tmpThumb.getbuffer()).decode()
-                    tmpThumb.close()
-                    thumbs.insert(0, {
-                        'width': im.width, 'height': im.height,
-                        'size': len(thumbSmall), 'data': thumbSmall
-                    })
-        except Exception as e:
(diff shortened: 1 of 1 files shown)


Arksine:
Should Ultimaker merge your patch I'm ok with adding support for it in Moonraker.  I do want to keep UFP compatibility, it wouldn't be appropriate to break it.

For future reference, all commits [need to be signed off](https://github.com/Arksine/moonraker/blob/master/docs/contributing.md).  Thanks.

Gravedigger7789:
Ah! I completely missed the contributing guidelines, I'll fix it up and put the support back in for ufp 😁

Gravedigger7789:
I fixed the commit message to follow the contributing guidelines and updated the logic to support both the UFP and the gcode thumbnails.

I am not sure if the upstream changes will get merged by ultimaker but I will keep an eye out and update this accordingly if it does.

scripts/extract_metadata.py
```diff
@@ -302,34 +302,52 @@ def parse_thumbnails(self):
...
+                with open(thumbPath, 'rb') as thumbFile:
+                    fbytes = thumbFile.read()
+                    with Image.open(io.BytesIO(fbytes)) as im:
+                        thumbFull = base64.b64encode(fbytes).decode()
+                        thumbs.append({
+                            'width': im.width, 'height': im.height,
+                            'size': len(thumbFull), 'data': thumbFull
+                        })
+                        # Create 32x32 thumbnail
+                        im.thumbnail((32, 32), Image.ANTIALIAS)
```
cadriel, line 320: Is there a reason these are hard coded?

32x32, on a retina display - effectively means 16x16 if you want to keep clarity.

If there's a true requirement to hard code dimensions here, then maybe a better approach would be to create 64x64 - or some variation of.

Gravedigger7789: I'm sorry for responding so late! I didn't see this. The reason this is here is for backwards compatibility with the original code. I wasn't sure why these numbers in particular were chosen but I didn't want to modify existing implementation.
Everything from ```if thumb_matches:``` and below is the new code

Gravedigger7789:
The support from this was recently accepted into Cura so I am changing this to ready for review

Arksine:
Commit dd81bea should resolve this.  The patch was just a few lines of code, so rather than have you change the PR I decided to submit it myself.  It would be good if you could test it against the thumbnails Cura generates, although if the format is identical to PrusaSlicer it should work.

scripts/extract_metadata.py
Arksine: It is hardcoded because I can be sure it is functional.  32x32 is simple the size for the smaller thumbnail, the idea was to be compatible with the Prusa defaults and generate a thumbnail that works well with low resolution screens.  I think if a client want 64x64 they could simply scale the 300x300 thumbnail to whatever size they need.

cadriel: Is this not configurable user side? Does it just generate 32x32 and a 300x300?

Gravedigger7789:
@Arksine That commit works for me! I will close out this PR. Thank you very much for implementing this so quickly!
image

PR #55, opened 2020-12-28
Add shebang to moonraker.py

MagicRB:


Arksine:
Thanks.  To merge your commit needs to be signed off.  Please see [contributing.md](https://github.com/Arksine/moonraker/blob/master/docs/contributing.md) for details.

MagicRB:
Oh, ok. I'll close this then, I really don't want to give out my email address unless absolutely necessary, which this change definitely isn't. You can just commit it yourself. I'm really sorry for the inconvenience, I should have read that .md prior to opening this PR... If I have anything actually substantial I'll make an exception to that email address rule :) While we are talking about substantial, would you accept a PR with a `flake.nix` file for NixOS? I'm currently working on one for my own purposes, but I'll open a PR with my email if if comes to that.

Arksine:
I went ahead and added the shebang and made moonraker.py executable.

I'm not familiar with NixOS, is it correct to say that flake.nix allows you to specify moonraker dependencies?  Is this a file that could exist in the `scripts` directory?  Generally speaking I don't have a problem adding this, as I suspect moonraker will end up having more install scripts for other distros in the future as well(or I will end up making `install_moonraker.sh` configurable for other distros).

MagicRB:
the `flake.nix` specifies the whole build and packaging process. First it describes how to build the actual package, then it creates a configuration interface
```services.moonraker = {
  user = root;
  instances = {
    ender3 = {
      config = {
        port = 8989;
      };  
    } ;
  };
};```
That creates a configuration file and a systemd service.

Oh and the `flake.nix` would have to be in the root of the repo, since you can't as of now specify sub-folders I think, I'll double check that.

Arksine:
Thanks for the explanation.  Is the process packaging process automated?  As in, do you supply the package manager with Moonraker's github URL, then it clones the repo, finds flake.nix, then builds the package?

MagicRB:
Yeah, sort of, flakes are in a weird state where they're functional but not the default. But it can work with "legacy" nix packages, you just need a compatibility "layer"

Issue #66, opened 2021-01-11
FR: Add Klipper's PWM value to /server/temperature_store


CapnBry:
When I warm up my printer, I like to wait long enough for the heated bed to soak up to temperature. It takes a while for a thick MIC6 plate to actually stabilize at print temperature, long after the thermistor reports it first gets there. To do this I usually tail the klippy.log with a python regex and watch the "pwm=" in the log and wait for that to stabilize, meaning the bed has reached equilibrium. To make this more sophisticated, I'd like to get this into some webui somewhere. Klipper already exports this information through its API, so moonraker is the next piece of software that needs to expose it.

Klipper exports it as "power" in its API, so it was trivial to add to your excellent framework as "power" to the endpoint `/server/temperature_store`.
```
diff --git a/moonraker/plugins/data_store.py b/moonraker/plugins/data_store.py
index 77760a7..4a65a0a 100644
--- a/moonraker/plugins/data_store.py
+++ b/moonraker/plugins/data_store.py
@@ -66,9 +66,10 @@ class DataStore:
                 else:
                     new_store[sensor] = {
                         'temperatures': deque(maxlen=TEMPERATURE_STORE_SIZE),
-                        'targets': deque(maxlen=TEMPERATURE_STORE_SIZE)}
+                        'targets': deque(maxlen=TEMPERATURE_STORE_SIZE),
+                        'powers': deque(maxlen=TEMPERATURE_STORE_SIZE)}
                 if sensor not in self.last_temps:
-                    self.last_temps[sensor] = (0., 0.)
+                    self.last_temps[sensor] = (0., 0., 0.)
             self.temperature_store = new_store
             # Prune unconfigured sensors in self.last_temps
             for sensor in list(self.last_temps.keys()):
@@ -86,17 +87,19 @@ class DataStore:
     def _set_current_temps(self, data):
         for sensor in self.temperature_store:
             if sensor in data:
-                last_temp, last_target = self.last_temps[sensor]
+                last_temp, last_target, last_power  = self.last_temps[sensor]
                 self.last_temps[sensor] = (
                     round(data[sensor].get('temperature', last_temp), 2),
-                    data[sensor].get('target', last_target))
+                    data[sensor].get('target', last_target),
+                    data[sensor].get('power', last_power))

     def _update_temperature_store(self):
         # XXX - If klippy is not connected, set values to zero
         # as they are unknown?
-        for sensor, (temp, target) in self.last_temps.items():
[...]

PR #67, opened 2021-01-12
Add Klippers heater PWM value to /server/temperature_store

CapnBry:
As presented in #66, add the pwm output to temperature_store to make it available downstream


Arksine:
Thanks.  The patch looks good, however the commit needs to be signed off as indicated in [contributing.md](https://github.com/Arksine/moonraker/blob/master/docs/contributing.md).

PR #68, opened 2021-01-12
Add Klippers heater PWM value to /server/temperature_store

CapnBry:
As presented in #66, add the pwm output to temperature_store to make it available downstream. Now with a proper commit message! Sorry about the closing the previous PR, I'm a bit daft about how to change an already pushed commit message.


Arksine:
No problem, thanks!

PR #72, opened 2021-01-22
Add option to lock power device while printing

jordanruthe:
This adds an option to lock a power device while printing. Specifically, it ensures that you cannot turn off the printer while the printer is in a printing state.

### DIFF
--- a/moonraker/plugins/power.py
+++ b/moonraker/plugins/power.py
@@ -60,6 +60,13 @@ class PrinterPower:
         IOLoop.current().spawn_callback(
             self._initalize_devices, list(self.devices.values()))
 
+    async def _check_klippy_printing(self):
+        klippy_apis = self.server.lookup_plugin('klippy_apis')
+        result = await klippy_apis.query_objects(
+            {'print_stats': None}, default={})
+        pstate = result.get('print_stats', {}).get('state', "").lower()
+        return pstate == "printing"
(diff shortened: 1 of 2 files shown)


moonraker/plugins/power.py
```diff
@@ -97,6 +104,14 @@ async def _handle_power_request(self, web_request):
...
     async def _process_request(self, device, req):
         if req in ["on", "off"]:
+            printing = await self._check_klippy_printing()
+            logging.info("Device locked: %s" % device.get_locked_while_printing())
+            logging.info("Klippy printing: %s" % printing)
+            if device.get_locked_while_printing() and printing:
+                self.server.error(f"Unable to change power for {device} " +
+                    "while printing")
+                dev_info = device.get_device_info()
+                return dev_info['status']
```
Arksine, line 114: I think you need to raise the exception.  When doing so it won't be necessary to return device status, the client will receive the error instead.

Arksine (commented):
Thanks!  Overall the changes look good to me. We just need to raise the exception when a power request is made while printing.

moonraker/plugins/power.py
jordanruthe: Thanks for the feedback. I just changed it to raise an error

Arksine:
Thanks!

PR #89, opened 2021-02-07
Add Shelly Support to power plugin

zellneralex:
Added support for Shelly relay devices. Testet it with an Shelly 1 and Shelly 2.5. As the API is equal at all Shelly devices of the class it will also work with: Shelly 1 PM, Shelly 1L, Shelly 2, Shelly Plug and Shelly PlugS

Signed-off-by:  Alex Zellner <alexander.zellner@googlemali.com>


Arksine:
Looks good.  I'll merge it soon.

Arksine:
Merged in commit 0fcae5b

PR #94, opened 2021-02-10
Loxone Smart Home Miniserver V1 Power Plugin

EraserFX:
As I saw the implementation for Shelly I thought this should be possible for the loxone devices too. 
Here are the results. With the help of @zellneralex I was able to control outputs at the loxone miniserver v1

greedings
Daniel H. / EraserFX

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -202,7 +202,17 @@ output_id:
 #   If password is set but user is empty the default user "admin" will be used
 #   Provided an output_id (relay id) if the Shelly device supports
 #   more than one (default is 0).
-
+type: loxonev1
+address:
+user:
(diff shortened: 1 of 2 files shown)


Arksine:
Thanks.  After looking at this closer there seems to be a lot of code duplication between Tasmota, Shelly, and Loxone.  My assumption is that the underlying firmware is essentially the same, they simply use different URLs.  What I think I would like to do is rework the Tasmota implementation a bit so that Shelly and Loxone can take advantage of it.

Also, for future reference, all commits needs a signed off line.  See [contributing.md](https://github.com/Arksine/moonraker/blob/master/docs/contributing.md) for details.

PR #97, opened 2021-02-16
Dev power meross

cotepat:
Adds compatibility with Meross SmartPlug and Power Strips devices. Tested extensively on my side (Fluidd, API, G-code...). Works with the 3 channels of my SmartPlug.

Patrice

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -152,7 +152,8 @@ and Tasmota (via http) devices are supported.
 
 [power device_name]
 type: gpio
-#   The type of device.  Can be either gpio, tplink_smartplug or tasmota.
+#   The type of device.  Can be either gpio, tplink_smartplug, meross_smartplug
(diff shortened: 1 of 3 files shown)


Arksine:
Thanks.  The Power Manager code looks okay to me.  I presume that the dependency is necessary to manage credentials?  Do these devices have a way to communicate over a LAN vs the cloud?

If the dependency is absolutely necessary then that is ok, however my preference would be to wait until it is out of RC before merging.

cotepat:
Hi. Thanks for your feedback. Yes, unfortunately it seems Meross devices have to go through the cloud (no LAN only solutions)… Also the dependency is required for the credentials and, more importantly, it manages some kind of limits related to the volume of requests over a period of time (in order not to be kicked out by Meross). It will, for instance, delay a request (1 second or so) if too many requests were done in the last x seconds...

PR #99, opened 2021-02-21
file_manager: Add listen capability for gcodes with inotify

jordanruthe:
Adds inotify listening for gcode directory

### DIFF
--- a/moonraker/plugins/file_manager.py
+++ b/moonraker/plugins/file_manager.py
@@ -10,6 +10,8 @@ import io
 import zipfile
 import logging
 import json
+import subprocess
+import threading
 from tornado.ioloop import IOLoop, PeriodicCallback
 from tornado.locks import Event
 
@@ -21,9 +23,11 @@ METADATA_SCRIPT = os.path.abspath(os.path.join(
 class FileManager:
     def __init__(self, config):
         self.server = config.get_server()
+        self.file_changes_ignore = []
         self.file_paths = {}
         self.gcode_metadata = MetadataStorage(self.server)
         self.fixed_path_args = {}
+        self.inotify_processes = []
 
         # Register file management endpoints
         self.server.register_endpoint(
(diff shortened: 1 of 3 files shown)


Arksine:
Thanks.  I have had a chance for initial review, and I have a few thoughts:
-  Should we implement inotify then I think the best approach would be to use something like [inotify-simple](https://github.com/chrisjbillington/inotify_simple).  It is reasonably up to date and it provides a file descriptor with which we can register using Tornado's ioloop.  This avoids the need to use multithreading and potential issues introduced when combining it with asyncio.
-  Should we decide to proceed with implementing inotify I think that it should handle all "file_list_changed" notifications.  Trying to filter out notifications generated by Moonraker adds unnecessary complexity.  This would result in an API breakage, as the "move_file", "copy_file", and "upload_file" actions would be removed.
- Current work on moving metadata to a database is likely going to result in conflicts with this, so it is probably best that I complete that before proceeding further with inotify support.

That said, I'm not convinced that adding inotify support is desirable, but I am certainly open to discussion.  The current event system works well when Moonraker is used as the file manager.   The argument for inotify is that slicers that don't support Moonraker's upload API can export gcode files to a samba share and have it picked up by the client without refreshing a directory. My counter argument is that this is an issue that could be resolved by the slicers by adding support for Moonraker or through Octoprint's legacy upload API.

cadriel:
- If an API change comes about, then I think that's ok if we can sync releases.
- I also agree that the database support is more important, then history / this (purely bias opinion here.. ).
- I don't think samba should be something users would normally need to have for decent filesystem support, and shouldn't be used as an argument for or against. Uploads should be handled directly in slicer, because that's what users expect.
- Users are going to modify files on their filesystem. I think maybe part of the question here is, do we want to support users directly, outside of the UI? I know we've been focused on preventing users having the need to SSH - and so **with that in mind, is this necessary?**

A concern is that by implementing full inotify support, we'll introduce more complexity in that we'll likely need to better support case changes (among other things). Knowing the bugs we've already spotted / resolved with the existing implementation, are we likely to introduce more headache? 

My 2c. I'm good either way, but I wonder if the benefit is worth the investment.

PR #102, opened 2021-02-28
Cura Octoprint plugin support

grigi:
This PR is a minimal implementation of the Octoprint REST API that is required for Cura to be able to establish a connection and  send gcode/UFP files to moonraker without errors. Currently it only supports the "global apikey authentication" method.

I didn't implement all the API's, as this will now meet my requirements.
Some functionality that the Cura plugin supports that I have not implemented:
* "Selecting" a print (not applicable)
* Starting a print
* Monitoring print progress

I tried to keep this PR as non intrusive as possible.

DONE:
* [x] Connection handshake
* [x] Printer status (errored or not)
* [x] Heater info
* [x] Simulate an Octoprint plugin to allow UFP upload (so images come along)
* [x] Documentation

### DIFF
--- a/moonraker/app.py
+++ b/moonraker/app.py
@@ -169,14 +169,6 @@ class MoonrakerApp:
             params['is_remote'] = False
             self.mutable_router.add_handler(uri, DynamicRequestHandler, params)
             self.registered_base_handlers.append(uri)
-        if "http_octo" in protocol:
-            msg += f" - HTTP: ({' '.join(request_methods)}) {uri}"
-            params = {}
-            params['methods'] = request_methods
-            params['arg_parser'] = api_def.parser
-            params['callback'] = callback
-            self.mutable_router.add_handler(uri, LocalOctoRequestHandler, params)
-            self.registered_base_handlers.append(uri)
         if "websocket" in protocol:
             msg += f" - Websocket: {', '.join(api_def.ws_methods)}"
             self.wsm.register_local_handler(api_def, callback)
@@ -342,20 +334,6 @@ class DynamicRequestHandler(AuthorizedRequestHandler):
         if self.wrap_result:
             result = {'result': result}
         self.finish(result)
-
-class LocalOctoRequestHandler(LocalRequestHandler):
-    async def _process_http_request(self, method):
-        try:
-            args = json.loads(self.request.body)
-        except json.decoder.JSONDecodeError:
-            args = {}
-        try:
-            result = await self.callback(args)
-        except ServerError as e:
-            raise tornado.web.HTTPError(
-                e.status_code, str(e)) from e
-        self.finish(result)
-
 
 class FileRequestHandler(AuthorizedFileHandler):
     def set_extra_headers(self, path):
--- a/moonraker/app.py
+++ b/moonraker/app.py
@@ -142,8 +142,7 @@ class MoonrakerApp:
         self.mutable_router = MutableRouter(self)
         app_handlers = [
             (AnyMatches(), self.mutable_router),
-            (r"/websocket", WebSocket),
-            (r"/api/version", EmulateOctoprintHandler)]
+            (r"/websocket", WebSocket)]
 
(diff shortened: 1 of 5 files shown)


moonraker/moonraker.py
```diff
@@ -30,7 +30,7 @@
 
 CORE_PLUGINS = [
     'database', 'file_manager', 'klippy_apis',
-    'machine', 'data_store', 'shell_command']
+    'machine', 'data_store', 'shell_command', 'octoprint_compat']
```
Arksine, line 33: I don't think that this should be a core plugin.  My ideal scenario is that slicers will eventually implement support for Moonraker directly, at which point this plugin can be deprecated.  In the meantime users can enable it by adding `[octoprint_compat]` in moonraker.conf.

moonraker/app.py
```diff
@@ -347,6 +354,20 @@ async def _process_http_request(self, method):
         self.finish({'result': result})
 
 
+class LocalOctoRequestHandler(LocalRequestHandler):
+    async def _process_http_request(self, method):
```
Arksine, line 358: I recently refactored the dynamic request handlers to accept a JSON body to better support the database APIs, so it should no longer be necessary to add a new handler. Simply register endpoints with the protocol set to ["http"] and set the wrap_result argument to False. Your handler callbacks will need to accept a "web_request" object, with which you can use to parse the arguments or get the args dict directly.

moonraker/plugins/octoprint_compat.py
```diff
@@ -0,0 +1,280 @@
...
+                    'offset': 0,
+                    'target': data['target'],
+                }
+        return temps
+
+    async def _get_version(self, args):
+        """
+        Version information
+        """
+        ver = utils.get_software_version()
```
Arksine, line 130: Rather than importing utils to fetch the software version, you can get it from the config object in `__init__()`, for example:
```python
self.software_version = config['system_args'].get("software_version")
```

Arksine (commented):
Thanks.  I have added a few detailed comments to the review, other than that the implementation looks good.

moonraker/moonraker.py
grigi: I was considering that.
I decided that since I moved the octoprint compatibility login that was there already, I should keep it as core functionality.

I can change it, but note that users that are currently using the slic3r Octoprint functionality will experience breakage. We might need a transition period to give users enough time to notice that they need to change the config file.

e.g. at least get it in the default config files on both mainsailpi and fluiddpi.

moonraker/app.py
grigi: Ah, thanks. That will be useful. I wasn't too fond of this workaround.

moonraker/moonraker.py
Arksine: I think that is ok.   I can also update config_changes.md after this is merged.

grigi:
@Arksine I implemented your changes, but have not yet tested it thoroughly due to a hardware issue. I did run it locally to test that the changes don't regress.
One thing I noticed is that if I send data in JSON format, but forgot to set the `Content-Type:application/json` header correctly, I get this error:
```
  File "/home/grigi/coding/moonraker/moonraker/app.py", line 277, in _default_parser
    args[key_parts[0]] = self._convert_type(val, key_parts[1])
TypeError: _convert_type() takes 2 positional arguments but 3 were given
```
Which doesn't seem quite right.

Lastly, should I add `[octoprint_compat]` section to the sample config file together with `user_changes.md`, or do you want to control that?

Arksine:
Yes, it is important that the API request send the correct content type so the form-data parser doesn't get confused.  However I agree that I should catch that error and re-raise as something more explanatory.

I'll take care of the user_changes.md.  However GitHub is reporting a merge conflict, so we'll need to take care of that before I can merge.  I'll take a deeper at that later tonight.

Arksine:
I fixed the conflicts and merged this along with an update to `user_changes.md`.  Thanks.

PR #107, opened 2021-03-04
Feature indigo power interface

hanzov69:
power: added support for Indigo server

Added class for Indigo (Insteon/x10) server for power control. Touched 
moonraker-requirements.txt, added example to configuration.md and brief 
help to plugins.md

I'm not python proficient so optimization/refactoring may be warranted, but it's been tested and functional in my environment

Signed-off-by: Christian Sullivan hanzo@freezerpants.com

### DIFF
--- a/moonraker/plugins/power.py
+++ b/moonraker/plugins/power.py
@@ -544,7 +544,10 @@ class Indigo(PowerDevice):
                 url = f"http://{self.addr}/{out_cmd}"
                 info_cmd = f"devices/{self.output_name}.json"
                 info_url = f"http://{self.addr}/{info_cmd}"
-                requests.put(url, auth=HTTPDigestAuth(self.user, self.password))
+                requests.put( \
+                    url, \
+                    timeout=5, \
+                    auth=HTTPDigestAuth(self.user, self.password))
                 response = await http_client.fetch( \
                     info_url, \
                     auth_mode='digest', \
@@ -561,7 +564,10 @@ class Indigo(PowerDevice):
                 url = f"http://{self.addr}/{out_cmd}"
                 info_cmd = f"devices/{self.output_name}.json"
                 info_url = f"http://{self.addr}/{info_cmd}"
-                requests.put(url, auth=HTTPDigestAuth(self.user, self.password))
+                requests.put( \
+                    url, \
+                    timeout=5, \
+                    auth=HTTPDigestAuth(self.user, self.password))
                 response = await http_client.fetch( \
                     info_url, \
                     auth_mode='digest', \
--- a/moonraker/plugins/power.py
+++ b/moonraker/plugins/power.py
@@ -11,12 +11,12 @@ import json
 import struct
 import socket
 import gpiod
-import requests
-from requests.auth import HTTPDigestAuth
+
 from tornado.ioloop import IOLoop
 from tornado.iostream import IOStream
 from tornado import gen
 from tornado.httpclient import AsyncHTTPClient
+from tornado.httpclient import HTTPClientError
 from tornado.curl_httpclient import CurlAsyncHTTPClient
 from tornado.escape import json_decode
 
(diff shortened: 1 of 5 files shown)


moonraker/plugins/power.py
```diff
@@ -522,6 +527,100 @@ async def set_power(self, state):
...
+                msg = f"Error sending indigo command: {command}"
+                logging.exception(msg)
+                raise self.server.error(msg)
+        elif command == "off":
+            try:
+                out_cmd = f"devices/{self.output_name}?isOn=false"
+                url = f"http://{self.addr}/{out_cmd}"
+                info_cmd = f"devices/{self.output_name}.json"
+                info_url = f"http://{self.addr}/{info_cmd}"
+                requests.put(url, auth=HTTPDigestAuth(self.user, self.password))
```
grigi, line 564: You'd want to avoid using a synchronous client like requests inside an event loop function (in `async def` function).
It would block the event loop for the duration of the request, which is especially a problem when there's a network error (default timeout is something like 20 minutes) and will make everything look like it broke.

You should be able to use the async client like you did just below this:
```py
await http_client.fetch(url, method='PUT', auth_mode='digest', auth_username=self.user, auth_password=self.password)
```
https://www.tornadoweb.org/en/stable/httpclient.html#tornado.httpclient.HTTPRequest

hanzov69: Thanks for the feedback!
This was initially the path I pursued, but I ran in to issues. After receiving the PUT, Indigo responds with a 303 (which is normal/expected). Tornado tries to follow the 303 (also expected), but strangely does another PUT rather than a GET. This results in a 500 from Indigo, which causes an error and the plugin stops responding. 

I attempted to direct tornado to ignore redirects with `follow_redirects=0`, but then the 303 was passed back and treated as an error which also caused the plugin to stop responding. 
Digging down farther, this seems to be a libcurl issue.  Looking at tornado, they call this out here:
https://github.com/tornadoweb/tornado/blob/6cdf82e927d962290165ba7c4cccb3e974b541c3/tornado/simple_httpclient.py#L644

As I'm using the FluiddPi (which is a Raspbian Lite based image), it's relying on "Buster" provided libcurl, which is 7.64.0 ( https://packages.debian.org/buster/libcurl4 ). 

At this point I gave up and used `requests`, but I do see the problem you present. I'm totally open to a better/more correct way to do this, I just don't know what that looks like. 

And now, for some logs showing what I'm seeing. 
[ignore_redirects.txt](https://github.com/Arksine/moonraker/files/6085020/ignore_redirects.txt)
[follow_redirects.txt](https://github.com/Arksine/moonraker/files/6085021/follow_redirects.txt)

grigi: Did you try wrapping it in a try-catch statement?
e.g.
```python
try:
    await http_client.fetch(
        url, method='PUT', auth_mode='digest',
        auth_username=self.user, auth_password=self.password,
        follow_redirects=0)
except httpclient.HTTPError:
    # We expect an HTTPError due to 303 redirect
    pass
```

hanzov69: This was exactly the cluebat I needed to be hit with! Thank you so much!

grigi: Yay! A pleasure :-)

Arksine:
Thanks.  Generally speaking I would like to avoid adding new dependencies to support additional "power" devices.  At a high level I think adding MQTT support to Moonraker would offer more flexibility and should cover communicating with hubs such as Indigo or Home Assistant.  Adding this support is on my TODO list, however I have a few things I need to take care of before I can get there.

hanzov69:
Makes sense. 
In a similar vein, do you have plans for "third party" plugin support? Would be nice to be able to tailor moonraker to my needs/tastes, but be able to keep the core up to date without "tainting" the local repo.

Arksine:
Yes, that is being discussed and it is definitely something I would like to provide.  The trick is providing a 3rd party interface that is robust enough to be useful, but safe enough that users don't oops themselves into a malicious plugin.

hanzov69:
Sweet, I'll take that route instead. I super appreciate the guidance from y'all. Going to close the PR, thanks!

PR #110, opened 2021-03-05
docs: fix incorrect wording. 'mainsail' should correctly read 'moonraker'

th33xitus:
Fixes a small technically incorrect wording in configuration.md and web_api.md. G-code files are uploaded to a moonraker instance and not to a mainsail instance.


Arksine:
Thanks.

PR #115, opened 2021-03-11
power: Add CMD device to allow using shell commands for power control

Doridian:
This adds a cmd "device" to the power module to allow running shell scripts to check for status and power on/off the printer.
It is configured  as follows:

```
[power printer]
type: cmd
on: /my/shell/script.sh arguments go here
off: /my/other/script.sh arguments go here
status: /my/third/script.sh arguments go here
```

The `status` script is expected to indicate current printer status via exit code.
Code 0 means "on", code 1 means "off", any other code triggers an error.
This follows (mostly) the usage of the PSU control OctoPrint plugin (as to make scripts port easier).

### DIFF
--- a/moonraker/plugins/power.py
+++ b/moonraker/plugins/power.py
@@ -558,6 +558,7 @@ class NeoPixel(PowerDevice):
         pass
 
     def set_all(self, color):
+        print("Set color", color)
         for i in range(self.neopixel.numPixels()):
             self.neopixel.setPixelColor(i, color)
         self.neopixel.show()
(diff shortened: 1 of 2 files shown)


Arksine:
Thanks.  Due to the undefined nature of executing configurable shell scripts and unpredictable results of allowing them this isn't something I will merge into the official repo.  I do have plans to add MQTT support which could potentially be used control devices that otherwise might need a script.

I also plan to offer a 3rd party plugin interface which will have the ability to register additional power devices.  Something like this could be implemented through such a plugin.  In this scenario the 3rd party developer would be responsible for supporting the plugin and any potential issues that arise.

PR #117, opened 2021-03-11
power: Add NeoPixel control for RasPi

Doridian:
Make sure to read the instructions/warnings at https://pypi.org/project/rpi-ws281x/

As you indicated, you don't want shell scripts for security reasons, which is understandable, this is my attempt to bring the control I need directly into Moonraker (as opposed to calling out an external python script).
I understand if you don't want this either, as it is RasPi specific and adds an additional dependency.

I will also be contributing a HomeSeer (home automation system) integration as that is what I use to control the printer's power itself. That should be less of an open question as that is just "the usual" HTTP call-outs.

On another note, I am aware Klipper can control NeoPixels itself, however not on a RasPi (as a RasPi's GPIO is technically too slow to control them well, the rpi_ws281x module gets around this on certain pins by using DMA)

### DIFF
--- a/moonraker/plugins/power.py
+++ b/moonraker/plugins/power.py
@@ -11,6 +11,7 @@ import json
 import struct
 import socket
 import gpiod
+from rpi_ws281x import Adafruit_NeoPixel, Color
 from tornado.ioloop import IOLoop
 from tornado.iostream import IOStream
 from tornado import gen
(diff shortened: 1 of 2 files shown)


Arksine:
I think that this would be a better candidate for a 3rd party plugin when the interface is made available.  Controlling neopixels is outside of the scope of what the power plugin is intended to do, and as you mentioned I don't want to pull in additional dependencies to support additional device implementations.

Issue #124, opened 2021-03-24
[FR] Maintenance counter/timer


grigi:
## Background:
With our local South African Voron WhatsApp group we discussed having a counter that would remind you to "service" (that is oil the rails/rods, and check that everything is still good) your printer. Now that Mainsail keeps track of printing history we think this small feature can work.

We discovered that HiWin [lubrication spec](https://www.hiwin.com/pdf/lubricating_instructions.pdf) recommended a 200km re-oiling interval for MGN9 (which Voron2 uses).

After some extrapolation of what a typical 0.4mm nozzle would have to move we deduced that you could estimate it based off how much filament is printed we estimated that the axis should move about 10× the distance of 1.75mm filament used. That should put a service interval of roughly 20km (or 60kg of filament). This seems to line up with how often the Print-It-Forward printers get serviced.

## Proposal:
Model service intervals in a similar way as is done in the automotive industry:
* Distance traveled
* Time since last service

(Whichever comes first)

Since with History we now have reasonable filament length used (it doesn't need to be perfectly accurate), along with a "last serviced timestamp", we could use this simple model to remind the user to re-service.

We could even calculate an estimate % of progress till next service with this data:
```py
wear_filament_ratio = (history.filament_printed - maintenance.last_service_filament) / config.service_filament
wear_time_ratio = (now() - maintenance.last_service_time) / config.service_time
estimated_wear = max(wear_filament_ratio, wear_time_ratio)
```

## Questions

Assuming this model is sane, how would we notify the user about wear level, or if it's overdue? Is this a presentation issue, and we just need to expose it as an API here?
This obviously needs access to the history plugin, but should store service counter data in it's own datastore. I'm not certain how this would work in the code.

Arksine:
Interesting.  A client could feasibly implement this functionality now without any addition to moonraker.  Clients can store persistent data in the database, so they could track the service counter.

> Can we trust the total_printed counter to not shrink when we reach the job limit and start cleaning up old data in the history plugin?

Yes.  The `job_totals` are updated as jobs complete, they are not calculated from job entries.

Another alternative would be to track the toolhead position and store total distance traveled for each axis.  This would be relatively simple and provide a very accurate representation of distance traveled.

grigi:
> Interesting. A client could feasibly implement this functionality now without any addition to moonraker. Clients can store persistent data in the database, so they could track the service counter.

[...]

PR #125, opened 2021-03-25
Axes tracking

jordanruthe:
Initial commit for #124

### DIFF
--- a/docs/web_api.md
+++ b/docs/web_api.md
@@ -500,6 +500,34 @@ included.
     }
 }
 ```
+
+#### Retrieve axes total distance traveled
+HTTP request:
+```http
+GET /server/axes_distance
+```
+JSON-RPC request:
+```json
+{
+    "jsonrpc": "2.0",
+    "method": "server.axes_distance",
+    "id": 7643}
+```
+
+Returns:
+
+An object with the total distance traveled for each axes and the epoc timestamp
+since the axes have been tracked.
+```json
+{
+    "x": 0.,
+    "y": 0.,
+    "z": 0.,
+    "e": 0.,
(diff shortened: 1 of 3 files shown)


grigi:
This is looking pretty good already :+1: 

This is for tracking moves, so it won't be able to track things like movement due to tilt adjustment, and will be less real for Delta printers.

I still like the ability to track E, that would be more accurate than the "filament used" in history tracking.
Could you make it work for multiple extruders as well please?

Arksine:
I think its a good start, although I have a few early comments.
* I don't think we should use gcode_move to derive the position.  The toolhead position is the actual movement after transforms, so it should produce a more accurate result.
* I think its worth exploring accumulating the position data in Klipper itself, which would make the above point moot.  If we accumulate position data in Klipper we would just need to keep a running total across restarts.  One nice thing about doing it in Klipper is that it would take care of homing for you (in fact homing movement would be included in the accumulated position).
* I think we can put this in `data_store.py`.  Its role is to track and store printer data, currently it tracks temperature data and gcode requests/responses.  The data it stores isn't persistent, but there is no harm in tracking additional data that is.

FWIW, `filament_used` does track E.  It excludes extrusion moves during a pause, which IMO makes it is the appropriate field to track print history.

The toolhead reports the extruder its using, so it is possible to track position per extruder.  However if we move the accumulation on E would be a total of all extruders.  Accumulating position in Klipper would be far more accurate though.  Moonraker receives updates ever 250 ms, which means that we will miss some data when a tool changes direction, so IMO it would be a good tradeoff.

Arksine:
Since this is no longer being worked on I am closing it.  Thanks.

PR #129, opened 2021-03-27
Added support for TPLinkSmartPlug multi plug devices

reztierk:
Just a couple changes to add support for TP-Link power strips to the power component module. Tested on two separate KP303 power strips. Works the same as the octoprint plugin by adding a `/#` at the end of the IP address to define which "child" you are assigning.

### DIFF
--- a/moonraker/components/power.py
+++ b/moonraker/components/power.py
@@ -317,10 +317,18 @@ class TPLinkSmartPlug(PowerDevice):
     async def _send_tplink_command(self, command):
         out_cmd = {}
         if command in ["on", "off"]:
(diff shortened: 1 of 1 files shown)


Arksine:
The patch looks good, however some of the lines need to be capped to 80 characters in length.  Also the commit needs to be signed off as describing in the [contributing document].(https://moonraker.readthedocs.io/en/latest/contributing/)

Thanks.

reztierk:
Thanks for the feedback, I have addressed the character length and signed off on the most resent commit.

Arksine:
Thanks!

PR #131, opened 2021-03-29
Extend history API with reverse order (by default)

ayufan:
This improves history API:

1. By default returns items in a reverse order (as you expect newest at top)
2. Optimises a search procedure to define a lower and upper boundary assuming that `cached_job_keys` is `order asc`

I noticed this bug when looking at [fluidd](https://github.com/cadriel/fluidd) implementing history API.

### DIFF
--- a/moonraker/components/history.py
+++ b/moonraker/components/history.py
@@ -88,7 +88,8 @@ class History:
 
     async def _handle_jobs_list(self, web_request):
         i = 0
-        end_num = len(self.history_ns)
+        count = 0
+        end_num = len(self.cached_job_ids)
         jobs = []
         start_num = 0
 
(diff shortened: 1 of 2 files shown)


Arksine:
The patch looks good.  We will need @meteyou @cadriel and @jordanruthe to weigh in on the behavior before we merge.

Also the commit needs to be signed off as outlined in [the contributing documentation](https://moonraker.readthedocs.io/en/latest/contributing/).

Thanks.

cadriel:
Thanks for the submission, it looks good. I have some comments (sorry for the novel...);

Generally speaking, it's very easy to sort this data client side by default. The table has built in support for this, I'd just forgotten to set the default sort :)

In the web world, data sorting, paging, filtering and search are generally handled either server side, or client side. You can't realistically have some of these things happening on the server, and some on the client.

Why? Because the client can only manage the data it has already received. I.e., I could request the first page of data, but the client can only then sort, filter and search that 1 page of data - which makes little sense.

I think ideally, I'd like to see full server side paging implemented. This way, the user has exactly the same experience on the client, but the client doesn't need to request ALL history.

The other option is to leave it as is, with the expectation that we'll pull all history data if required.

We were theorizing numbers as to how big the history data may get - coming up with around the 512kb range per year roughly. Overall I like to keep requests below 150ms max - and as that data grows it's going to take longer to pull.

Both Mainsail and Fluidd use the same table component, which applies a standard method of handling server side data. If we did decide to go down this path, then matching that would be great.

https://vuetifyjs.com/en/components/data-tables/#server-side-paginate-and-sort

I'm by no means suggesting we have to do this, but I'd certainly like to reliably let the user sort and browse all history data, instead of only a subset of it without the risks associated of pulling a large dataset.

ayufan:
@cadriel 

> The other option is to leave it as is, with the expectation that we'll pull all history data if required.

This in general is not the best idea given the amount of data that needs to be transmitted :)

> I'm by no means suggesting we have to do this, but I'd certainly like to reliably let the user sort and browse all history data, instead of only a subset of it without the risks associated of pulling a large dataset.

I think this change to some extent allows this to be fulfilled. It also returns data in a order that makes sense.

But I lack the clear answer: should this be merged in this way or not? :)

ayufan:
@meteyou Pinging you as I noticed that you also fetch a default of `limit=50`. Which is fine, but this improves a few aspect :)

@cadriel For example this MR also fixes `reprint` button support where we use a history data.

moonraker/components/history.py
```diff
@@ -88,34 +89,58 @@ async def _handle_job_request(self, web_request):
...
         start = web_request.get_int("start", 0)
-        if start >= end_num or end_num == 0:
-            return {"count": 0, "jobs": []}
+        order = web_request.get_str("order", "desc")
 
-        for job_id in self.cached_job_ids:
-            job = self.history_ns[job_id]
-            if since != -1 and since > job.get('start_time'):
+        if not order in ["asc", "desc"]:
+            raise ServerError(f"Invalid `order` value: {order}")
```
jordanruthe, line 104: The other references in this module do
`return self.server.error()`
I think we should keep this consistent by changing the others or using self.server.error here.

jordanruthe:
Overall I don't have any objections and think it's a good idea to have the order capability.

meteyou:
thx for the ping. this PR looks good for me.

but the server side rendering (that @cadriel posted) would be also a very nice feature, but then we (fluidd & mainsail) need a "printed" flag or/and the job_id of the last history entry to pull this data in the file manager.

but i'm still trying to figure out what would be best and how much data it would be... whether we should worry about anything else or this worry is justified...

ayufan:
> but the server side rendering (that @cadriel posted) would be also a very nice feature, but then we (fluidd & mainsail) need a "printed" flag or/and the job_id of the last history entry to pull this data in the file manager.

I think we can work on this iteratively. Especially to cross-match data.

moonraker/components/history.py
ayufan: This is very good suggestion. Fixed.

Arksine:
I'm okay with merging this as a stopgap while we look into pagination if that is the consensus.

ayufan:
@Arksine Do you think you could merge this?

Arksine:
Yes.  I wanted to verify that there were no objections given that it changes the default behavior of the endpoint.  Since there are none, I have proceeded with the merge.

Issue #90, opened 2021-02-08
FR: Support TPLink Tapo Smart Plugs


claitz:
Hi there,
I have written a mockup plugin for Moonraker that thanks to [fishbigger's TapoP100](https://github.com/fishbigger/TapoP100) Python library correctly controls my Tapo P100 smart plug.
I just think it'd be great to have this built-in in the existing power plugin.

I could try supporting with this FR but don't expect the pretties Python code :)

PR #132, opened 2021-03-31
Adding support for TPLink Tapo plugs

fsnuxer:
Add support for TPLink Tapo plugs.
Thanks to https://github.com/fishbigger/TapoP100 Library
Disclaimer :  I'm no python coder
Works for me at least ;)

https://github.com/Arksine/moonraker/issues/90

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -177,8 +177,8 @@ and Tasmota (via http) devices are supported.
 
 [power device_name]
 type: gpio
-#   The type of device.  Can be either gpio, tplink_smartplug, tasmota
-#   or homeseer.
+#   The type of device.  Can be either gpio, tplink_smartplug, tplink_tapo,
+#   tasmota or homeseer.
 #   This parameter must be provided.
 off_when_shutdown: False
 #   If set to True the device will be powered off when Klipper enters
@@ -214,6 +214,13 @@ port:
 #   address should be a valid ip or hostname for the tplink device.
 #   The port should be the port the device is configured to use.  The
 #   address must be provided. The port defaults to 9999.
+address:
+email:
+password:
+#   The above options are used for "tplink_tapo" devices.  The
+#   address should be a valid ip address for the tapo device.
+#   The email and password must be those supplied for the device
+#   registration within the tplink tapo mobile application. 
 address:
 password:
 output_id:
--- a/moonraker/components/power.py
+++ b/moonraker/components/power.py
@@ -16,6 +16,8 @@ from tornado.iostream import IOStream
 from tornado import gen
 from tornado.httpclient import AsyncHTTPClient
 from tornado.escape import json_decode
+from PyP100 import PyP100
+
 
 class PrinterPower:
     def __init__(self, config):
@@ -32,6 +34,8 @@ class PrinterPower:
                     dev = GpioDevice(cfg, self.chip_factory)
                 elif dev_type == "tplink_smartplug":
                     dev = TPLinkSmartPlug(cfg)
+                elif dev_type == "tplink_tapo":
+                    dev = TPLinkTapo(cfg)
                 elif dev_type == "tasmota":
                     dev = Tasmota(cfg)
                 elif dev_type == "shelly":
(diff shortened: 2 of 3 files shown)


Arksine:
Thanks.  At the moment I am not accepting contributions to the power plugin that requires dependencies.  I have plans to add support for MQTT integration which should allow Moonraker to interact with nearly all devices like this though a 3rd party.  I also plan to implement a 3rd party plugin interface that will allow users to register their own power modules, with the caveat that the plugin developer will be responsible for maintaining it.

fsnuxer:
Fair enough. (I guess in this case the library used could be included or  copy/pasted since it's very small.)

Another solution (for this plug and other) could be a generic  "external command" system ?
=> call a cli command just with an ON / OFF / STATUS parameter ?
And each and every plug could work without adding a dependency for moonraker - and without requiring an mqtt setup : just an external program/script glue...

Arksine:
If the implementation can be added without the dependency I would be open to merging.  After all, this is how the tplink smartplug implementation works.  We would just need to make sure that the license is compatible and any borrowed code is properly credited.

Given that moonraker provides remote access with the ability to modify your configuration files I am not comfortable with supporting configurable CLI commands.  I suspect that someone will end up implementing it in a 3rd party plugin, however my position will be to discourage it.

NpFaith:
@fsnuxer Can you provide the configuration code? I'm new in the Klipper universe and I have p100 smartplug. Thanks in advance

fsnuxer:
> @fsnuxer Can you provide the configuration code? I'm new in the Klipper universe and I have p100 smartplug. Thanks in advance

Hello. Not sure if the code is still ok since I'm not using it anymore.
( The patch included updated documentation.md  ;) )

[power mytapoplug]
type: tplink_tapo
email: youremailusedwhenyouregisteredtheplug
password: yourpasswordusedwhenyouregisteredtheplug
address: ip.of.the.plug

Arksine:
Moonraker does not currently support Tapo plugs.  As mentioned above, support could be added however it needs to be done without the fishbigger TapoP100 dependency.  I wasn't specific as to the issues with the library, so for clarity I will elaborate:
- The TapoP100 package relies on the `requests` module to make http requests.  These are synchronous/blocking requests and they will block Moonraker's asyncio event loop.  There are numerous ways around this, however the best solution is probably to use Tornado's async http client.
- The package also depends on `pycryptodome`.   This isn't a problem for a Raspberry Pi or common x86 linux distros like Ubuntu as they will have prebuilt wheels.  Many other SBCs running Armbian will need to build the wheel for this package and it would be important to validate that a) the wheel can be built in a timely manner and b) the wheel can be built without any additional dependencies or user intervention.

NpFaith:
@¶Arksine I'm noob in this, it is possible for you to explain with more detail the right way to put moonraker communicating with Tapo P100? I'm enjoying allot Klipper, thanks for the help.

NpFaith:
@fsnuxer thanks for the help

Arksine:
@NpFaith An interested developer would need to write an implementation that accesses the Tapo P100 APIs using asynchronous HTTP requests.  They would need to be familiar with asyncio, Moonraker's component API, and the Tapo P100 API.  Unfortunately I there isn't much more detail I can add, these are items the developer would need to research.  If there are specific implementation questions with regard to Moonraker I would be happy to answer them.

NpFaith:
@Arksine I'm not a programmer, I understand algorithmic logic but I have no knowledge of these languages, as I managed to get it to work through the octoprint I thought that with a few little tricks it would be the same in moonraker. My knowledge is really limited. I don't want to use octoprint because I prefer Fluidd, what I need to be fine-tuned for me is to work with the smartplug tapo p100. What languages ​​do you need to know to be able to make the connection between smartplug and moonraker? Thanks anyway and if I need more help I'll talk to you.

Arksine:
Octoprint very likely has a 3rd party plugin that supports the Tapo P100.  Unfortunately there is no way to trick support into Moonraker if you want the device visible in Fluidd, support must be added to the `power` component, which is what this Pull Request attempted to do.  Moonraker is written Python, specifically targeting CPython 3.7+.

julianvolden:
Hi there, any updates on TPlink Tapo plugs? 

Currently i have a tapo p115 plug i want to use

Arksine:
It should be possible to support Tapo plugs with the `generic` http implementation.   Since Moonraker is no longer accepting HTTP specific implementations I am closing this.  Thanks.

PR #135, opened 2021-04-02
power.py: fix issue with single relay tasmota

### LINKED ISSUE #134: Tasmota 1 relay powerdevice  not working
Hi, me again.

I've found a bug in the code : 

Say I have a tasmota device, a sonoff S26 (that's what I have). this device has only 1 relay. The firmware is correctly flashed (I've tried on my "old one" and also on a fresh wemos d1 as a sonoff basic (the default after tasmota flash).

moonraker defaults output_id to "1", that is logical.

If I send a request to get the status, here is what happens :
```
arnaud@Arnaud-2650v3:~/core/config/custom_components$ curl -X GET "http://192.168.1.52/cm?user=admin&password=&cmnd=Power1"
{"POWER":"OFF"}
```

see the "get 'power1', receive 'power' " ?  you can argue it's a tasmota bug, but as long as the device has only one relay configured, the power device number is stripped out in the response (which is logical).

But moonraker [EXPECTS](https://github.com/Arksine/moonraker/blob/c36f989c529584753623eb14ab475511eeb5e0ef/moonraker/components/power.py#L458)  to see a number in the response .... so it does not get any status for the power device. no error is thrown, no log ... 

I think that here, moonraker SHOULD try to see if it's talking to a single relay device if it fails to get a multi-relay device status (that's actually what moonraker exepects by default)

aschor:
add a check and handle consequently in case tasmota device is only one relay and would return a power value when asked a power1 value

Fixes #134 

Signed-off-by:  Arnaud Schaer <arnaud.schaer@wanadoo.fr>

### DIFF
--- a/moonraker/components/power.py
+++ b/moonraker/components/power.py
@@ -455,7 +455,13 @@ class Tasmota(PowerDevice):
     async def refresh_status(self):
         try:
             res = await self._send_tasmota_command("info")
-            state = res[f"POWER{self.output_id}"].lower()
+            try: 
+                state = res[f"POWER{self.output_id}"].lower()
+            except KeyError as e:
+                if self.output_id == 1 :
+                    state = res[f"POWER"].lower()
+                else:
+                    raise KeyError(e)
         except Exception:
             self.state = "error"
             msg = f"Error Refeshing Device Status: {self.name}"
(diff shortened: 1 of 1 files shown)


Arksine:
Looks good.  Thanks.

PR #137, opened 2021-04-05
machine.py : allow user shutdown and reboot scripts

aschor:
add the ability for a user to alter the default behaviour when using the API to reboot or poweroff the host by executing shell scripts if present :

- scripts/pre_shutdown_sequence.sh
- scripts/pre_reboot_sequence.sh

files must be executable.

it is mainly useful for "last will" commands, for example if you want to tell your home automation software that the host (pi) is going to shutdown, and it can switch the smartplug "off" in some time

if you are open to this change, I suppose that some doc/example should be modified too (I personnaly need this to shutdown the smartplug after I shutdown the printer, as I can change OS systemd scripts, but that is un-reliable and one time out of 2, my home automation is not aware of the printer shutdown - the curl does not fire before network shutdown)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 __pycache__/
 *.py[cod]
 *$py.class
+scripts/pre_*_sequence.sh
--- a/moonraker/components/machine.py
+++ b/moonraker/components/machine.py
@@ -4,6 +4,7 @@
 #
 # This file may be distributed under the terms of the GNU GPLv3 license.
 import logging
+import os
 from tornado.ioloop import IOLoop
 
 ALLOWED_SERVICES = ["moonraker", "klipper", "webcamd"]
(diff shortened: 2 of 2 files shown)


Arksine:
Thanks.  Generally speaking I'm weary of running user configured shell scripts from Moonraker.  It can be difficult to support undefined behavior and there are potential security implications.

That said, a notification component will be coming and it could likely be used to accomplish your goals.  I think we could fire off a `machine:shutdown` event that a notification component could pick up and broadcast to a configured URL prior to shutting down.   Other components could capture such an event as well.  

Also of note, there is a pull request to implement "timer" functionality for tasmota and shelly devices.  Apparently this is built in.  There are a couple of minor issues with it, however they should be easy to resolve. This may be applicable if you need a device to power off after shutdown.

aschor:
ok, I'll wait and see how it works. But I doubt that a notification handler can handle shutdown/reboot notifications : since it would work async from the shutdown/reboot command, it would be killed by the O.S. before handling the event most probably (unless notifications are subscribtions and we don't shutdown until all clients have acked the reception of the notification, but what if a client does not ack ...)

Arksine:
The notification should work as it would fire and process before the shutdown command is executed.  I will need to implement a blocking version of `send event`, that shouldn't be an issue though.

Arksine:
I'm going to go ahead and close this.  Thanks.

PR #142, opened 2021-04-10
extract_metadata: assign correct data parameter

th33xitus:
This should fix an issue where the total filament of g-code files generated by ideaMaker is not parsed correctly and therefore can't be displayed by clients.

screenshot

For testing purposes, here an example g-code created by ideaMaker 3.6.1.
[ideamaker_3.6.1.gcode.zip](https://github.com/Arksine/moonraker/files/6290062/ideamaker_3.6.1.gcode.zip)


Arksine:
Thanks!

PR #143, opened 2021-04-13
extract_metadata: parse filament_weight_total

th33xitus:
There was a request of adding filament weight as an additional key for clients to work with as can be seen here: https://discord.com/channels/758059413700345988/826580737154089051/829425183168135189

A new key `filament_weight_total` was defined clients can access.
Supports G-Code from:
 - PrusaSlicer
 - SuperSlicer
 - Slic3r
 - Simplify3D
 - ideaMaker
 - Cura

**Note:**
Cura offers a special placeholder for the filament weight that was used here.
This requires the user to have a specific comment in the start g-code section defined:
`;Filament weight = {filament_weight}`

**Client developers should mention this requirement in their documentation.**

---

Additional change:
- Added parsing of `filament_total` for Slic3r

### DIFF
--- a/scripts/extract_metadata.py
+++ b/scripts/extract_metadata.py
@@ -109,6 +109,9 @@ class BaseSlicer(object):
     def parse_filament_total(self):
         return None
 
+    def parse_filament_weight_total(self):
+        return None
+
     def parse_estimated_time(self):
         return None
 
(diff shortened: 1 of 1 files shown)


Arksine:
The patch looks good, I just need you to sign off on your commits.  Thanks.

th33xitus:
You are right, i forgot!
I have force pushed the reworded commit messages.

Arksine:
Thanks!

PR #156, opened 2021-05-02
power: Add Home Assistant switch support

sjousse:


Arksine:
Thanks.  In general the patch looks good, however there appear to be a some lines over 80 characters in length, starting at line 647 and ending at 657.

sjousse:
You're right, I just fixed this.

Arksine:
Thanks.

PR #159, opened 2021-05-08
file_manager:  add support for external file events via inotify

Arksine:
This is a proposal to add support for inotify events to Moonraker.  This amounts to a significant refactoring of the file_manager, so I want to slow roll merging this PR, with the option to abort if we feel that it could cause significant issues. 

From the Client API perspective the changes are rather mild, and mostly pertain to the `notify_filelist_changed` notification.  Specifically the actions have changes, as some of the exisiting actions do not make sense when dealing with inotify.  The [api_changes.md](https://github.com/Arksine/moonraker/blob/17f9eff89d13db9af11b958fda3b5b6d09a49ad5/docs/api_changes.md) document in this PR contains details on all the API changes in this pull request.  It may also be useful to look at the diff of `web_api.md`. 

As mentioned in the linked api_changes doc, there are some additional changes such as the removal of `notify_metadata_changed` and differences to the return values of some of the file operation requests.  Ultimately my goal with this is to create more consistency within the file manager API, where clients can expect a request to return before the a notification is received.  It also makes sense that if metadata processing is necessary then it should occur before a request returns, as the client is then free to request metadata for a file or directory after a notification has been received.

@meteyou @cadriel @jordanruthe

### DIFF
--- a/docs/api_changes.md
+++ b/docs/api_changes.md
@@ -1,6 +1,48 @@
 ##
 This document keeps a record of all changes to Moonraker's web APIs.
 
+### May 8th 2021
+- The `file_manager` has been refactored to support system file
+  file events through `inotify`.  Only mutable `roots` are monitored,
+  (currently `gcodes` and `config`).  Subfolders within these
+  these roots are also monitored, however hidden folders are not.
+  The following changes API changes have been made to acccommodate
+  this functionality:
+    - The `notify_filelist_changed` actions have changed.  The new
+      actions are as follows:
+        - `create_file`: sent when a new file has been created.  This
+          includes file uploads and copied files.
+        - `create_dir`: sent when a new directory has been created.
+        - `delete_file`: sent when a file has been deleted.
+        - `delete_dir`: sent when a directory has been deleted.
+        - `move_file`: sent when a file has moved.
+        - `move_dir`: sent when a directory has moved.
+        - `modify_file`: sent when an existing file has been modified
+        - `root_update`: sent when a root directory location has been set.
+          For example, if a user changes the gcode path in Klipper, this
+          action is sent with a `notify_filelist_changed` notification.
+    - File list notifications for gcode files are now only sent after
+      all metadata has been processed.  Likewise, requests to copy,
+      move, or upload a file will only return after metadata has been
+      processed.  Notifications are synced with requests so that the
+      request should always return before the notification is sent.
+    - Thumbnails are now stored in the `.thumbs` directory to prevent
+      changes to thumbnails from emitting filelist notications. This
+      change will be reflected in the metadata's `relative_path` field,
+      so clients that use this field should not need to take additional
+      action.  Note that existing thumbnails will remain in the `thumbs`
+      directory and filelist notifications will be sent for changes to
+      these thumbnails.
+    - The `notify_metadata_update` notification has been removed  Clients
+    - can reliably expect metadata to be available for new or moved gcode
+      files when a request returns.
+    - The return values for several endpoints have been updated.  They
+      now contain information similar to that which is pushed by the
+      `notify_filelist_changed` notification.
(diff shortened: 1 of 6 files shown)


Arksine:
Yeah, I think we need to sync release.   Hopefully client changes will be minimal, so everyone could create a branch with the patch to support the changes then easily merge it when ready.  You can have testers pull this PR to play with your changes if necessary.

I have been working on this off and on for quite some time so it has seen a lot of local testing, but the more testing the better given that the file manager is a critical part of Moonraker.

cadriel:
Yep, sounds good. I'll get a local branch going in Fluidd.

jordanruthe:
Can there be consistency between the server.files.list method and the inotify notifications? 

For instance, server.files.list uses `filename` and the inotify notifications use `path`

Arksine:
Its possible, however there is a good reason for the difference.  The file list will always return items with a relative path to a file, whereas the `notify_filelist_changed` notification could apply to a file or directory.  Right now the nomenclature for `server.files.list` is consistent with `server.files.get_directory`.

That said, I'm not opposed to changing `server.files.list` to report `path` instead of `filename` given that it is a relative path to a file.  The directory request is a bit different as the results are always going to be relative to the requested directory, so `filename` and `dirname` make more sense there.

jordanruthe:
I think `path` makes more sense for `server.files.list` as it's a full path rather than just a filename. 

@cadriel @meteyou This would also have implications for you two, what are your thoughts? I'm not sure if the possible ramifications on the client side would be worth the change.

Arksine:
I went ahead and pushed this change to the PR.  I'm reasonably confident that both Mainsail and Fluidd use `server.files.get_directory` to populate their file trees, so a change to `server.files.list` is unlikely to affect them.    However it would be a good idea to make absolutely sure.

cadriel:
I don't use `server.files.list` - its redundant for me because I only load the directory at runtime when a user navigates to it - saving me having to load all at once. So the suggested change wouldn't be a problem over and above everything else ;)

In saying that tho, when retrieving a list of files - being consistent with what they return would be good.

Arksine:
As of tonight I plan to merge this pull request on Tuesday around 7:30pm EDT (UTC-23:00).  If anyone has any objections feel free to raise them.  Thanks.

meteyou:
would it be possible to send also a notify for new root paths with "root_update" or add a "root_create", "root_delete"?

Arksine:
@meteyou You want the new system path sent with `root_update`?  I could do that.

With regard to `root_create` and `root_delete`, that would be a bit more involved.  Delete would be possible as a watched directory does receive inotify events.  Receiving create events wouldn't be possible without watching the root's parent, which would open a big can of worms.

meteyou:
i don't care about the "create" itself, but about a new root directory being added. i just tested it and noticed that after booting without klipper only config & gcode exists and after klipper starts the first time config_examples and docs are added. would be nice to have a notification for "new" root dirs.

Arksine:
Ah, currently I only send root_update for the mutable roots (gcodes and config).  However its not a problem to send it for the static directories as well.

meteyou:
I was surprised, that gcodes is every time in the registered directories. I stopped Klipper and restarted moonraker and checked /server/info. I would have expected that gcodes would only be added with the start of Klipper with virtual_sdcard. Was something wrong on my test?

Arksine:
Just pushed the change, you should be notified for the static directories now.

With regard to the `gcodes` root, there is nothing wrong with your test.  I save the last known `gcodes` root path in the database and load it immediately when Moonraker starts.  I did this to facilitate immediate access to the gcode files.

meteyou:
Ah. Ok thx! I will test it!

meteyou:
@Arksine thx! notification works fine.

PR #160, opened 2021-05-09
update mananger: conf sanity check for other envs

opcow:
use site module to find missing paths

Signed-off-by: Mitch Crane <mitch.crane@gmail.com>

I'm running Moonraker and Klipper in docker containers. I've built both Alpine-based and Debian-based containers and with both I get a plugin error displayed in the Fluidd web UI due to Moonraker not finding these paths. My solution was to use `site.getusersitepackages()` and `site.getsitepackages()` to find the proper paths.

The configured path is checked first and the configured value is accepted if the path is valid. If not `site.getsitepackages()` or `site.getusersitepackages()` is called. In the case of `site.getsitepackages()` which returns a list, a glob search for `gpio*` is done. If found then it is assumed the correct path has been found.

### DIFF
--- a/moonraker/components/update_manager.py
+++ b/moonraker/components/update_manager.py
@@ -15,6 +15,7 @@ import io
 import asyncio
 import time
 import tempfile
+import site
 import tornado.gen
 from tornado.ioloop import IOLoop, PeriodicCallback
 from tornado.httpclient import AsyncHTTPClient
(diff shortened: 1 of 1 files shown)


Arksine:
Thanks.  I'm not sure that falling back to the user site package directory is the preferable behavior.  Any application that needs to symlink a dist-package should do so in the app's virtualenv.

I find it curious that the site path is incorrect.  What is the correct path to the site packages in your venv?  Or are you not installing moonraker in a venv in your containers?

opcow:
I wasn't sure if this was good fix. I figured you would know better. Maybe
I should just symlink the right path?

I was using a virtual environment when I first discovered the issue. I
later started building without it. Using virtualenv or not on an alpine
based container /usr/lib/python3.8/site-packages is the right path. Using a
Debian container with no venv I get the following error:
confighelper.ConfigError: No match for 'env_package_path':
/usr/lib/python*/site-packages

I suppose that may be because I'm running without the virtual env, yet I
don't get that error on Alpine.

So I was thinking it would be useful to discover the correct paths at
runtime, but now I'm thinking I should do it when building the images and
create links.

Arksine:
For the time being I think it makes sense to focus support on virtualenv installs.  You can use symlinks to get around that if you desire a custom install.   In the future I plan to look into `pex` distributions which may make it moot.

That said, if you want to use virtualenvs and the site package path is incorrect when installed in your container then I would certianly be interested in supporting it.  If the path is not at `~/moonraker-env/lib/python*/site-packages`, then we just need to determine where the venv site packages are being installed.

opcow:
Yeah, makes sense. It was definitely wrong using Alpine images as my base
with a virtualenv install, but I think I'm going to move to Debian. Alpine
is nice because it's small and the package manager is really fast, but by
the time you add everything you need to make moonraker and klipper work it
bloats up a lot.

Arksine:
I'm going to go ahead and close this.  Thanks.

PR #165, opened 2021-05-19
app: Fixed download for the files with non-ASCII characters in name.

vladimir-poleh:
Replace non-ASCII characters with "?" and add escaped unicode version of the file name to the "Content-Disposition" header to fix problems with thumbnails and downloading for the gcode file with non-ASCII characters in filename.

"?" is not the best replace character as it is forbidden character for the filenames on Windows but it is simple and fast solution and anyway this fix is providing Unicode filename that should be used in the modern browsers.

### DIFF
--- a/moonraker/app.py
+++ b/moonraker/app.py
@@ -11,6 +11,7 @@ import logging
 import json
 import datetime
 import traceback
+import urllib.parse
 import tornado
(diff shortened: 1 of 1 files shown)


Arksine:
Looks good.  Thanks.

PR #171, opened 2021-05-29
Fix Home Assistant https support

aussiegeek:
My Home Assistant instance only uses https, and the existing code was hardcoded to http.

Added https support, and fixed a bug where it was expected Home Assistant would provide an updated status, but actually returns an empty body


Arksine:
Looks good, thanks!

PR #176, opened 2021-06-08
history: add support for cancelled state

ayufan:
Adds a support for `cancelled` state introduced by https://github.com/KevinOConnor/klipper/pull/4366.

Related:
- https://github.com/KevinOConnor/klipper/pull/4366

### DIFF
--- a/moonraker/components/history.py
+++ b/moonraker/components/history.py
@@ -188,7 +188,11 @@ class History:
                 elif self.current_job is not None:
                     if new_state == "complete":
                         self.finish_job("completed", new_ps)
-                    if new_state == "standby":
+                    elif new_state == "cancelled":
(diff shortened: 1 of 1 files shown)


ayufan:
@Arksine WDYT?

Arksine:
I'm fine with these changes, I'll merge when support has been added to klipper

ayufan:
@Arksine I believe it is better to merge it to Moonraker first, as merging into Klipper first will make Moonraker to be broken :)

Arksine:
That is fair.  Generally I think its best practice to merge changes as necessary rather than in anticipation of such, however there is no harm in merging this now as it seems like Kevin is ok with merging your changes in Klipper.

Issue #178, opened 2021-06-12
Unified camera support


ayufan:
## Problem

I'm following for some time the klipper ecosystem and see a few missing features:

- ability to have timelapse
- ability to have a picture of a print after the print completes
- a way to support RTSP camera (pretty common for Xiaomi camera using unofficial software)
- a way to have a consistent configuration of cameras across all frontends
- make Moonraker in some cases to produce a video stream/snapshot to use in Mainsail/Fluidd
- provide out of box experience that if camera is configured in Moonraker it just works, regardless what type of camera it is: it can be USB connected V4L, or RTSP, or a simple IP camera that outputs h264 stream
- Moonraker runs whatever is needed to provide output, which can be: `live555`, `mjpgstreamer`, or anything else that can transform output into desired form

Today:

- each frontend defines its own configuration of cameras
- the support of different camera differs between, and you have to configure it many times

### Camera configuration

```
[camera default]
type = mjpg / v4l / rtsp
snapshot_url = # used by mjpg
stream_url = # used by mjpg or rtsp
device = # used by v4l
fps = # can be provided in each case to transform the stream (?)
resolution = # can be provided in each case to transform the stream (?)
flip_horizontal = false | true
flip_vertical = false | true
proxy = false | true # indicates if the camera is accessed directly or via moonraker => a default is that Moonraker tries to expose direct access

[camera my_mjpg]
type = mjpg
snapshot_url = http://my-webcam/snapshot
# stream_url:
# set if available, then the adaptive `mjpg-streamer` is used
# that requests snapshots fps-times per second
stream_url = http://my-webcam/stream 
fps = 30

[camera my_v4l]
type = v4l
device = /dev/video0
resolution = 1920x1080

[camera my_rtsp]
type = rtsp
stream_url = rtsp://my-camera:12315/stream.ch
```

### Camera API

The frontend can request `/server/camera/list` that returns
all available cameras with different access patterns as possible,
so frontend can pick the most efficient one:

- if `h264` is available we should prefer to use mkv/mp4/ts container as this is least amount of data to transfer and can perform stream copy
- if `mjpg` is available we provide a stream of jpg frames
- we try our best to transform the stream into desired output
- the added benefit is that `octoprint_compact` can also benefit from the camera configuration of which a Cura can present `mjpg` stream
- 
```json
{
  "default": {
     "type": "mjpg/v4l/rtsp",
     "stream_video_url": "a url to a video output if available",
[...]

Issue #138, opened 2021-04-05
Implement notification component


ayufan:
I would like to write a notification component that would allow to trigger a set of actions on a different events happening in Klipper.

For that, I would like to:
- abstract out of history component the `print_stats` state tracking
- rework `history` to use a common base interface
- implement notification component that would implement actions similar to: `on_print_start/on_print_resume/pause`
- notification would have many pluggable backends (similar to power) allowing to use http or gcode macros initially
- the common base interface could also be used by another components, like timelapse to consistently track a print state

Is this a good direction?

ayufan:
@Arksine 

WDYT? Is this a good direction? I'm in particular asking about abstracting the state tracking which is likely to be reimplemented by many components even though it should be pretty common set of behaviors.

Arksine:
I think the general idea is a good one, I have plans to add support for MQTT and I could simply interface with your component rather than implement from scratch.

With regard to the interface, I would say that "it depends".  It might make sense to create an interface (or a helper) that tracks `print_stats` and reports various state transitions.   Beyond that, what each component does with this is going to be different.  The question  becomes as to whether or not there is enough code reuse to matter.

I think a good approach would be to create the interface, have your component use the interface, then at a later date we could look at modifying `history`.  I don't think we would want to do both in the same pull request.

ayufan:
@Arksine 

I was thinking more about 2-4 PRs :)

Arksine:
Since Moonraker now has a notifier component (along with MQTT) I think we can mark this issue as resolved.  That said, there is probably room to expand the notifier to support more events.

PR #179, opened 2021-06-12
components: Add `StateComponent` that can hook to `print_stats`

ayufan:
It is pretty common to hook to print state change.
This generalises this to ensure a consistent behavior
across the `Moonraker`.

Related to:
- https://github.com/Arksine/moonraker/issues/138
- https://github.com/Arksine/moonraker/issues/178

Based on:
- https://github.com/Arksine/moonraker/pull/176

Remarks:
- This is was moderately tested
- This is needed to implement a consistent way to track state change, implement additional components that hook on it, like: `history_snapshot` or `timelapse` without re-inventing a wheel for each case
- This also can recover a print that was started before Moonraker was run: it will fetch data and properly recover the state

### DIFF
--- a/moonraker/components/__init__.py
+++ b/moonraker/components/__init__.py
@@ -4,3 +4,127 @@
 # Copyright (C) 2020 Eric Callahan <arksine.code@gmail.com>
 #
 # This file may be distributed under the terms of the GNU GPLv3 license.
+
+from __future__ import annotations
+import logging
+
+# Annotation imports
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Union,
+    Optional,
+    Dict,
+    List,
+)
+if TYPE_CHECKING:
+    from confighelper import ConfigHelper
+    from components.klippy_apis import KlippyAPI
+
+PRINTING_STATE = ["printing", "paused"]
+FINISHED_STATE = ["standby", "complete", "error", "cancelled"]
+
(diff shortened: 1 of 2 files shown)


Arksine:
I'm not sure that this exact pattern will be that common, so for the moment I don't think that this change is necessary.  If we get to a point where the same pattern is repeated multiple times I would consider breaking it out into another module, however I don't think it should be in the package definition.

ayufan:
@Arksine I actually have in plans the mentioned ones. Looking through code there's a lot of duplication of handling `print_stats` already.

Do you have suggestion how this should look like differently?

Arksine:
In that case I think the better solution would be to submit the "StateComponent" at the time you want to PR that code, then afterward we can make changes to `history` in a separate pull request.

One thing I would do differently is implement it as a core component rather than a base class.  It could send server events for each state change.  Consumers could register those events and handle them appropriately.  Each Consumer should probably register their own on_exit() method and handle this appropriately.

Arksine:
PR #265 added a `job_state` component that emits events for changes in print state.   It should detect and emit the following events:
- `job_state:started`
- `job_state:paused`
- `job_state:resumed`
- `job_state:cancelled`
- `job_state:standby`
- `job_state:error`
- `job_state:complete`

The `history` component was refactored to use this, as does the new `job_queue` component.

PR #182, opened 2021-06-15
pause_resume: Make Moonraker to use `pause_resume/*` API

ayufan:
This switches Moonraker from calling `CANCEL_PRINT` and alikes
to rather use `pause_resume/` API that underneath calls
the relevant G-Codes.

This done, this way, allows Klipper to be able to interrupt
blocking operations gracefully.

Related:
- https://github.com/KevinOConnor/klipper/pull/4398
- This makes `Cancel of a print` (when called via API) to be immediate when during a heating phase

### DIFF
--- a/moonraker/components/klippy_apis.py
+++ b/moonraker/components/klippy_apis.py
@@ -60,13 +60,13 @@ class KlippyAPI(Subscribable):
             "/printer/firmware_restart", ['POST'], self._gcode_firmware_restart)
 
     async def _gcode_pause(self, web_request: WebRequest) -> str:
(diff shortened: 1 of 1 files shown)


Arksine:
Thanks.  I'm ok with calling the `pause_resume` endpoints here, however I don't think its necessary to check if the module has been loaded.  We can simply let it 404 rather than attempt to call the corresponding gcode, since those gcodes won't be available if the module isn't loaded.

ayufan:
@Arksine Updated

Arksine:
Looks good, thanks.

PR #184, opened 2021-06-17
MQTT notifications

jordanruthe:


jordanruthe:
This will do the following:

Send state change events and post state changes to
instance_name/moonraker/printer/state

Send file load/unload events and post file "loaded fname.gcode" and unloaded to
instsance_name/moonraker/printer/file

I know this isn't comprehensive and doesn't allow customization of notifications for mqtt, but I'm thinking this may be a good place to start.

Arksine:
I definitely think we are getting closer.  There are some parts of this pull request that will overlap some work I'm currently doing.  I'm implementing a "job queue" that will allow users to queue jobs up even if the printer is not available.  If configured, the queue will start when Klipper is ready.  This requires monitoring print state, so I broke out that functionality into its own module, `print_state.py`.  It operates in a similar fashion to what you have in `notification.py` in that it emits server events for a state change and will be a core module.  Likewise I made changes to history.py to accommodate this.  I anticipate that it will be complete soon.

I do think that we will want users to be able to configure topics.  We should probably discuss exactly how we want to handle configuration.  I was thinking that we could have a section for each notification.  The drawback is that it would require more configuration for users to get *all* notifications, however the upside is that it would provide the ability to fune tune exactly what notifications are sent and how they should behave.  It also might allow us to provide a way for users to create custom notifications, I haven't thought far enough ahead on that.  I would likely need to consider implementing 'include' functionality similar to what Klipper does to allow users to better organize the config.

Arksine:
Oh, one other thing.  I noticed that we forgot to add your name to the copyright header in `history.py`.  If you want to add it you could do in another PR and I would merge immediately.

pedrolamas:
This is a great start for MQTT and I look forward in being able to use this! 🙂

If I may, the only further improvement that I do believe it should be prioritized on this is to expose the `device_power` elements so we can get status updates on them.

Arksine:
Since we have MQTT support and separate notification support I think we can close this.  Thanks.

PR #202, opened 2021-07-30
mqtt: publish print_stats changes

pedrolamas:
Publishes print_stats changes to the MQTT topic "<instance_name>/moonraker/status/print_stats"

### DIFF
--- a/moonraker/components/mqtt.py
+++ b/moonraker/components/mqtt.py
@@ -31,8 +31,10 @@ from typing import (
 if TYPE_CHECKING:
     from app import APIDefinition
     from confighelper import ConfigHelper
+    from . import klippy_apis
     FlexCallback = Callable[[bytes], Optional[Coroutine]]
     RPCCallback = Callable[..., Coroutine]
+    APIComp = klippy_apis.KlippyAPI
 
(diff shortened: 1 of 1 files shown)


pedrolamas:
Hum... this seems to be a bit "chatty" when printing, as it will send updates for `*_duration` fields...

I'm thinking in making a small change that would push notifications immediately if the `state` or `message` changes, but any other attribute would use a timed update (once every 3 seconds? maybe even configurable?)

Arksine:
There are plans to add something like this in the future through a "notification" component, which would allow users to configure specific notification types to be pushed over a selected protocol (mqtt, http, email, etc).  It would be less verbose than publishing status updates.

The `mqtt` module is designed so that other components may look it up to publish/subscribe to topics, so generally speaking we don't want print specific functionality in there.  That said, they may be some use to allowing users to configure subscriptions to be pushed over a topic, however I think we would want to make it configurable and not limited to `print_stats`.

pedrolamas:
Thank you for your comments @Arksine, it makes complete sense, and given that, I think I will close this now.

Is the existing "notifications" PR still under development? Seems a bit stall right now and I noticed it is using a different type of MQTT components compared to the existing module.

PR #204, opened 2021-08-02
Add `on_when_upload:` to `power`

ayufan:
This makes a printer to auto power on when upload
happens on all plugs requested.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -201,6 +201,9 @@ type: gpio
 #   The type of device.  Can be either gpio, tplink_smartplug, tasmota
 #   shelly, homeseer, homeassistant, or loxonev1.
 #   This parameter must be provided.
+on_when_upload: False
+#   If set to True, Moonraker will power on the device when a upload happens.
+#   The default is False, thus no attempt to start printer is being made.
 off_when_shutdown: False
 #   If set to True the device will be powered off when Klipper enters
 #   the "shutdown" state.  This option applies to all device types.
--- a/docs/web_api.md
+++ b/docs/web_api.md
@@ -522,6 +522,7 @@ included.
         "power green_led": {
             "type": "gpio",
             "locked_while_printing": false,
+            "on_when_upload": false,
             "off_when_shutdown": false,
             "restart_klipper_when_powered": false,
             "pin": "gpiochip0/gpio26",
--- a/moonraker/components/file_manager/file_manager.py
+++ b/moonraker/components/file_manager/file_manager.py
@@ -37,9 +37,11 @@ if TYPE_CHECKING:
     from websockets import WebRequest
     from components import database
     from components import klippy_apis
+    from components import power
     from components import shell_command
     DBComp = database.MoonrakerDatabase
     APIComp = klippy_apis.KlippyAPI
+    PowerComp = Optional[power.PrinterPower]
     SCMDComp = shell_command.ShellCommandFactory
     _T = TypeVar("_T")
 
@@ -499,13 +501,7 @@ class FileManager:
         await self.gcode_metadata.parse_metadata(
             upload_info['filename'], finfo).wait()
         if start_print:
-            # Make a Klippy Request to "Start Print"
-            kapis: APIComp = self.server.lookup_component('klippy_apis')
-            try:
-                await kapis.start_print(upload_info['filename'])
-            except self.server.error:
-                # Attempt to start print failed
-                start_print = False
+            start_print = await self._start_print(upload_info['filename'])
         await self.notify_sync_lock.wait(300.)
         self.notify_sync_lock = None
         return {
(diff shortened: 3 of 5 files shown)


Arksine:
Thanks.  I know that this has been requested previously, however I dont think this that this is a good idea as there are a lot of edge cases where things could go wrong making this option unreliable.  Whilewe can work around that by gracefully failing, the result could be several "why isn't it working" help requests.  IMO it isn't a big ask to have users power on their printer manually.

ayufan:
> There are a lot of edge cases where things could go wrong making this option unreliable

Yes, this is likely. I don't know if this catches all quirks of detecting klipper state, since there's a lot of async operations happening.

> IMO it isn't a big ask to have users power on their printer manually.

Yes, but this is quite inconvenience. Especially if you have smart plug that has hard to reach power button. This is probably the most annoying difference that I find between Moonraker and Octoprint, where Octoprint plugs can easily power on printer, and printer than can automatically power off.

I have been happily using this since it was published, and this at least for me have been a big UX improvement :)

Arksine:
> Yes, but this is quite inconvenience. Especially if you have smart plug that has hard to reach power button. This is probably the most annoying difference that I find between Moonraker and Octoprint, where Octoprint plugs can easily power on printer, and printer than can automatically power off.

By "manually" I mean using a client UI to power on the switch.  Both Mainsail and Fluidd should have this functionality.  AFAIK this is not much different from how Octoprint handles it.

If the goal is to simply toggle the switch on when a gcode file is uploaded I could get behind that.  We could do that using the the event system (the file manager emits a "create_file" event).  Users would have to start the print after they verify that Klipper has successfully started up.  This may be of slight inconvenience but it would be the most reliable way of attempting this.

ayufan:
> Both Mainsail and Fluidd should have this functionality. AFAIK this is not much different from how Octoprint handles it.

It is, the difference is actually significantly bigger :)

The current workflow with `fluidd/mainsail`:

1. Go to the Web, find a power switch that is behind a drop down, click it. This alone requires at least one more click to open dropdown, but also a few seconds for interface to load. This is definitely something that can in general be improved: the accessibility of the button.
2. Go to Cura or PrusaSlicer and upload gcode

The Octoprint workflow (with auto start):

1. Configure once `tplink smartplug` to power on the upload event with a `start_print` ([On with Upload](https://github.com/jneilliii/OctoPrint-TPLinkSmartplug))
2. It just works with Cura or PrusaSlicer

The Octoprint workflow (without auto start, this is simular to current workflow with `fluidd/mainsail`):

1. Go to Octoprint, the power plug button is on top bar. The UX is comparable, because Octoprint takes longer to load than `fluidd/Mainsail`
2. Use Cura or PrusaSlicer

> If the goal is to simply toggle the switch on when a gcode file is uploaded I could get behind that
> This may be of slight inconvenience but it would be the most reliable way of attempting this

Can you describe more that proposal? If this still requires opening Web to accept file and auto start it, this is still not great experience.

Maybe there are cases that you need to inspect print each time before start of the print (this is also this is configurable, and not on by default), but in my case where printer is behind me it is good enough to just turn around head. I can hear and see if printer started, klipper is running and whether print did start.

I think if the concern is reliability we can figure out how to do this reliable. As I noticed the async processing of states is to some extent a little twisted, but we can definitely figure out how to deal with that. The question is do you want to proceed this path?

Arksine:
With regard to the octoprint plugin, you can be sure that the "on with upload" functionality won't work for some Klipper installations. That is my area of concern. The behavior in which Klipper fails to start is not well defined.

That said, I think there may be a way to accomplish your goals.  First, as hinted at above, I don't think its a good idea to attempt to tie the power module into the file manager.  Instead the power module should register the `file_manager:create_file` event.  In the handler, it should make sure that the root is the gcodes root and the extension is a valid gcode file.  If so, it could iterate through devices and power on any device configured with `on_with_upload`.

Once Klipper restarts the user would then need to start the print.  However, it is possible to automate this through the use of a "job queue", which is something I am working on.  I can modify the upload handler to queue files tagged with "start", and the queue can start the print when the printer is ready.

ayufan:
> First, as hinted at above, I don't think its a good idea to attempt to tie the power module into the file manager

Thanks @Arksine. Yes, I'm aware of that, but this is approach was taken due to current limitations.
I was actually thinking about something similar to what you describe, but came across a few shortcomings:

1. I initially tried to add `file_manager:print_on_upload` that would be fired an appropriate flag is specified
2. Power adapters would hook to this event and execute set of operations
3. The shortcomings was that there was no queue of commands, instead `start_print` is executed synchronously

What do you think about the following:

1. Since you work on something that resembles job queue to execute (on `start/being-connected` whatever it is :)), this would be pretty handy. Ideally if we could somehow specify timeout to ensure that it does not linger too long in a queue this would be great
1. I would introduce an additional hook `file_manager:print_on_upload` (or `file_manager:create_file` [which is better?] with a flag to indicate `start_print` request?) that would fire when `start_print:` is passed
1. The `power` would hook and look at `on_with_upload` flag, then for each plug it would power on and execute `firmware_restart` as this is done today
1. We would change current job handler to use this job queue

Arksine:
FYI, support for this was added in PR #265 with the job queue component.  The following conditions are necessary for this to work:

- The `[job_queue]` must be configured with `load_on_startup: True`
- The `[file_manager]` must be configured with `queue_gcode_uploads: True`
- Your power device must be configured with `on_when_upload_queued: True`
- Your upload (slicer, Fluidd, etc) must have the `start` flag set

The [documentation](https://moonraker.readthedocs.io/en/latest/configuration/#power-on-g-code-uploads) contains an example.

While this did add some additional configuration, it was necessary for maximum flexibility.  For example, a user can still take advantage of queued uploads if they don't have a power device configured.

ayufan:
Wow. Amazing job Eric. Thank you :)

PR #213, opened 2021-08-20
Fixes rounding issue in home assistant sensor

pedrolamas:
There was a minor issue with the Home Assistant template related to the priority of Jinja expressions (filters seem to take precedence over operations), so it would show "progress: 1.234567%" instead of "progress: 1.2%".


Arksine:
Thanks.

PR #223, opened 2021-09-05
server: report deprecated klipper options 

Arksine:
This adds a warning for deprecated options in Klipper to the list of warnings reported in `/server/info`.  The configuration warnings generated by Moonraker have been modified for additional clarity.

Signed-off-by:  Eric Callahan <arksine.code@gmail.com>

### DIFF
--- a/moonraker/confighelper.py
+++ b/moonraker/confighelper.py
@@ -6,6 +6,7 @@
 
 from __future__ import annotations
 import configparser
+from logging import warning
 import os
 from utils import SentinelClass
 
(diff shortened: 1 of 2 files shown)


meteyou:
for me it is not necessary that moonraker pull this output and passes it. because it is also available via `printer.configfile.warnings`.

Arksine:
Ok.  I think that is probably the best way to handle it, that way the frontend can make a better decision of how to present it.

What I will do is make sure it is clear that moonraker's config warnings specify that the issue is in moonraker.conf.

meteyou:
yep! thats the reason, why i will use two different panels for moonraker and klipper warnings and different buttons for download moonraker log for moonraker warnings and download klippy.log for klipper warnings and more...

Arksine:
I'm gonna go ahead and close this since we have decided that its better to allow clients handle the warnings directly.

PR #226, opened 2021-09-12
Allow reset of history totals

jordanruthe:
Signed-off-by: Jordan <jordan.ruthe@gmail.com>

### DIFF
--- a/moonraker/components/history.py
+++ b/moonraker/components/history.py
@@ -61,7 +61,9 @@ class History:
         self.server.register_endpoint(
             "/server/history/list", ['GET'], self._handle_jobs_list)
         self.server.register_endpoint(
-            "/server/history/totals", ['GET', 'DELETE'],
+            "/server/history/totals", ['GET'], self._handle_job_totals)
+        self.server.register_endpoint(
+            "/server/history/delete_totals", ['DELETE'],
             self._handle_job_totals)
 
         database.register_local_namespace(HIST_NAMESPACE)
@@ -174,15 +176,17 @@ class History:
                                  ) -> Dict[str, Union[bool, Dict[str, float]]]:
         action = web_request.get_action()
         if action == "DELETE":
-            for x in self.job_totals:
-                if isinstance(self.job_totals[x], float):
-                    self.job_totals[x] = float(0.)
-                else:
-                    self.job_totals[x] = int(0)
+            self.job_totals = {
+                'total_jobs': 0,
+                'total_time': 0.,
+                'total_print_time': 0.,
+                'total_filament_used': 0.,
+                'longest_job': 0.,
+                'longest_print': 0.
+            }
             database: DBComp = self.server.lookup_component("database")
             database.insert_item(
                 "moonraker", "history.job_totals", self.job_totals)
-            return {"success": True}
 
         # Action is GET
         return {'job_totals': self.job_totals}
--- a/moonraker/components/history.py
+++ b/moonraker/components/history.py
@@ -61,7 +61,8 @@ class History:
         self.server.register_endpoint(
             "/server/history/list", ['GET'], self._handle_jobs_list)
         self.server.register_endpoint(
-            "/server/history/totals", ['GET'], self._handle_job_totals)
+            "/server/history/totals", ['GET', 'DELETE'],
+            self._handle_job_totals)
 
         database.register_local_namespace(HIST_NAMESPACE)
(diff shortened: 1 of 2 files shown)


moonraker/components/history.py
```diff
@@ -170,7 +171,20 @@ async def _handle_jobs_list(self,
...
                                  web_request: WebRequest
-                                 ) -> Dict[str, Dict[str, float]]:
+                                 ) -> Dict[str, Union[bool, Dict[str, float]]]:
+        action = web_request.get_action()
+        if action == "DELETE":
+            for x in self.job_totals:
+                if isinstance(self.job_totals[x], float):
+                    self.job_totals[x] = float(0.)
+                else:
+                    self.job_totals[x] = int(0)
```
Arksine, line 181: I think rather than use a for loop here it might be more clear to re-initialize self.job_totals, ie:
```python
self.job_totals = {
    'total_jobs': 0,
    'total_time': 0.,
    'total_print_time': 0.,
    'total_filament_used': 0.,
    'longest_job': 0.,
    'longest_print': 0.
}
```

moonraker/components/history.py
```diff
@@ -170,7 +171,20 @@ async def _handle_jobs_list(self,
...
+            database: DBComp = self.server.lookup_component("database")
+            database.insert_item(
+                "moonraker", "history.job_totals", self.job_totals)
+            return {"success": True}
```
Arksine, line 185: I would just use the same return value as the `GET` method here, essentially just the empty job totals.

Arksine (commented):
Thanks, I added a couple of review comments.

I think this functionality is useful and adding a DELETE method is the correct approach, however it will break the existing websocket API (`server.history.totals` will become `server.history.get_totals`).  So we'll have to make an announcement and give front end devs a chance to update.  Also the documentation needs to be updated to reflect the change and the new `delete_totals` API.

Also I don't think we need the `check_code.sh` script, that should probably be part of your dev enviroment and added to your global gitignore.

jordanruthe:
I'm torn, I could make a new endpoint `/server/history/delete_totals` which would translate to the same `server.history.delete_totals` websocket method. I did confirm that both fluidd and mainsail currently use `server.history.totals`. What are your thoughts?

Arksine:
I don't think we should create a new endpoint, the original endpoint registration was the correct way of handling it IMO.  I understand the concern about breaking the API, however if we are going to have multiple endpoints for totals then `server.history.get_totals` is more clear than `server.history.totals`.

Arksine:
Apparently I implemented this as a `reset` command in my job queue branch.  Last month has been full, I forget the things I have been working on:
```diff
diff --git a/moonraker/components/history.py b/moonraker/components/history.py
index 123f4a1..35dbc26 100644
--- a/moonraker/components/history.py
+++ b/moonraker/components/history.py
@@ -66,6 +66,9 @@ class History:
             "/server/history/list", ['GET'], self._handle_jobs_list)
         self.server.register_endpoint(
             "/server/history/totals", ['GET'], self._handle_job_totals)
+        self.server.register_endpoint(
+            "/server/history/reset_totals", ['POST'],
+            self._handle_job_total_reset)
 
         database.register_local_namespace(HIST_NAMESPACE)
         self.history_ns = database.wrap_namespace(HIST_NAMESPACE,
@@ -167,6 +170,26 @@ class History:
                                  ) -> Dict[str, Dict[str, float]]:
         return {'job_totals': self.job_totals}
 
+    async def _handle_job_total_reset(self,
+                                      web_request: WebRequest,
+                                      ) -> Dict[str, Dict[str, float]]:
+        if self.current_job is not None:
+            raise self.server.error(
+                "Job in progress, cannot reset totals")
+        last_totals = dict(self.job_totals)
+        self.job_totals = {
+            'total_jobs': 0,
+            'total_time': 0.,
+            'total_print_time': 0.,
+            'total_filament_used': 0.,
+            'longest_job': 0.,
+            'longest_print': 0.
+        }
+        database: DBComp = self.server.lookup_component("database")
+        database.insert_item(
+            "moonraker", "history.job_totals", self.job_totals)
+        return {'last_totals': last_totals}
+
     def _handle_print_started(self,
                               prev_stats: Dict[str, Any],
                               new_stats: Dict[str, Any]
```
So I'm working on this branch now, when I merge it would create a `reset_totals` which is a POST request.  This has the advantage of not breaking the API, and may be a better match for what we are trying to accomplish.

jordanruthe:
Sounds good, looks like you've got it covered.

PR #232, opened 2021-09-26
Adding support for 433Mhz RF transmitter as power devices

danijoo:
This adds support for 433Mhz RF transmitters (i.e. for wireless sockets like [this](https://de.aliexpress.com/item/1005002291642865.html) one) to switch on/off devices.

 It duplicates some code from the gpio device to configure the pin. This could be done in a base class instead to remove the code duplication. Let me know if this is desired.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -190,7 +190,7 @@ gcode:
 ```
 
 ## `[power]`
-Enables device power control.  Currently GPIO (relays), TPLink Smartplug,
+Enables device power control.  Currently GPIO (relays), RF transmitter, TPLink Smartplug,
 and Tasmota (via http) devices, HomeAssistant switch are supported.
 
 ```ini
@@ -198,7 +198,7 @@ and Tasmota (via http) devices, HomeAssistant switch are supported.
 
 [power device_name]
 type: gpio
-#   The type of device.  Can be either gpio, tplink_smartplug, tasmota
+#   The type of device.  Can be either gpio, rf, tplink_smartplug, tasmota
 #   shelly, homeseer, homeassistant, or loxonev1.
 #   This parameter must be provided.
 off_when_shutdown: False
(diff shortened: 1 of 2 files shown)


Arksine:
Thanks.  All commits need a signed off line, see [contributing.md](https://moonraker.readthedocs.io/en/latest/contributing/) for details.

With regard to the implementation, there are a few details that stand out to me:
- As you said, it duplicates some GpioDevice code.  It could probably just inherit from GpioDevice and override methods as necessary.
- We can't directly call time.sleep() in Moonraker as it would block the asyncio event loop. You could try using asyncio.sleep(), or
  if necessary you could do the blocking calls a threadpoolexecutor. (Moonraker's eventloop object provides a `run_in_thread` convenience method).

danijoo:
I squashed the commits together, added a signed-off line and addressed some of your comments (constants, base class). However, I'm having a hard time getting asyncio to work. RF devices depend on the exact timing on the sent waveform, so I think `asyncio.sleep` is messing with the timing too much - at least it's not working for me if I replace `time.sleep` with `async.sleep` and call the methods with `async def`+`await`. On the other hand, I failed to get `event_loop.run_in_thread` to work properly.

Arksine:
The changes look good so I'll go ahead and merge this.  The time spent blocking looks very short so it won't likely cause any issues for the time being, I'll make an attempt to get run_in_thread working and PR it for you to test.

As an aside, I think perhaps a specific document outlining how to use this might be useful (ie specifically what devices it is compatible with, how to wire a device, and how to determine what codes to use).  I suspect that there may be users interested in something like this but wont know how to use it.

danijoo:
> As an aside, I think perhaps a specific document outlining how to use this might be useful (ie specifically what devices it is compatible with, how to wire a device, and how to determine what codes to use). I suspect that there may be users interested in something like this but wont know how to use it.

Sure! I will prepare something for this and open another PR.

PR #233, opened 2021-09-26
metadata: add SliCR-3D support (fork of SuperSlicer)

meteyou:
this PR adds SliCR-3D to the metadata export. It is a SuperSlicer fork.

Here you can find more informations about it: https://github.com/CR-3D/SliCR-3D

Signed-off-by: Stefan Dej <meteyou@gmail.com>

### DIFF
--- a/moonraker/components/file_manager/metadata.py
+++ b/moonraker/components/file_manager/metadata.py
@@ -339,6 +339,16 @@ class SuperSlicer(PrusaSlicer):
             }
         return None
 
+class SliCR3D(SuperSlicer):
+    def check_identity(self, data: str) -> Optional[Dict[str, str]]:
+        match = re.search(r"SliCR-3D\s(.*)\son", data)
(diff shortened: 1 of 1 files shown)


ChristianReil:
We are manufacturers of industrial 3D printers and are already very established in the German market ... we will of course continue to pursue the topic of SliCR-3D, as it is the right addition to our printers and we need this for our easy integration of slicer, printer and filament. Currently it is still a pure "ReBrand" ... but depending on how strong our community is, we try to add our own functions to the slicer ... 🙂

 We just got a lot of feedback that many user wanted to use it with moonraker.
🙂 Thank you very much for the support!

Arksine:
I added support for this via aliases in commit fdfab669516b6a8ccf02dff1466126ff64d5510a.  Should we need to support any more forks of PrusaSlicer in the future it will make it a little bit easier to add.

PR #234, opened 2021-09-29
Nginx install option

pandel:
Additional command line option for the install script to not install Nginx per default. If not given, Nginx will be installed as usual.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -131,6 +131,10 @@ particularly for those upgrading:
   Specifies the path to Moonraker's config file. The default location
   is `/home/<user>/moonraker.conf`.  When using this option to modify
   an existing installation it is necessary to add `-f` as well.
+- `-n`:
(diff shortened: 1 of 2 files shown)


Arksine:
Thanks.  There is one issue with this approach, Moonraker's update manager parses this script to look for changes in dependencies.  It would always pick up nginx and install it since its in the script.

I'm inclined to simply remove nginx as a dependency.  We could simply have Mainsail and Fluidd add its installation as an additional step.  Users could choose to use haproxy, apache, etc.

pandel:
Ok, I did not know that. So simply decline my PR. And, from a general point of view, I think you are going the right direction when removing the Nginx dependency. In my opinion every tool has to care for its own dependencies and so have Mainsail or Fluidd. As Moonraker does not need it, it simply should not install it :-)

Arksine:
Yeah, its kind of sneaky.  I went ahead and removed it, this should resolve #234.

PR #251, opened 2021-10-23
Richardjm/wled

richardjm:
Attempted integration of wled functionality into moonraker. Based largely of existing code trying to do "when in Rome".

Added documenation for GCODE methods
Not sure about documentation of very simple WebApi methods

Roughly tested on my voron and it all seems to replace the functionality from the klipper module I created.
You can see my klipper config at https://github.com/richardjm/voronpi-klipper-backup

### DIFF
--- a/moonraker/components/wled.py
+++ b/moonraker/components/wled.py
@@ -191,7 +191,7 @@ class Strip:
                     {"seg": {"i": [0, self.chain_count-1, elem]}})
             elif self.send_full_chain_data:
                 # Send a full set of color data (e.g. previous preset)
-                state = {"seg": {"i": []}}
+                state: Dict[str, Any] = {"seg": {"i": []}}
                 cdata = []
                 for i in range(self.chain_count):
                     idx = i * elem_size
@@ -351,15 +351,14 @@ class WLED:
         else:
             await self.strips[strip].wled_on(preset)
 
-    # Individual pixel control, due to non-transmission this is kept as a
-    # separate call
+    # Individual pixel control, for compatibility with SET_LED
     async def set_wled(self: WLED,
                        strip: str,
                        red: float = 0.,
                        green: float = 0.,
                        blue: float = 0.,
                        white: float = 0.,
-                       index: int = None,
+                       index: int = -1,
                        transmit: int = 1) -> None:
         if strip not in self.strips:
             logging.info(f"Unknown WLED strip: {strip}")
@@ -411,14 +410,14 @@ class WLED:
             if strip is not None:
                 result[name] = await self._process_request(strip, req, -1)
             else:
-                result[name] = "strip_not_found"
+                result[name] = {"error": "strip_not_found"}
         return result
 
     async def _process_request(self: WLED,
                                strip: Strip,
                                req: str,
                                preset: int
-                               ) -> str:
+                               ) -> Dict[str, Any]:
         strip_onoff = strip.onoff
 
         if req == "status":
@@ -434,7 +433,7 @@ class WLED:
             else:
                 strip_onoff = OnOff.off
                 await strip.wled_off()
-            return strip_onoff
+            return strip.get_strip_info()
 
         raise self.server.error(f"Unsupported wled request: {req}")
 
--- a/moonraker/components/wled.py
+++ b/moonraker/components/wled.py
@@ -45,14 +45,14 @@ class ColorOrder(str, Enum):
         return 4
 
 class OnOff(str, Enum):
-    on: str ="on"
+    on: str = "on"
     off: str = "off"
 
 class Strip:
     def __init__(self,
-                name: str,
-                color_order: ColorOrder,
-                cfg: ConfigHelper):
+                 name: str,
+                 color_order: ColorOrder,
+                 cfg: ConfigHelper):
         self.server = cfg.get_server()
         self.client = AsyncHTTPClient()
         self.request_mutex = asyncio.Lock()
@@ -89,7 +89,7 @@ class Strip:
             "chain_count": self.chain_count,
             "preset": self.preset,
             "color_order": self.color_order,
-            "error" : self.error_state
+            "error": self.error_state
         }
 
     async def initialize(self):
@@ -98,15 +98,15 @@ class Strip:
         self.preset = self.initial_preset
         if self.initial_preset >= 0:
             self._update_color_data(self.initial_red,
-                        self.initial_green,
-                        self.initial_blue,
-                        self.initial_white)
+                                    self.initial_green,
+                                    self.initial_blue,
+                                    self.initial_white)
             await self.wled_on(self.initial_preset)
         else:
             await self.set_wled(self.initial_red,
-                               self.initial_green,
-                               self.initial_blue,
-                               self.initial_white)
+                                self.initial_green,
+                                self.initial_blue,
+                                self.initial_white)
 
     def _update_color_data(self, red, green, blue, white, index=None):
         red = int(red * 255. + .5)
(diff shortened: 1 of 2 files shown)


moonraker/components/wled.py
```diff
@@ -0,0 +1,502 @@
...
+                                    self.initial_blue,
+                                    self.initial_white)
+            await self.wled_on(self.initial_preset)
+        else:
+            await self.set_wled(self.initial_red,
+                                self.initial_green,
+                                self.initial_blue,
+                                self.initial_white)
+
+    def _update_color_data(self, red, green, blue, white, index=None):
```
Arksine, line 113: The arguments and return value here should be annotated.

moonraker/components/wled.py
```diff
@@ -0,0 +1,502 @@
...
+            led_data = [red, green, blue, white]
+
+        if index is None:
+            self._chain_data[:] = led_data * self.chain_count
+        else:
+            elem_size = len(led_data)
+            self._chain_data[(index-1)*elem_size:index*elem_size] = led_data
+
+    async def _send_wled_command(self,
+                                 state: Dict[str, Any]):
```
Arksine, line 130: Return value needs an annotation.

moonraker/components/wled.py
```diff
@@ -0,0 +1,502 @@
...
+                    self.wled_palettes = [p.lower() for p in data['palettes']]
+
+                self.error_state = None
+            except Exception as e:
+                msg = f"WLED: Error {e}"
+                self.error_state = msg
+                logging.exception(msg)
+                raise self.server.error(msg)
+
+    async def wled_on(self, preset):
```
Arksine, line 168: The argument and return value needs an annotation.

moonraker/components/wled.py
```diff
@@ -0,0 +1,502 @@
...
+    # Lower level api control - https://kno.wled.ge/interfaces/json-api/
+    async def wled(self,
+                   bri: int = -1,
+                   transition: int = -1,
+                   fx: int = -1,
+                   effect: str = '',
+                   sx: int = -1,
+                   ix: int = -1,
+                   pal: int = -1,
+                   palette: str = ''):
```
Arksine, line 188: The return value should be annotated.

moonraker/components/wled.py
```diff
@@ -0,0 +1,502 @@
...
+                    pal = self.wled_palettes.index(palette.lower())
+                except ValueError:
+                    logging.info(f'WLED: Palette not found: {palette}')
+            if pal >= 0:
+                seg['pal'] = pal
+            await self._send_wled_command(command)
+        except Exception as e:
+            logging.exception(e)
+
+    async def wled_off(self):
```
Arksine, line 224: The return value should be annotated.

moonraker/components/wled.py
```diff
@@ -0,0 +1,502 @@
...
+        logging.debug(f"WLED: off {self.name}")
+        self.onoff = OnOff.off
+        await self._send_wled_command({"on": False})
+
+    async def set_wled(self, red, green, blue, white,
+                       index=None, transmit=1):
```
Arksine, line 230: Needs annotations.

moonraker/components/wled.py
```diff
@@ -0,0 +1,502 @@
...
+
+    async def wled_off(self, strip: str) -> None:
+        if strip not in self.strips:
+            logging.info(f"Unknown WLED strip: {strip}")
+            return
+        await self.strips[strip].wled_off()
+
+    async def set_wled(self, strip: str,
+                       red=0., green=0., blue=0., white=0.,
+                       index=None, transmit=1) -> None:
```
Arksine, line 403: Needs annotations.

moonraker/components/wled.py
```diff
@@ -0,0 +1,502 @@
...
+
+            # Register methods for GCODE remote methods
+            self.server.register_remote_method(
+                "wled_on", self.wled_on)
+            self.server.register_remote_method(
+                "wled", self.wled)
+            self.server.register_remote_method(
+                "wled_off", self.wled_off)
+            self.server.register_remote_method(
+                "set_wled", self.set_wled)
```
Arksine, line 318: As mentioned on Discord, I think that we need to reduce the number of remote methods registered.  The `set_led` method seems ok, it accepts parameters matching that of the SET_LED gcode.

I would recommend combining `wled_on` and `wled_off` into a single `set_wled_state()` that takes a `strip` and `state` argument.  It seems that a `preset` argument could be applied if present here.

The `wled()` method seems too low level to expose to Klipper via remote method.  Given that the wled supports presets, my suggestion would be to rely on it and expose the preset argument as mentioned above.

Arksine (commented):
Thanks, I have added some initial review comments.

As mentioned on discord, I am currently working on a set of patches that will affect this module.  Specifically, it changes how Moonraker handles initialization.  It removes the `load_component_multi` method as it isn't necessary to distinguish between multi-instance components in Moonraker.  It also requires a `component_init` method for components that require async initialization.  I'll have more details on both after I push the changes.

moonraker/components/wled.py
richardjm: OK sounds good, I'll back out the lower level controls, thanks for the feedback.

moonraker/components/wled.py
richardjm: Code removed

moonraker/components/wled.py
richardjm: Believed fixed

moonraker/components/wled.py
richardjm: Believed fixed

moonraker/components/wled.py
richardjm: Believed fixed

moonraker/components/wled.py
richardjm: Believed fixed

moonraker/components/wled.py
richardjm: Believed fixed

moonraker/components/wled.py
richardjm: Believed fixed

moonraker/components/wled.py
richardjm: Tried to make those changes now

moonraker/components/wled.py
```diff
@@ -0,0 +1,445 @@
...
+                color_order = color_orders.get(color_order_cfg)
+                if color_order is None:
+                    raise config.error(
+                        f"Color order not supported: {color_order_cfg}")
+
+                self.strips[name] = Strip(name, color_order, cfg)
+
+            event_loop = self.server.get_event_loop()
+            event_loop.register_callback(
+                self._initalize_strips, list(self.strips.values()))
```
Arksine, line 255: With the new changes to Moonraker initialization you no longer need this callback.  More details to follow in the next comment.

moonraker/components/wled.py
```diff
@@ -0,0 +1,445 @@
...
+            self.server.register_endpoint(
+                "/machine/wled/strip", ["GET", "POST"],
+                self._handle_single_wled_request)
+
+        except Exception as e:
+            logging.exception(e)
+
+    async def _initalize_strips(self: WLED,
+                                initial_strips: List[Strip]
+                                ) -> None:
```
Arksine, line 286: This method should be renamed to `component_init()`.  It can't take an argument (outside of `self`) so you will need to directly reference `self.strips.values()` instead of `initial_strips`.

Moonraker will call the `component_init()` method after the event loop starts but before the http server starts.

moonraker/components/wled.py
```diff
@@ -0,0 +1,445 @@
...
+
+    async def wled_on(self: WLED, strip: str, preset: int) -> None:
+        if strip not in self.strips:
+            logging.info(f"Unknown WLED strip: {strip}")
+            return
+        await self.strips[strip].wled_on(preset)
+
+    # Full control of wled, True, False, "on", "off", preset 42 as int or
+    # preset "42" as string
+    async def set_wled_state(self: WLED, strip: str, state: str) -> None:
```
Arksine, line 324: I think it would make sense to have an additional "preset" argument here that defaults to -1 rather than attempt to use the state for both, ie:

```python 
async def set_wled_state(self: WLED, strip: str, state: str, preset: int = -1) -> None:
```

richardjm: Misunderstood your previous suggestion, fixed up

moonraker/components/wled.py
richardjm: Thanks, working again now

moonraker/components/wled.py
richardjm: Removed and moved

Arksine:
Looks good.  FWIW, I could see a scenario where this gets rolled into a generic led module, where there are multiple underlying implementations for controlling leds.  This is actually how the `power` module evolved, initially all it did was toggle gpios.  Should that occur I would tag you and give you a heads up.

Issue #262, opened 2021-11-05
[FR] Add MQTT Last Will And Testament (LWT) support


pedrolamas:
A good practice on MQTT is to have the client set a Last Will And Testament (LWT) that the broker will publish automatically if the device connection terminates in a non-conformant way.

It's like telling the MQTT broker "if I get disconnected for some reason publish this message to this topic".

Paho-MQTT does have support for this as shown [here](https://mntolia.com/mqtt-last-will-testament-explained-with-examples/) so hopefully it should be an easy improvement to add!

Arksine:
This seems useful and as you said it isn't difficult to add.  I suppose it comes down to hashing out implementation details, specifically which topic the will should be set to.  Presumably it would make sense to apply this to the status topic given that it is the one clients will be listening to, however we could make it configurable as well.

pedrolamas:
This is now complete, so closing this FR now.

PR #264, opened 2021-11-07
mqtt: implements last will and testament

pedrolamas:
Implements #262

While we discussed in the FR to have this under the "status" topic, I've gone with a separate "state" topic to ensure it is not overwritten by subsequent messages.

That way, one can check the "state" topic anytime to see if Moonraker is up and running!

### DIFF
--- a/moonraker/components/mqtt.py
+++ b/moonraker/components/mqtt.py
@@ -1,4 +1,4 @@
-# MQTT client implemenation for Moonraker
+# MQTT client implementation for Moonraker
 #
 # Copyright (C) 2021  Eric Callahan <arksine.code@gmail.com>
 #
@@ -191,6 +191,7 @@ class MQTTClient(APITransport, Subscribable):
         self.api_request_topic = f"{self.instance_name}/moonraker/api/request"
         self.api_resp_topic = f"{self.instance_name}/moonraker/api/response"
         self.klipper_status_topic = f"{self.instance_name}/klipper/status"
+        self.moonraker_status_topic = f"{self.instance_name}/moonraker/status"
         status_cfg = config.get("status_objects", None)
         self.status_objs: Dict[str, Any] = {}
         if status_cfg is not None:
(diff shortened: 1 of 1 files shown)


moonraker/components/mqtt.py
```diff
@@ -191,6 +191,7 @@ def __init__(self, config: ConfigHelper) -> None:
         self.api_request_topic = f"{self.instance_name}/moonraker/api/request"
         self.api_resp_topic = f"{self.instance_name}/moonraker/api/response"
         self.klipper_status_topic = f"{self.instance_name}/klipper/status"
+        self.klipper_state_topic = f"{self.instance_name}/klipper/state"
```
pedrolamas, line 194: Come to think of it, I do wonder if this should be `{self.instance_name}/moonraker/state` or even just `{self.instance_name}/state`

pedrolamas:
For the record, I've been using this locally with my printer and so far it seems to be working ok!

I've also confirmed that MQTT server will set to "offline" automatically if I use `kill -9 <pid>` on Moonraker process.

Arksine:
Ok, from what I a gather you want to introduce a new topic that reports the server's state and set the last will and testament on that topic.  I think that this makes sense and adds value, however I do think that we need to change the topic.  Since the data is specific to Moonraker's status the topic should probably be `{instance_name}/moonraker/status`.  The payload could be a json object, initially set to `{"server": "online"}`, with the lwt set to `{"server": "offline"}`.  This gives us flexibility to expand what is reported through this topic in the future.

There doesn't appear to be a merge conflict with my recent changes, however it would probably be a good idea for you to rebase this branch against master.

pedrolamas:
Excellent, I'll make the changes in a couple of hours when I'm back home!

pedrolamas:
Changes we discussed are in now! 🙂

pedrolamas:
@Arksine I did the change to JSON blob (`{"server": "online"}` and `{"server": "offline"}`) instead of a simple value (`online` and `offline`), but I think I should point that there seems to be a standard to use the simple value!

Here's an example of this, straight from Home Assistant documentation: https://www.home-assistant.io/integrations/sensor.mqtt/#availability_topic

Arksine:
Looks good.  FWIW, I'm not sure I would consider that a standard.  It appears to be how Home Assistant implements its MQTT Sensor availability, but even there the payload is configurable.  I'm somewhat surprised that they didn't allow for the use of templates.  Regardless, I'm not sure it would be correct to configure Moonraker as a Sensor in Home Assistant.  Most likely it would need its own plugin.

pedrolamas:
> I'm not sure it would be correct to configure Moonraker as a Sensor in Home Assistant

I've tried it already, it works fine and this small change for availability makes it even more powerful!! 😁

PR #267, opened 2021-11-12
power: improves GPIO initial_state handling

### LINKED ISSUE #266: GPIO not working correctly in Debian Bullseye / Python 3.9.2
Until today, I had my RaspberryPi 3B+ with FluiddPi and in, I have a relay on GPIO23 that controls some LED lights.

This is what I have in moonraker.conf:

```
[power Lights]
type: gpio
pin: gpiochip0/gpio23
initial_state: off
```

When the pi starts (or when moonraker restarts) the lights are off until the moment I choose to turn them on via Fluidd interface.

A few hours ago I decided to flash my RaspberryPi with the latest RaspiOs and then used Kiauh to install Klipper and Moonraker.

After this, I stopped both services, restored the configs and Moonraker database, and restarted both services.

When I did this, the lights turned on and a split second afterwards off again!

I noticed that if I restart Moonraker, that is happening every single time!!

I ran the following commands on a python REPL and it works perfectly:
```python
import gpiod
from gpiozero import LED
led = LED(23)
led.on()
led.off()
```

So my guess is that there is some issue with Python on the newest RaspiOS which is Debian Bullseye based.

pedrolamas:
Fixes #266

I can confirm this is working fine on my RaspberryPi 3B+ running latest RaspiOS (Debian Bullseye), but better to ensure it also works on other systems (FluiddPi, Mainsail OS, ...)

### DIFF
--- a/moonraker/components/power.py
+++ b/moonraker/components/power.py
@@ -478,10 +478,12 @@ class GpioDevice(PowerDevice):
             if invert:
                 self.line.request(
                     consumer="moonraker", type=gpiod.LINE_REQ_DIR_OUT,
-                    flags=gpiod.LINE_REQ_FLAG_ACTIVE_LOW)
+                    flags=gpiod.LINE_REQ_FLAG_ACTIVE_LOW,
+                    default_val=int(self.initial_state))
             else:
                 self.line.request(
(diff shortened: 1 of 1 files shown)


Arksine:
Thanks.

PR #277, opened 2021-11-17
mqtt: retry connection if failed to resolve hostname

pedrolamas:
I've seen quite a few times this error with MQTT on my production instance:

image

Here's the output from moonraker's `/server/info`:

```json
{
    "result": {
        "klippy_connected": true,
        "klippy_state": "ready",
        "components": [
            "database",
            "file_manager",
            "klippy_apis",
            "machine",
            "data_store",
            "shell_command",
            "proc_stats",
            "job_state",
            "job_queue",
            "authorization",
            "octoprint_compat",
            "history",
            "update_manager",
            "gpio",
            "power",
            "mqtt"
        ],
        "failed_components": [
            "mqtt"
        ],
        "registered_directories": [
            "config",
            "logs",
            "gcodes",
            "config_examples",
            "docs"
        ],
        "warnings": [
            "Component 'mqtt' failed to load with error: [Errno -3] Temporary failure in name resolution"
        ],
        "websocket_count": 1,
        "moonraker_version": "v0.7.1-152-gdf18928-dirty"
    }
}
```

I realized that this is just a temporary error, possibly due to the wifi still turning on and thus unable to resolve the hostname of the mqtt server.

That "Temporary failure in name resolution" is of type `socket.gaierror`.

Retrying it seems to fix the problem, so this is the fix for it.

### DIFF
--- a/moonraker/components/mqtt.py
+++ b/moonraker/components/mqtt.py
@@ -243,8 +243,8 @@ class MQTTClient(APITransport, Subscribable):
         for _ in range(retries):
             try:
                 self.client.connect(self.address, self.port)
-            except ConnectionRefusedError:
-                logging.info("Unable to connect to MQTT broker, "
+            except (ConnectionRefusedError, socket.gaierror) as e:
+                logging.info(f"MQTT connection error, {e}, "
                              f"retries remaining: {retries}")
                 await asyncio.sleep(2.)
             else:
@@ -368,7 +368,7 @@ class MQTTClient(APITransport, Subscribable):
                 break
             try:
                 self.client.reconnect()
-            except ConnectionRefusedError:
+            except (ConnectionRefusedError, socket.gaierror):
                 continue
             self.client.socket().setsockopt(
                 socket.SOL_SOCKET, socket.SO_SNDBUF, 2048)


moonraker/components/mqtt.py
```diff
@@ -247,6 +247,10 @@ async def component_init(self) -> None:
                 logging.info("Unable to connect to MQTT broker, "
                              f"retries remaining: {retries}")
                 await asyncio.sleep(2.)
+            except socket.gaierror:
+                logging.info("Unable to resolve MQTT broker hostname, "
+                             f"retries remaining: {retries}")
+                await asyncio.sleep(2.)
```
Arksine, line 253: I think we can handle this in a single except statement with something like the following:
```python
for _ in range(retries):
    try:
        self.client.connect(self.address, self.port)
    except (ConnectionRefusedError, socket.gaierror) as e:
        logging.info(f"MQTT connection error, {e}, "
                     "retries remaining: {retries}")
        await asyncio.sleep(2.)
```

Arksine (commented):
Thanks.  Its a good addition, I just have a minor suggestion for how to handle the exception.  Also, we should probably handle `socket.gaierror` in `_do_reconnect()` (line 371,375) as well.

moonraker/components/mqtt.py
pedrolamas: Done!

FYI, this is output right now:

```text
MQTT connection error, [Errno -3] Temporary failure in name resolution, retries remaining: 5
```

Arksine: That was fast!.  I think that is fine.  If necessary we can refactor the error message later.

PR #280, opened 2021-11-21
zeroconf: add module for zeroconf discovery

cmroche:
Enable Zeroconf service registration allowing external services to more easily detect and use Moonraker instances.

Example of the service announcement:

image

### DIFF
--- a/moonraker/components/zeroconf.py
+++ b/moonraker/components/zeroconf.py
@@ -1,0 +1,59 @@
+# Zeroconf registration implementation for Moonraker
+#
+# Copyright (C) 2021  Eric Callahan <arksine.code@gmail.com>
+#
+# This file may be distributed under the terms of the GNU GPLv3 license.
+from __future__ import annotations
+import socket
+import asyncio
+
+from zeroconf import IPVersion
+from zeroconf.asyncio import AsyncServiceInfo, AsyncZeroconf
+
+from typing import TYPE_CHECKING, List, Optional
+if TYPE_CHECKING:
+    from confighelper import ConfigHelper
+
+class AsyncRunner:
+    def __init__(self, ip_version: IPVersion) -> None:
+        self.ip_version = ip_version
+        self.aiozc: Optional[AsyncZeroconf] = None
+
+    async def register_services(self, infos: List[AsyncServiceInfo]) -> None:
+        self.aiozc = AsyncZeroconf(ip_version=self.ip_version)
+        tasks = [self.aiozc.async_register_service(info) for info in infos]
+        background_tasks = await asyncio.gather(*tasks)
+        await asyncio.gather(*background_tasks)
+
+    async def unregister_services(self, infos: List[AsyncServiceInfo]) -> None:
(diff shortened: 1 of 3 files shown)


moonraker/components/zeroconf.py
```diff
@@ -0,0 +1,64 @@
...
+        self.ip_version = ip_version
+        self.aiozc: Optional[AsyncZeroconf] = None
+
+    async def register_services(self, infos: List[AsyncServiceInfo]) -> None:
+        self.aiozc = AsyncZeroconf(ip_version=self.ip_version)
+        tasks = [self.aiozc.async_register_service(info) for info in infos]
+        background_tasks = await asyncio.gather(*tasks)
+        await asyncio.gather(*background_tasks)
+        while True:
+            await asyncio.sleep(1)
```
Arksine, line 31: I'm not sure what the purpose is of the while loop here.

moonraker/components/zeroconf.py
```diff
@@ -0,0 +1,64 @@
...
+            f"{host_name}._moonraker._tcp.local.",
+            addresses=addresses,
+            port=port,
+            properties={'path': '/'},
+            server=f"{host_name}.local."
+        )
+        self.runner = AsyncRunner(IPVersion.All)
+        self.event_loop.register_callback(
+            self.runner.register_services, [self.service_info]
+        )
```
Arksine, line 58: Rather than register a callback you can add a `component_init` method that Moonraker will call just prior to starting the server.  This guarantees that initialization occurs before the server starts.  For example:
```python
async def component_init(self):
    await self.runner.register_services([self.service_info])
```

Of course this assumes that the while loop I commented on above doesn't have a purpose I'm not seeing.

Arksine (commented):
Thanks, I think adding mDNS to Moonraker would be a great addition.  I have some questions/comments which I have included in the review.

moonraker/components/zeroconf.py
cmroche: Heh yes... sorry this was a quick side task in a bigger project, I'll look it over and test today then update the MR.

Arksine:
Thanks.  The module looks good to me.  I merged #279, however it appears to have created a conflict with this branch now.  Specifically it appears that the first commit of this PR adds the serial number and the second commit revert those changes when it adds the zeroconf module.

One other thing I noticed is the header:
```python
# MQTT client implementation for Moonraker
#
# Copyright (C) 2021  Eric Callahan <arksine.code@gmail.com>
#
# This file may be distributed under the terms of the GNU GPLv3 license.
```

Looks like you used the MQTT module as a template for the header, which is understandable as its what I do when creating a new module.  If you could just briefly describe the zeroconf module here and replace my info with yours it would be great.  Thanks.

cmroche:
Yeah, will push a change soon. Actually I realize now that I didn't branch clean off master before working on this, hence the conflict. I will rebase of master and fix this, but it will mean another force-push.

Arksine:
Looks good, thanks

PR #281, opened 2021-11-23
WIP: Exclude object preprocessing

troy-jacobson:


moonraker/components/file_manager/file_manager.py
```diff
@@ -518,7 +518,17 @@ async def _finish_gcode_upload(self,
...
+        # of this method will need to be reimplemented.  Also some
+        # additional error handling throughout this process
+        # finfo = await self._process_uploaded_file(upload_info)
+
+        finfo = self.get_path_info(
+            upload_info['tmp_file_path'], upload_info['root'])
+        if upload_info['unzip_ufp']:
+            finfo['ufp_path'] = upload_info['tmp_file_path']
+        else:
+            finfo['gcode_path'] = upload_info['tmp_file_path']
```
Arksine, line 531: This is going to break inotify functionality. When an inotify watch detects a new file in the gcode path its going to request the metadata parser to process the file at the existing location.  The combination of this change and the change to the metadata parser is going to result in a failure.

moonraker/components/file_manager/file_manager.py
```diff
@@ -1537,18 +1549,40 @@ async def _process_metadata_update(self) -> None:
...
+                exclude_object_enabled = "exclude_object" in kobjects
+        except Exception:
+            # After logging, continue processing with the assumption that
+            # exclude_object is not enabled.
+            logging.exception(
+                "Error querering klipper for exclude_object status")
+
+        if exclude_object_enabled:
+            timeout = 300.
+            cmd += " --exclude-object"
```
Arksine, line 1584: Rather than attempt to detect if `exclude_object` is configure in Klipper I think we should just add a configuration option to the `file_manager`, ie `enable_object_processing`.  It is perfectly acceptable for a user to upload a file when Klipper in not available and I think we want to be able to process objects in that scenario.

Arksine:
Before we get too far with this we need to verify that your module's dependency on Shapely isn't going cause issues on platforms that don't provide wheels, specifically Armbian.  The wheel must build in a timely fashion and it can't require any sort of intervention or additional dependencies from the user.  If it does then I think we need to hold off on Moonraker integration until we can come up with a pure python solution.

As mentioned above, the existing approach in this PR is going to break metadata processing for gcode files detected by inotify.  Specifically, when inotify detects a new file its going to make a request to process the file in the gcode path.  Since there is no file at the tempdir your changes to `metadata.py` will exit with an error.  We need to keep changes to `file_manager.py` minimal.

I think I can put something together reasonably quickly that will achieve what we want to accomplish.  Right now I think the primary concern is the dependency.

troy-jacobson:
I believe shapely is optional for kageurufu's  preprocessor.

I'm confused by what you're saying about inotify and not having the file in the temp directory.  I think what I'm doing now is essentially what the UPF processing is doing.  Of all the things I experimented with, this is the cleanest as far as UI activity and from what I can gather in the logs as far as what processing is happening there.

Arksine:
FWIW, I believe that PR #286 will handle this and resolve the issues you may have had with extra inotify events.

> I'm confused by what you're saying about inotify and not having the file in the temp directory. I think what I'm doing now is essentially what the UPF processing is doing.

Its similar, however it requires that the gcode file exist in the temp directory.  As far as I can tell if the gcode path is not specified then the metadata parser will exit with an error.  The thing to keep in mind is that the metadata parser is not just called from the gcode upload handler, its also called in the following sitations:
- When Moonraker starts and detects a gcode file that does not have metadata
- When inotify detects that a new gcode file has been copied into the gcode folder
- When a folder containing gcode files is copied into the gcode folder
In each of the above scenarios, the `gcode_path` wouldn't be passed to the metadata parser.  With the changes in this PR the metadata parser will exit with an error when this occurs, because the file at `gc_temp` will not exist.

Arksine:
I went ahead and merged the branch that implements support for object processing, so I'm going to close this.  Thanks.

PR #284, opened 2021-11-24
zeroconf: fix detection of interface addresses

cmroche:
Previous method of relying on **socket.gethostbyname()** would typically return "127.0.1.1" which would then cause an avahi-daemon conflict to be detected. Avahi would then revoke it's record breaking the **hostname.local** resolution shortly after starting moonraker.

To prevent the conflict it's necessary to use non link-local addresses in the record. The new approach first uses the bound interface only, if provided, or scan and uses any non link-local and non loopback IPv4 addresses on the system.

The **zeroconf** modules is broken and will have undesirable effects for most users until this fix is merged.

Signed-off-by: Clifford Roche <clifford.roche@gmail.com>

### DIFF
--- a/moonraker/components/machine.py
+++ b/moonraker/components/machine.py
@@ -430,7 +430,8 @@ class Machine:
                 addresses: List[Dict[str, Any]] = [
                     {
                         'family': IP_FAMILIES[addr['family']],
-                        'address': addr['local']
+                        'address': addr['local'],
+                        'is_link_local': addr.get('scope', "") == "link"
                     }
                     for addr in interface.get('addr_info', [])
                     if 'family' in addr and 'local' in addr
@@ -441,6 +442,7 @@ class Machine:
                 }
         except Exception:
             logging.exception("Error processing network update")
+            return
         prev_network = self.system_info.get('network', {})
         if notify and network != prev_network:
             self.server.send_event("machine:net_state_changed", network)
--- a/moonraker/components/proc_stats.py
+++ b/moonraker/components/proc_stats.py
@@ -163,6 +163,7 @@ class ProcStats:
                     self.server.send_event("proc_stats:cpu_throttled", ts)
                 self.last_throttled = cur_throttled
                 self.total_throttled |= cur_throttled
+            await self.machine.parse_network_interfaces()
         await self.machine.update_service_status()
 
     async def _check_throttled_state(self) -> Dict[str, Any]:


Arksine:
We are going to need to implement this without depending on netifaces.  The issue is that it isn't pure python and its wheel will need to be built on some platforms (Odroid, Armbian, etc).  We need to find a pure python solution for this, or write our own solution that uses CFFI if necessary.

moonraker/components/zeroconf.py
```diff
@@ -32,20 +38,26 @@ async def unregister_services(self, infos: List[AsyncServiceInfo]) -> None:
...
+
+class ZeroconfRegistrar:
     def __init__(self, config: ConfigHelper) -> None:
         self.server = config.get_server()
         self.event_loop = self.server.get_event_loop()
 
         host_name, port = self.server.get_host_info()
-        addresses = [socket.gethostbyname(host_name)]
+        addresses = [socket.inet_aton(self.server.host)]
+        if self.server.host == "0.0.0.0":
```
Arksine, line 49: We don't want to directly access attributes of other classes.  If necessary we can add a getter to the server class or refactor get_host_info to return the bound address.

cmroche:
> We are going to need to implement this without depending on netifaces. The issue is that it isn't pure python and its wheel will need to be built on some platforms (Odroid, Armbian, etc). We need to find a pure python solution for this, or write our own solution that uses CFFI if necessary. One potential solution is to use shell command and parse the output of `ip addr`, which is probably easier than writing a low level interface.
> 
> We may also need to consider that some instances may only be bound to ipv6 addresses.

I'll take a look into other potential solutions then over the weekend. Agreed with the ipv6, I was planning to tackle that in another pass though since I would need to setup again to test this specific case (no ipv4 present). That being said, avahi is propagating the IPv6 record automatically, but I think it's because it's matching the IPv4 on the interface so will still break down if IPv6 only.

cmroche:
FYI, found that other Zeroconf users are using this library which is pure python: https://pypi.org/project/ifcfg/

So, could be an easy solution. I'll keep investigating.

cmroche:
@Arksine Ok, the move to pure python was far easier than expected, and **ifcfg** should work with both older **ifconfig** systems as well as newer **ip** based systems, and as well as Windows if needed. IPv6 was an easy add, address records are correct and no conflicts which is a good sign, but I have not tested in IPv6 only configuration.

Arksine:
That library effectively does what I suggested, however its making a direct call to POpen with `shell=True`.   My preference here would be to implement the parser directly in Moonraker using the `shell_command` module.

I'll see if I can get around to implementing the parser directly into Moonraker when I get a chance.

cmroche:
> That library effectively does what I suggested, however its making a direct call to POpen with `shell=True`. My preference here would be to implement the parser directly in Moonraker using the `shell_command` module.
> 
> I'll see if I can get around to implementing the parser when I get a chance.

Do we only have to support **ip** or does **ifconfig** and **ipconfig** have to be supported for Windows and MacOS?

Arksine:
We don't need to support Windows or MacOS, Moonraker only runs on linux.  I'm also not worried about ipconfig, Moonraker requires a minimum of Python 3.7, any linux distribution that ships with it will also ship with `ip`.

That said, I'm not sure we really need their parser.  I dont think it will be too much trouble to come up with a regex that will parse the output from `ip address`.  I think it just comes down to how much information we need.  It looks like we  need to organize IP addresses by interface and family.  Its probably also a good idea to parse out the lease time for each address and set up a timer to update the ip address when the next lease expires.  While ipv4 addresses are less likely to change when they renew, ipv6 addresses can change quite often depending on the ISP (ie: no prefix delegation).

cmroche:
That simplifies the problem a lot in any case. My time is a bit limited before the weekend, but I'll try to work on that. I'm not entirely sure we actually need to worry about DHCP renewal, I think it will be taken care of for us. I'll run a test and see what happens.

cmroche:
Ok, I've tested adding, removing and changing. We will need to update our records, maybe timing as avahi-daemon doesn't handle this for us even if it consider the service records to be the same. However I don't think it will be sufficient to look at the release time, because typically DCHPClients are free to request a renewal once half the lease time has expired (and not obligated to do so). The behaviour about when it will be renewed is not well defined.

Since we need to remain as a pure python approach, I think the next best option is going to be polling on some frequency, 10 seconds and reregister the record when a change in any address is detected. This is probably enough to enable a successful reconnect after on the second retry for more clients.

cmroche:
@Arksine It's internal now, I added the 10 second check too for IP changes.

Arksine:
Ok, I pushed some changes that created some conflicts with your pull request, but it should help accomplish what we need.

First, I refactored the get_host_info() method so it returns a dict:
```python
def get_host_info(self) -> Dict[str, Any]:
        return {
            'hostname': socket.gethostname(),
            'address': self.host,
            'port': self.port,
            'ssl_port': self.ssl_port
        }
```

Second, I added the network interface parser interface parser to `machine.py`. I discovered that `ip -json address` outputs the data in json format which makes parsing quite a bit easier.   You can initialize in your component init method:
```python
def component_init(self) -> None:
    machine: self.machine: Machine = self.server.lookup_component('machine')
    network = machine.get_system_info()['network']
    ips = [ x for x in self._extract_ip_addresses(network)]
    ...

def _extract_ip_addresses(self, network: Dict[str, Any]) -> Iterator[str]:
    for ifname, ifinfo in network.items():
        for addr_info in ifinfo['ip_addresses']:
             if addr_info['is_link_local']:
                 continue
             yield addr_info['address']
````
The network parser will not include the loopback interface nor will it include interfaces that are down, so you really only need to skip link local addresses.  If you prefer you can do that after you instantiate the `ip_address`.

The `proc_stats` module handles polling the addresses for updates.  The `machine:net_state_changed` event will be emitted when a change is detected.  The network dict is passed to that event, so you can register an event handler that takes the network dict, processes it, and updates the zeroconf service info.

As an aside, doesn't zeroconf have a method to update services?  Is it necessary to unregister and re-register them?

cmroche:
> doesn't zeroconf have a method to update services

Looking through the code, indeed it does. In any case, I’ll look through your changes and update later tonight.

PR #286, opened 2021-11-24
file_manager: add support for object processing

Arksine:
This enables support processing gcode files with object tags using the `postprocess-cancellation` module.  The primary differences between this approach an PR #281 are as follows:
- Metadata is parsed first.  At the time of metadata parsing Moonraker will determine if the gcode file has object tags.  If the file has tags and object processing is enabled the file is passed to the `preprocessor` after metadata extraction is complete.
- The `file_manager` checks for duplicate `create_file` notifications and discards them.  This prevents duplicate `create_file` events that occur when the processed file overwrites the original gcode.

I have tested this with PrusaSlicer.  This PR still needs testing with Cura and Ideamaker, however I expect that they should work correctly.

Another thing to note is that it would be possible for Moonraker's metadata parser to extract objects identified by the `DEFINE_OBJECT` gcode.  I know that your module in klipper reports this as well, however there may be an advantage to reporting them in metadata as it would be possible for clients to identify objects before the print has started.  That said, if we do this it would be ideal if the `preprocessor()` method returned a dict containing the objects, otherwise I would have to do a recursive call to `extract_metadata()` after processing.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -66,7 +66,19 @@ log_path:
 queue_gcode_uploads: False
 #   When set to True the file manager will add uploads to the job_queue when
 #   the `start_print` flag has been set.  The default if False.
+enable_object_processing: False
+#   When set to True gcode files will be run through a "preprocessor"
+#   during metdata extraction if object tags are detected.  This preprocessor
+#   replaces object tags with G-Code commands compatible with Klipper's
+#   "cancel object" functionality.  Note that this process is file I/O intensive,
+#   it is not recommended for usage on low resource SBCs such as a Pi Zero.
+#   The default is False.
 ```
+
+!!! Note
+    It is also possible to enable object processing directly in the slicer.
+    See the [preprocess-cancellation](https://github.com/kageurufu/cancelobject-preprocessor)
+    documentation for details.
 
 ### `[database]`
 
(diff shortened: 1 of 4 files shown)


wlhlm:
Thanks everyone involved for working on this :+1:

I have been trying this out and it has been working nicely so far without a major hitch. a couple of notes though concerning `preprocess-cancellation`:
- the preprocessor can use quite a bit of RAM on plates with many parts (big plates with many parts are especially interesting for exclude object functionality)
  - for a 270x270 plate of Voron parts it needs more than 1GB of RAM requiring a Pi 4 for doing "on printer processing"
- the preprocessor has AFAICT no licensing information attached to it neither in the [repo](https://github.com/kageurufu/cancelobject-preprocessor), nor on PyPi (see kageurufu/cancelobject-preprocessor#1)
  - should be clarified before merging support in moonraker IMO
 
Also, it seems there is a 0.1.5 release on PyPi, which is not available in the GH repo.

/cc @kageurufu

kageurufu:
@wlhlm I've updated with a LICENSE file, and pushed 0.1.5 to github. It was a minor requirements change specifically to make moonraker support easier.

In regards to the 1GB+ ram, were you using shapely during processing? The non-shapely path should be fairly low-ram, if it's hitting that much usage I'd like to look into it

wlhlm:
@kageurufu 

> In regards to the 1GB+ ram, were you using shapely during processing? The non-shapely path should be fairly low-ram, if it's hitting that much usage I'd like to look into it

Ahh indeed, I've used shapely.

Arksine:
It sounds like it was a good idea to remove the shapely requirement, anything other than a Pi 4 would struggle with that kind of Memory Usage.

I'll probably merge this in a couple of days if there are no issues reported with it

PR #288, opened 2021-11-26
zeroconf: Implement new approach for ip detection

cmroche:
GitHub is for some reason showing the old PR as merge, this is an error, and it won't let me restore so I need to create a new PR: https://github.com/Arksine/moonraker/pull/284

However this implements the latest and greatest approach to getting and updating IP address for the Zeroconf service.

### DIFF
--- a/moonraker/components/zeroconf.py
+++ b/moonraker/components/zeroconf.py
@@ -6,13 +6,18 @@
 from __future__ import annotations
 import socket
 import asyncio
+import logging
+
+from .machine import Machine
 
 from zeroconf import IPVersion
 from zeroconf.asyncio import AsyncServiceInfo, AsyncZeroconf
 
-from typing import TYPE_CHECKING, List, Optional
+from typing import TYPE_CHECKING, Any, Dict, Iterator, List, Optional
+
 if TYPE_CHECKING:
     from confighelper import ConfigHelper
+
 
 class AsyncRunner:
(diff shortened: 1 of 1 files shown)


cmroche:
@Arksine  Sorry to fracture the conversation here, GitHub refuses to let me use the old PR any longer. But I think this should do it.

moonraker/components/zeroconf.py
```diff
@@ -32,29 +40,65 @@ async def unregister_services(self, infos: List[AsyncServiceInfo]) -> None:
         await asyncio.gather(*background_tasks)
         await self.aiozc.async_close()
 
-class ZeroconfRegistrar():
+    async def update_services(self, infos: List[AsyncServiceInfo]) -> None:
+        assert self.aiozc is not None
+        tasks = [self.aiozc.async_update_service(info) for info in infos]
+        background_tasks = await asyncio.gather(*tasks)
+        await asyncio.gather(*background_tasks)
+        await self.aiozc.async_close()
```
Arksine, line 48: Should `async_close()` be called here?  I would think that this would shut down the zeroconf.

Arksine:
No problem, Github probably just didn't like the merge conflict.  For the most part everything looks good, just need to check on the `async_close()` call.

moonraker/components/zeroconf.py
cmroche: Nope, missing commit.. one moment.

PR #294, opened 2021-12-02
Add API key info to example-home-assistant.yaml

RafaelSchridi:
Currently the file makes no mention of any type of auth, so I've added a short example with an explanation of when you should use it.

Signed-off-by:  Rafael Schridi <r.schridi@gmail.com>


Arksine:
Thanks.

Issue #289, opened 2021-11-26
[FR] Add secrets file


pedrolamas:
Some of us have passwords and host names on the moonraker config file that ideally should not be exposed either on the logs (that we share for support/debug purposes) or even when we share our configs in GitHub or elsewhere.

What I propose is something like Home Assistant and ESPHome have for secrets: https://www.home-assistant.io/docs/configuration/secrets/

Arksine:
This is now available, the documentation is [here](https://moonraker.readthedocs.io/en/latest/configuration/#secrets).

pedrolamas:
Thank you!! 😁

pedrolamas:
@Arksine secrets is working perfectly for username and password - thanks for that - but what do you think about adding at least for server names also? I ask this as I use a registered domain name as a local one (so my devices are `device.local.mydomain.com`) and would prefer not to expose the public domain!

Arksine:
Understood.  There are some options that take an address where I could add support for secrets, however there are some where I cannot.  For example, it is necessary to see what is in the `cors_domains` option to troubleshoot it.  Also it is likely that in some cases the supplied address gets logged.

PR #310, opened 2021-12-12
feat: support for token file for Home Assitant Power

RomRider:
I don't like to store the token in the configuration as I store that on git.

This PR allows to use a file to provide the token for Home Assistant Power devices.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -411,7 +411,9 @@ port:
 device:
 #   The device ID of the switch to control. This parameter must be provided.
 token:
-#   A token used for request authorization.  This paramter must be provided.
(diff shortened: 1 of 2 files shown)


Arksine:
Thanks.  I have a plan to resolve this in a way similar to what is requested in #289.  The way I want to do it will require Jinja2 template support, which will be added shortly, its part of an MQTT power device implementation I'm working on.

Arksine:
The [secrets] module is now available for storing credentials and tokens that you wish to keep out of the config and logs.  Thanks.

RomRider:
Amazing, thanks 🙏

PR #312, opened 2021-12-15
Reversed Data Source hierarchy from "distro" Module

KwadFan:
* By default distro.* reads from distro-release file at last.
* To show custom distro names in Frontends
* reversed that and grab first that distro-release file.
* See https://distro.readthedocs.io/en/latest/#distro-release-file
* If file present read that first, if not stick with default behavior.

default:
`"distribution":{
   "name":"Raspbian GNU/Linux 10 (buster)",
   "id":"raspbian",
   "version":"10",
   "version_parts":{
      "major":"10",
      "minor":"",
      "build_number":""
   },
   "like":"debian",
   "codename":"buster"
},`
 tested:
`"distribution":{
   "name":"MainsailOS 0.5.0",
   "codename":"buster",
   "id":"raspbian",
   "version":"10",
   "version_parts":{
      "major":"10",
      "minor":"",
      "build_number":""
   },
   "like":"debian"
},`

/etc/mainsailos-release :
`MainsailOS release 0.5.0 (Jack Sparrow)`

(Note: This codename is a joke ;) And only to be LSB conform. )

Signed-off-by: Stephan Wendel <me@stephanwe.de>

### DIFF
--- a/moonraker/components/machine.py
+++ b/moonraker/components/machine.py
@@ -42,6 +42,8 @@ SD_MFGRS = {
     '74': "PNY"
 }
 IP_FAMILIES = {'inet': 'ipv4', 'inet6': 'ipv6'}
+
+
 class Machine:
     def __init__(self, config: ConfigHelper) -> None:
         self.server = config.get_server()
@@ -55,8 +57,6 @@ class Machine:
             }
         else:
             dist_info = {"name": distro.name(pretty=True)}
-        dist_info.update(distro.info())
-        dist_info = {'name': distro.name(pretty=True)}
         dist_info.update(distro.info())
         self.inside_container = False
         self.virt_id = "none"
@@ -460,6 +460,7 @@ class Machine:
         if notify and network != prev_network:
             self.server.send_event("machine:net_state_changed", network)
         self.system_info['network'] = network
+
 
 def load_component(config: ConfigHelper) -> Machine:
     return Machine(config)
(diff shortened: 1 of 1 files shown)


Arksine:
Thanks.  My initial thought here is that the fix should be in MainsailOS, not Moonraker.  I don't think it is a good idea to make changes to Moonraker that accommodate specific clients and/or distributions.  It can lead to code churn (ie: some clients may want Moonraker to report info in a specific manner, whereas other want it reported another way).

I'm okay with adding another field that reports the supplemental release info though.

KwadFan:
> Thanks. My initial thought here is that the fix should be in MainsailOS, not Moonraker.
Unfortunatly, there cant be a fix for any Distro based on another Distro like Raspberry PI OS.
Simply you can't manipulate /etc/os-releases without having downsides form other programs.
So, therefor any raspbian-like distro is affected by that. The only way to get arround this is the "/etc/distro-release" file. But this is only read by Py Module distro in last position if all other fail. This was my intention to PR that.
But I understend your concerns, so lets stash that :)

Regards

Arksine:
Thanks.  I went ahead and pushed commit 3e70e4b4cee1a9ad4673f48e138d32859492e80d which adds a `release_info` object to the `distribution` field.  This is sort of a compromise, while the `name` field will continue to report what the OS provides, Mainsail can check the `release_info` field and apply the name there if it exits.

PR #320, opened 2021-12-30
feat: secrets support for hosts in mqtt and power

RomRider:
Adding support for secrets in hostnames config fields for mqtt and power modules

Ideally, `cors_domains` and `trusted_clients` should also support secrets but that would require some refactoring in `getlists`, or it would require the secrets file to be in json. What do you think?

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -466,6 +466,7 @@ address:
 #   to control socket index 1:
 #     192.168.1.127/1
 #   This parameter must be provided.
+#   This option accepts Jinja2 Templates, see the [secrets] section for details.
 port:
 #   The port to connect to.  Default is 9999.
 #
@@ -494,7 +495,8 @@ The following options are available for `tasmota` device types:
 #   Tasmota.
 address:
 #   A valid ip address or hostname for the tasmota device.  This parameter
-#   must be provided.
+#   must be provided. This option accepts Jinja2 Templates, see the [secrets]
+#   section for details.
 password:
 #   A password used to authenticate requests.  Default is no password.
 output_id:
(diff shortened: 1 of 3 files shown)


Arksine:
I'm not sure I want to add support for this.  Generally speaking the host isn't sensitive information, and its possible that knowing what is in that option is useful for debugging.  In fact it is likely logged in some places.

The `trutsted_clients` and `cors_domains` option absolutely need to be visible for debugging.

RomRider:
In terms of storing information on git, hostnames/ip addresses are considered sensitive information.

Could we imagine that some specific debug levels/a flag to enable in the config would display those information in the log file with user opt-in?

Arksine:
There are a few solultions available to you if you are concerned about distirbuting your local IP/hostnames on git:
1) Make your git repo private
2) Use git daemon to serve your repos locally
3) Set up your own git server to serve your repos 

Its necessary to see the IP/hostname to address some connection issues, and as mentioned above its absolutely necessary to see the trusted clients and cors domains options to provide help for authorization problems.  Providing an opt-out for this is counterproductive IMO.

RomRider:
I was thinking that maybe a good way to handle this is to use the same mechanism as klipper to include files.
So I'd include a secrets.cfg file with overrides for the hostnames/secrets in general and not store that on git.

In that case, it would work with any fields in the config, so no need for specific development each time on each component.

Arksine:
I agree that include functionality would probably be useful and its on my list of things I plan to add.

Arksine:
Include functionality is supported so its possible to override these values in a separate configuration not stored in your git repo.  Given that this should be resolved I'll close this PR.

PR #322, opened 2021-12-31
metadata.py: add more metadata

th33xitus:
This PR will add parsing of 3 new, optional metadata types for:
- PrusaSlicer and it's derivatives
- Simplify3D
- Cura
- IdeaMaker

Those types are:
* Nozzle diameter
* Filament type
* Filament name

Note: As far as i am aware Simplify3D does not provide any info about the filament type. At least i didn't find anything during research and testing. So in case of Simplify3D, this PR will only provide filament name and nozzle diameter.

For Cura and IdeaMaker it is necessary to add a custom start g-code comment because they do not output the requested metadata to the g-code file by default. But they do provide internal placeholder variables for those values. The required lines for the start g-code are as follows (it is important to have the exact syntax as shown) :

Cura:
```
;Nozzle diameter = {machine_nozzle_size}
;Filament type = {material_type}
;Filament name = {material_name}
```

IdeaMaker:
```
;Nozzle diameter = {machine_nozzle_diameter1}
;Filament type = {filament_name_abbreviation1}
;Filament name = {filament_name1}
```

### DIFF
--- a/moonraker/components/file_manager/metadata.py
+++ b/moonraker/components/file_manager/metadata.py
@@ -26,7 +26,6 @@ from typing import (
     Dict,
     List,
     Type,
-    Union
 )
 if TYPE_CHECKING:
     pass
--- a/moonraker/components/file_manager/metadata.py
+++ b/moonraker/components/file_manager/metadata.py
@@ -179,6 +179,9 @@ class BaseSlicer(object):
     def parse_filament_weight_total(self) -> Optional[float]:
         return None
 
+    def parse_filament_name(self) -> Optional[str]:
+        return None
+
     def parse_filament_type(self) -> Optional[str]:
         return None
 
@@ -350,6 +353,10 @@ class PrusaSlicer(BaseSlicer):
         return _regex_find_string(
             r";\sfilament_type\s=\s(.*)", self.footer_data)
 
+    def parse_filament_name(self) -> Optional[str]:
+        return _regex_find_string(
+            r";\sfilament_settings_id\s=\s(.*)", self.footer_data)
+
     def parse_estimated_time(self) -> Optional[float]:
         time_match = re.search(
             r';\sestimated\sprinting\stime.*', self.footer_data)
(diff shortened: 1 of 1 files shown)


Arksine:
I'm ok with adding support for this in PrusaSlicer and its derivatives. I have some concerns about the custom gcode requirement for other slicers.  Generally speaking I only want to support extracting metadata when the slicer has native support for it, otherwise we end up with maintenance and legacy issues if they do end up adding support for it.

th33xitus:
I can understand your concerns. I thought it would be okay since there already was a PR successfully merged in the past which parsed the filament weight. That also requires Cura user to use a custom gcode to acquire that feature in a client. See: https://github.com/Arksine/moonraker/pull/143

Beside that i think it would be more feasible have such features now instead of getting the slicer developers of Cura and Ideamaker to include such exports "just for us" and wait an unknown amount of time to have that happen. I think the general response time of the Klipper community will be quick enough in case there would be some point in time they have native support for it which would result in some necessary maintenance work on the metadata.py code to fix any legacy issues. I think the time and effort to be spend for fixing that would be managable. In the end, those methods won't break anything in case native support will happen. Even with native support, this code would even be still working, unless the required placeholders in the slicer change.

Cura metadata in general is very sparse compared to PS for example. "Having nice things" on the client side, at least besides just the basic print time, layer height, object height or required filament length, is just not possible otherwise (currently).

But of course i am okay with it to remove all Cura and Ideamaker related changes i did and just leave the PS and S3D related code in that PR.

Arksine:
Ok, I must have missed that before.  Ideally the metadata we parse is there by default, we don't want to get into a situation where there are a lot of requests to add support for custom data.  I'll go ahead and merge this, but I think going forward I'll need to look into a configurable solution.

th33xitus:
Thanks for merging, i totally understand your point.
I got the info, that ideaMaker already exports the nozzle diameter. It is just in a place where i havent suspected it at all and just figured out by trial and error just this morning. So you will probably see a PR with a fix for that upcoming soon. That will already elimate at least one needed custom g-code line.

PR #325, opened 2022-01-03
wled: Add serial support for direct communication

richardjm:
- For mobile printers such as Voron 0.1 having to reconfigure wifi in each location is a pain
- In the 0.13 release (still in beta) direct serial communication was added - and seems to work very nicely

### DIFF
--- a/moonraker/components/wled.py
+++ b/moonraker/components/wled.py
@@ -219,17 +219,17 @@ class StripHttp(Strip):
         self.client = AsyncHTTPClient()
 
     async def send_wled_command_impl(self: StripHttp,
-                                state: Dict[str, Any]) -> None:
+                                     state: Dict[str, Any]) -> None:
         async with self.request_mutex:
             logging.debug(f"WLED: url:{self.url} json:{state}")
 
             headers = {"Content-Type": "application/json"}
             request = HTTPRequest(url=self.url,
-                                    method="POST",
-                                    headers=headers,
-                                    body=json.dumps(state),
-                                    connect_timeout=self.timeout,
-                                    request_timeout=self.timeout)
+                                  method="POST",
+                                  headers=headers,
+                                  body=json.dumps(state),
+                                  connect_timeout=self.timeout,
+                                  request_timeout=self.timeout)
             response = await self.client.fetch(request)
 
             logging.debug(
@@ -252,7 +252,7 @@ class StripSerial(Strip):
                                  write_timeout=0)
 
     async def send_wled_command_impl(self: StripSerial,
-                                state: Dict[str, Any]) -> None:
+                                     state: Dict[str, Any]) -> None:
         async with self.request_mutex:
             logging.debug(f"WLED: serial:{self.ser.name} json:{state}")
 
@@ -261,7 +261,7 @@ class StripSerial(Strip):
 
             # asyncio support is still experimental in pySerial
             self.ser.write(json.dumps(state).encode())
-    
+
     def close(self: StripSerial):
         if self.ser.is_open:
             self.ser.close()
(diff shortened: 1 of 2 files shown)


moonraker/components/wled.py
```diff
@@ -130,29 +144,52 @@ def _update_color_data(self: Strip,
...
-                logging.exception(msg)
-                raise self.server.error(msg)
+            logging.debug(f"WLED: serial:{self.ser.name} json:{state}")
+
+            if not self.ser.is_open:
+                self.ser.open()
+
+            # asyncio support is still experimental in pySerial
+            self.ser.write(json.dumps(state).encode())
+            self.ser.flush()
```
Arksine, line 188: We can't add blocking code to Moonraker.   After opening the serial port we must set the fd to a nonblocking state and we can't call flush().  If wled is sensitive to this it is possible to use asyncio to poll the fd and write to it when its ready.

Arksine:
Thanks.  I had a specific implementation comment as mentioned above.  Blocking the asyncio eventloop can lead to problems, there are ways around it without relying on any specific asyncio implementation in pyserial.  Its not a concern for the PanelDue, I set the fd to nonblocking and just write to it, however it isn't transferring much data so there is no real concern about the buffers filling resulting in an error.

At a high level, it might be better to break out the protocols into two separated classes and add a `type` option for each protocol.  You could have a base `Strip` class with the primary functionality then override the method that makes the request.  IMO this would make configuration and maintenance a bit easier.

moonraker/components/wled.py
```diff
@@ -130,29 +144,52 @@ def _update_color_data(self: Strip,
...
+
+            logging.debug(f"WLED: serial:{self.ser.name} flushed")
+
+            self.ser.close()
```
Arksine, line 192: Is it necessary to open and close the port for each request?  Can we just hold the serial port open in the constructor then close it when Moonraker is stopped?

richardjm:
Thanks just looking into things now but ABC isn't playing ball. I believe setting write_timeout to 0 should make things non-blocking.

moonraker/components/wled.py
richardjm: Done that via a close() method

moonraker/components/wled.py
richardjm: Removed the flush() and changed write_timeout to 0 which I believe is non blocking

richardjm:
I removed ABC as it seemed clever but not really needed. Changed the method to use type although this will break for users updating which I'm unsure about (maybe a default of http). Moved to separate classes as suggested.

richardjm:
@Arksine I think is ready for re-review

moonraker/components/wled.py
```diff
@@ -217,6 +202,72 @@ async def set_wled(self: Strip,
...
+                 cfg: ConfigHelper):
+        super().__init__(name, color_order, cfg)
+
+        # Read the serial information (requires wled 0.13 2108250 or greater)
+        serialport: str = cfg.get("serial")
+        baud: int = cfg.getint("baud", 115200, above=49)
+
+        # write_timeout of 0 is non-blocking
+        self.ser = serial.Serial(serialport, baud,
+                                 write_timeout=0)
```
Arksine, line 252: After the serial port is opened it is good practice to use `os.setblocking(fd, False)` to make sure that underlying file will not block.  In this case the fd is returned by `self.ser.fileno()`.

Arksine:
Aside making sure the fd can't block the changes look good to me.

moonraker/components/wled.py
richardjm: Made the suggested changes

PR #326, opened 2022-01-05
metadata.py: fix parsing of IdeaMaker nozzle_diameter

th33xitus:
This PR will improve a method previously introduced by #322.

IdeaMaker already exports the nozzle diameter, so using a custom g-code comment is obsolete. The new regex pattern extracts the natively exported nozzle diameter.

### DIFF
--- a/moonraker/components/file_manager/metadata.py
+++ b/moonraker/components/file_manager/metadata.py
@@ -740,7 +740,7 @@ class IdeaMaker(BaseSlicer):
 
     def parse_nozzle_diameter(self) -> Optional[float]:
         return _regex_find_first(
-            r";Nozzle\sdiameter\s=\s(\d+\.\d*)", self.header_data)
+            r";Dimension:(?:\s\d+\.\d+){3}\s(\d+\.\d+)", self.header_data)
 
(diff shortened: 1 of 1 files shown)


moonraker/components/file_manager/metadata.py
```diff
@@ -740,7 +740,7 @@ def parse_first_layer_bed_temp(self) -> Optional[float]:
 
     def parse_nozzle_diameter(self) -> Optional[float]:
         return _regex_find_first(
-            r";Nozzle\sdiameter\s=\s(\d+\.\d*)", self.header_data)
+            r";Dimension:(\s\d+\.\d+)*", self.header_data)
```
Arksine, line 743: While this will work, it could potentially match a value that doesn't strictly refer to the nozzle diameter.  I would recommend a more strict regex, such as:
```
r";Dimension:(?:\s\d+\.\d+){3}\s(\d+\.\d+)"
```

th33xitus: Thanks! Yes this more strict regex might be better. I force pushed a fix to this PR.

Arksine:
Looks good, thanks!

PR #327, opened 2022-01-12
power: add 'off_when_shutdown_delay' option

lukasfink1:
The added option allows to delay switching off a power source after klippy shuts down. This is useful when controlling the printer power through moonraker: In case klipper shuts down, the printer won’t be turned on indefinitely, but it will also have time to cool down the extruder before the fans are turned off.

### DIFF
--- a/moonraker/components/power.py
+++ b/moonraker/components/power.py
@@ -128,13 +128,9 @@ class PrinterPower:
                 "The following power devices failed init:"
                 f" {failed_names}")
 
-    async def _handle_klippy_shutdown(self) -> None:
-        for name, dev in self.devices.items():
-            if dev.has_off_when_shutdown():
-                logging.info(
-                    f"Powering off device [{name}] due to"
-                    " klippy shutdown")
-                await self._process_request(dev, "off")
+    def _handle_klippy_shutdown(self) -> None:
+        for dev in self.devices.values():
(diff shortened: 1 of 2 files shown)


moonraker/components/power.py
```diff
@@ -267,6 +263,14 @@ def __init__(self, config: ConfigHelper) -> None:
         self.locked_while_printing = config.getboolean(
             'locked_while_printing', False)
         self.off_when_shutdown = config.getboolean('off_when_shutdown', False)
+        self.off_when_shutdown_delay = 0.
+        if self.off_when_shutdown:
+            self.off_when_shutdown_delay = config.getfloat(
+                'off_when_shutdown_delay', 0.)
+            if self.off_when_shutdown_delay < 0.:
```
Arksine, line 270: The config "getfloat" and "getint" methods now accept optional `minval`, `maxval`, `above` and `below` arguments so it is no longer necessary to implement range checks.  For example:
```python
self.off_when_shutdown_delay = config.getfloat('off_when_shutdown_delay', 0., minval=0.)
```
Should accomplish this.  There are likely several places in the code that I need to switch over to this method, however I think its best to start using this going forward.

Arksine:
Thanks.  Overall it looks good, just the minor comment I left above.

PR #329, opened 2022-01-13
Add power device type GCode

dianlight:
Add power device type GCode. On/Off command send GCode to Klipper

The US is an RGB Led strip connected to mcu led controller ( in my case BTT Octopus led controller ). 

Signed-off-by: Lucio Tarantino <lucio.tarantino@gmail.com>


Arksine:
Thanks.  This had quite a bit of overlap with #336, so my thoughts were that the best approach would be to combine the two into a single device.  With commit 0b13e2da44bdd0f806165f84a81282643ac24b7e I added support for `gcode_macro` objects to the `klipper_device` type.

There are requirements for the macro:
- It must take a `VALUE` parameter.  If set to 0 the device should power off, if set to 1 it should power on
- It must configure a `variable_value` and set it to 0 when the device is turned off, 1 if its on.

This allows Moonraker to track device state should the macro be called outside of Moonraker's control.  It also gives Moonraker a way to query status if need be.  The `klipper_device` also excludes some config options that are common to all of the other device types, but would cause problems when combined with a device that relies on Klipper.

pedrolamas:
@Arksine am I correct to say that this is now implemented via [Klipper Device Configuration](https://moonraker.readthedocs.io/en/latest/configuration/#klipper-device-configuration) with `object_name: gcode_macro MY_MACRO` config of sorts?

Arksine:
Yes, a method of toggling Klipper devices has been implemented, so I will go ahead and close this.

PR #332, opened 2022-01-15
Richardjm/wled serial asyncio

richardjm:
Use pyserial_asyncio for serial communication due to dropped commands

### DIFF
--- a/moonraker/components/wled.py
+++ b/moonraker/components/wled.py
@@ -13,8 +13,8 @@ from enum import Enum
 import logging
 import json
 import asyncio
+import serial_asyncio
 import os
-import serial
 from tornado.httpclient import AsyncHTTPClient
 from tornado.httpclient import HTTPRequest
 from tornado.escape import json_decode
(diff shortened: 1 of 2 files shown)


moonraker/components/wled.py
```diff
@@ -245,31 +240,24 @@ def __init__(self: StripSerial,
...
 
-            if not self.ser.is_open:
-                self.ser.open()
+            logging.debug(f"WLED: serial:{self.serialport} json:{state}")
 
-            # asyncio support is still experimental in pySerial
             self.ser.write(json.dumps(state).encode())
 
     def close(self: StripSerial):
         if self.ser.is_open:
```
Arksine, line 258: Presumably if the serial port fails to open `self.ser` may not be set, which would result in a an AttributeError here.  Also I'm not sure if `StreamWriter` has an `is_open` attribute.  You could probably call `is_closing()`.

richardjm: That's only set by close() so changed to a hasattr() check instead

PR #334, opened 2022-01-17
wled: Show config errors

richardjm:
Support more synonym strings for colors
Case insensitive string type checks
Show errors in UI if e.g. wrong color was selected
Public color_config not color_order as it has confused users
Fix documentation for white stealthburner example

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1130,8 +1130,10 @@ initial_white:
 #   initial_white will only be used for RGBW wled strips (defaults: 0.5)
 chain_count:
 #   Number of addressable neopixels for use (default: 1)
-color_order:
-#   Color order for WLED strip, RGB or RGBW (default: RGB)
+color_config:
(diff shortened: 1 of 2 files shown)


Arksine:
I'm not sure what is confusing about `color_order`, can you expand on that?  We want to avoid breaking configuration if at all possible, because while some users may not like the semantics of `color_order` others are probably fine with it and this change would disable their wleds.

If we do break config an entry needs to be added to `docs/user_changes.md`, but before going there I think we need some further discussion to make sure its the correct approach.

richardjm:
I made it so that the old `color_order` setting would continue to work but a `color_config` setting would override it. This is due to the setting not actually controlling color_order and users getting confused when GRB wasn't working. This is set a lower level within wled directly. It was made worse by the error not appearing to the user.

richardjm:
Finding more bugs. I'll create another PR when I'm done

Issue #173, opened 2021-06-01
Control board pins as power plugin output


davide-molino:
I will connect a 12V led strip to one of the heater output on my skr and be able to control it via power plugin, but I can only choose between raspberry gpio and smart plugs. Is it possible to add such a function?

Arksine:
Unfortunately not.  Moonraker has limited access to GPIOs on the host machine that may be used with the `power` component.  You will have to use Klipper to control GPIOs on attached MCUs.

pedrolamas:
@davide-molino this is now implemented via [Klipper Device](https://moonraker.readthedocs.io/en/latest/configuration/#klipper-device-configuration) in `power`!

PR #336, opened 2022-01-18
power: adds support for Klipper Device

pedrolamas:
Implements #173 

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -303,7 +303,7 @@ The following configuration options are available for all power device types:
 
 [power device_name]
 type:
-#   The type of device.  Can be either gpio, mcu_output_pin, rf,
+#   The type of device.  Can be either gpio, klipper_device, rf,
 #   tplink_smartplug, tasmota, shelly, homeseer, homeassistant, loxonev1,
 #   or mqtt.
 #   This parameter must be provided.
@@ -379,7 +379,7 @@ initial_state: off
 timer:
 #    A time (in seconds) after which the device will power off after being.
 #    switched on. This effectively turns the device into a  momentary switch.
-#    This option is available for gpio, mcu_output_pin, tplink_smartplug,
+#    This option is available for gpio, klipper_device, tplink_smartplug,
 #    shelly, and tasmota devices.  The timer may be a floating point value
 #    for gpio types, it should be an integer for all other types.  The
 #    default is no timer is set.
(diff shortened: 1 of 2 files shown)


Arksine:
Overall the implementation looks solid, although we probably need to handle "klippy shutdown" events.  This could be done by overriding `process_klippy_shutdown()`.  There is some overlap between this PR and #329, which makes me wonder if an implementation should support output pins and specifically defined gcode macros. 

My biggest concern is the afore mentioned confusion that adding this support could introduce.  Initially the purpose of the `[power]` plugin was to control devices that Klipper cannot, as Klipper cannot power itself on.  I think we need to add a warning in the documentation that these devices cannot be used to toggle Klipper's power supply.  We may also need to disable options like `off_when_shutdown` and `restart_klipper_when_powered`.  We might also need to make sure that the device is not bound to the Klipper service.  I'll have to take a more in depth look to see what kind of problems those options could cause.

pedrolamas:
Thanks @Arksine!

Agreed on the overlap with #329, however this PR is ensuring we keep Moonraker state coordinated with the real one - so any changes to the MCU pins state (via UI, g-code, etc.) will be reflected on Moonraker state. One could probably improve the other PR by adding a `gcode_state_request` of sorts but I'm not sure that this would actually work to solve the sync state problem.

I was under the impression that monitoring `server:klippy_disconnect` would be enough (I'm still lacking a bit of understanding how these events actually work) but given your comments I will add the `server:klippy_shutdown` handler to ensure we cover all possibilities.

I will also improve the docs with better warnings.

Agreed that `off_when_shutdown` and `restart_klipper_when_powered` make no sense for this feature, so will code something to disable these!

pedrolamas:
I've now improved the docs with a small note and overrided `process_klippy_shutdown`.

Regarding the `off_when_shutdown` and `restart_klipper_when_powered`, I see a few options here...

## Create a new class that inherits/extends `PowerDevice`

Cleaner solution, refactor those 2 options out of `PowerDevice` into the new `PowerDeviceFull` class and make all existing power components inherit from this new `PowerDeviceFull` class (`McuOutputPin` would still inherit from `PowerDevice`).

This would also require updating this part of the docs to have 2 sections: https://moonraker.readthedocs.io/en/latest/configuration/#options-common-to-all-power-devices

## Log config warning on McuOutputPin init

Easy solution, we just log a warning when one of those 2 options are found on the config for a `McuOutputPin`, but we just set them to `False` anyway and carry on.

## Throw config error on McuOutputPin init

Easy solution, but it will make the `McuOutputPin` not work as it throws an error.

--

I have yet to check the bound stuff, but depending on what is decided on the above, we can move forward then!

@Arksine what's your take on the above?

Arksine:
Looks good, thanks

PR #339, opened 2022-01-22
Richardjm/wled bug fixes

richardjm:
- Removal of setting `color_order` documented and accepts (but ignores) setting - no change in functionality
- At least two off-by-one errors fixed in set_wled with heavy refactoring
- Workaround for wled bug (when setting pixels from off two calls are required)

### DIFF
--- a/moonraker/components/wled.py
+++ b/moonraker/components/wled.py
@@ -33,12 +33,12 @@ if TYPE_CHECKING:
     from . import klippy_apis
     APIComp = klippy_apis.KlippyAPI
 
-class ColorOrder(str, Enum):
+class ColorConfig(str, Enum):
     RGB: str = "RGB"
     RGBW: str = "RGBW"
 
     def Elem_Size(self):
-        if self is ColorOrder.RGB:
+        if self is ColorConfig.RGB:
             return 3
         return 4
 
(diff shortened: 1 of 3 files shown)


moonraker/components/wled.py
```diff
@@ -288,52 +282,46 @@ def __init__(self: WLED, config: ConfigHelper) -> None:
 
                 logging.info(f"WLED strip: {name}")
 
-                color_order_cfg: str = cfg.get("color_order", "RGB")
-                color_order = color_orders.get(color_order_cfg)
-                if color_order is None:
-                    raise cfg.error(
-                        f"Color order not supported: {color_order_cfg}")
+                # Discard old color_order setting, always support 4 color strips
+                _ = cfg.get("color_order", "")
```
Arksine, line 286: The ConfigHelper's various `get` methods accept a `deprecate` keyword argument that you can set to True .   When this flag is set Moonraker will add a warning.  Mainsail users will see this warning in the UI.  I'm not sure if other clients present warnings, but they are all logged as well so users may see it eventually.  So it this line could be something like
```python
_ = cfg.get("color_order", "", deprecate=True)
```

Arksine:
As far as I can tell everything looks good.  I do think it would be helpful to add the `deprecate` flag to the color_order option, as users will hopefully be made aware of the change and can take action as necessary.

moonraker/components/wled.py
richardjm: Thanks, didn't know about that (but didn't look for it either). Added as suggested, also added information on deprecation to configuration.md.

Issue #95, opened 2021-02-11
Feature request: PanelDue support for v3.2+ firmware


Norbs12:
The newer firmware has some useful features.

One that really stands out is you can now control one other heater which is useful for monitoring chamber or maybe a second extruder.

BastianGschrey:
+1
3.2 would be really helpfull.

Arksine:
I plan to look at this at some point, but it isn't a high priority.  Version 3.2 represents a significant change in the API therefore the plugin will need a lot of attention to support it, and it needs to be done without breaking compatibility with version 1.24.

trevjonez:
>  it needs to be done without breaking compatibility with version 1.24.

do we have data on those with panels that can't support 3.2 and are using moonraker+klipper? seems like if the croud size is small enough it would be best to just EOL support to ease the workload

Aianir:
Please keep supporting the PanelDue. I have one printer with PanelDue and one with the stock Ender 3 display. The PanelDue is far better when it comes to actually doing something on the printer, not just displaying progress.

Arksine:
I don't have any intention of dropping support for the PanelDue.  I will add support for version 3.2 when I have the opportunity, however as previously mentioned it isn't a priority and not something I have near term plans of working on.

SirDeath81:
Support for Paneldue Firmware 3.7.0 is on its way. Programming is completed, first tests with 3.7.0 all ok.
complete tests with Paneldue with Firmware 3.7.0 and 1.25.0 to follow within this week. If I'm really successful, I'll do a fork and then a Pull Request.
The only thing I want to ask you is a little bit of patience with me - this is my first time ever contributing something on github :-)

EDIT: The new Version of paneldue.py will support both Firmware Versions with autodetect of the connected Version so we don't destroy any existing installations :-)

PR #342, opened 2022-01-23
paneldue: add support for virtual macro folders

OasisOfChaos:
Added support for custom macro display names and a virtual folder structure.

Signed-off-by:  Alex <alex@nntw.nl>

### DIFF
--- a/moonraker/components/paneldue.py
+++ b/moonraker/components/paneldue.py
@@ -5,6 +5,7 @@
 # This file may be distributed under the terms of the GNU GPLv3 license.
 
 from __future__ import annotations
+from xmlrpc.client import boolean
 import serial
 import os
 import time
@@ -38,11 +39,13 @@ if TYPE_CHECKING:
 MIN_EST_TIME = 10.
 INITIALIZE_TIMEOUT = 10.
 
+
 class PanelDueError(ServerError):
     pass
 
 
 RESTART_GCODES = ["RESTART", "FIRMWARE_RESTART"]
+
 
 class SerialConnection:
     def __init__(self,
@@ -165,6 +168,7 @@ class SerialConnection:
                 return
         self.send_busy = False
 
+
 class PanelDue:
     def __init__(self, config: ConfigHelper) -> None:
         self.server = config.get_server()
@@ -204,19 +208,21 @@ class PanelDue:
         self.confirmed_gcode: str = ""
         self.mbox_sequence: int = 0
         self.available_macros: Dict[str, str] = {}
-        self.confirmed_macros = {
-            "RESTART": "RESTART",
-            "FIRMWARE_RESTART": "FIRMWARE_RESTART"}
+        self.confirmed_macros: Dict[str, str] = {}
+        # self.confirmed_macros = {
+        #     "RESTART": "RESTART",
+        #     "FIRMWARE_RESTART": "FIRMWARE_RESTART"}
+
         macros = config.getlist('macros', None)
         if macros is not None:
-            # The macro's configuration name is the key, whereas the full
-            # command is the value
-            self.available_macros = {m.split()[0]: m for m in macros}
+            self.available_macros = self.parse_macro_configline(macros)
+
         conf_macros = config.getlist('confirmed_macros', None)
         if conf_macros is not None:
             # The macro's configuration name is the key, whereas the full
             # command is the value
-            self.confirmed_macros = {m.split()[0]: m for m in conf_macros}
+            self.confirmed_macros = self.parse_macro_configline(conf_macros)
+
         self.available_macros.update(self.confirmed_macros)
 
         self.non_trivial_keys = config.getlist('non_trivial_keys',
@@ -425,7 +431,11 @@ class PanelDue:
         # Prepare GCodes that require special handling
         if cmd in self.special_gcodes:
             sgc_func = self.special_gcodes[cmd]
-            script = sgc_func(parts[1:])
+
+            if cmd == "M98":
+                script = sgc_func(parts)
+            else:
+                script = sgc_func(parts[1:])
 
         if not script:
             return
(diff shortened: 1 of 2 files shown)


moonraker/components/paneldue.py
```diff
@@ -5,6 +5,7 @@
 # This file may be distributed under the terms of the GNU GPLv3 license.
 
 from __future__ import annotations
+from xmlrpc.client import boolean
```
Arksine, line 8: I believe that this is an unintentional import, likely pulled in from an annotation below.

moonraker/components/paneldue.py
```diff
@@ -38,12 +39,14 @@
 MIN_EST_TIME = 10.
 INITIALIZE_TIMEOUT = 10.
 
+
```
Arksine, line 42: Please don't submit changes to whitespace in pull requests.  Moonraker follows the same standard as Klipper here, code style and whitespace changes aren't accepted unless they are made by the original author of the module.

moonraker/components/paneldue.py
```diff
@@ -204,19 +208,21 @@ def __init__(self, config: ConfigHelper) -> None:
         self.confirmed_gcode: str = ""
         self.mbox_sequence: int = 0
         self.available_macros: Dict[str, str] = {}
-        self.confirmed_macros = {
-            "RESTART": "RESTART",
-            "FIRMWARE_RESTART": "FIRMWARE_RESTART"}
+        self.confirmed_macros: Dict[str, str] = {}
+        # self.confirmed_macros = {
+        #     "RESTART": "RESTART",
+        #     "FIRMWARE_RESTART": "FIRMWARE_RESTART"}
```
Arksine, line 214: I would like to keep the `RESTART` and `FIRMWARE_RESTART` as default options.

moonraker/components/paneldue.py
```diff
@@ -425,7 +431,11 @@ def process_line(self, line: str) -> None:
         # Prepare GCodes that require special handling
         if cmd in self.special_gcodes:
             sgc_func = self.special_gcodes[cmd]
-            script = sgc_func(parts[1:])
+
+            if cmd == "M98":
+                script = sgc_func(parts)
+            else:
+                script = sgc_func(parts[1:])
```
Arksine, line 438: This doesn't appear necessary.  In the M98 handler doesn't need `parts[0]`, as that is just the command itself.

moonraker/components/paneldue.py
```diff
@@ -760,6 +773,44 @@ def _run_paneldue_M20(self, arg_p: str, arg_s: int = 0) -> None:
                 response['files'] = flist
         self.write_response(response)
 
+    def get_macro_dir(self, path: str, issubdir: boolean):
```
Arksine, line 776: The `boolean` annotation is incorrect here, it should be `bool`.  The return value also needs to be annotated.

moonraker/components/paneldue.py
```diff
@@ -497,15 +507,17 @@ def _prepare_M32(self, args: List[str]) -> str:
...
     def _prepare_M98(self, args: List[str]) -> str:
-        macro = args[0][1:].strip(" \"\t\n")
-        name_start = macro.rfind('/') + 1
-        macro = macro[name_start:]
-        cmd = self.available_macros.get(macro)
+        macro = args[1].replace("P0:/", "P/").replace("P/macros/", "")
+        for key, value in self.available_macros.items():
+            if value == macro:
+                cmd = key
+        logging.info(f"Trying to run macro '{cmd}'...")
```
Arksine, line 514: This should be debug level logging.

Arksine:
I left some review comments above.  At a high level we may need to look into an alternative method of configuration as its possible for macro parameters to contain commas.

Also your signed off line needs to contain your full name.

Thanks.

OasisOfChaos:
I've modified my working copy with your comments. It' my first Python project so bear with me please ;-). Also, I'm at about 98% of implementing the new object model that's used in firmware 3.3 and newer. Only things to do is multiple chamber heaters/generic heaters and some more testing with different configurations.


> I left some review comments above. At a high level we may need to look into an alternative method of configuration as its possible for macro parameters to contain commas.

True. Was thinking about gcode only and forgot about the Klipper parameters. In my working copy I'm using a '|' now, but if you find this a hacky solution and got another suggestion I can implement that as well. 

I thought about a list of key/value pairs under macros: but the delimiters cannot be brackets of course (but ```{}'s``` maybe?). This may be confusing for a user and doesn't really fit in the current config methods. Splitting the two into fields may also be a possibility but that can get messy fast, I guess. As of now, I unfortunately don't see an really elegant way of doing this and hope you or someone else does.

Arksine:
Not a problem.  We are a bit limited by the ini file structure here.  I think to be reliable we need to set strict requirements on folder names.  We could do something like the following:
```
macros:
  STATUS
  -my sounds
    @"test beep" PANELDUE_BEEP FREQUENCY=500 DURATION=1
  -calibration
    Z_TILT
    @bedmesh BED_MESH_CALIBRATE
    QUAD_GANTRY_LEVEL
```
In this example, folders must be prefixed by "-" and be on one line.    We can identify an alias for a macro with an @.  Aliases must either contain no whitespace or be quoted.  Honestly it might be better to just restrict it to no whitespace.  If no alias is available we fall back on the Macro's primary name.

To my knowledge macros cant start with a - or @, so we shouldn't have any conflicts.  Implementation shouldn't be too difficult.

With regard to adding support for the new object model, that is ok, however we want that in a different pull request.  Also we don't want to break compatibility with the existing model, so either the module needs to autodetect which model to use or it needs to be a config option.

OasisOfChaos:
Thanks. I propose to just make quotes mandatory after a '@' alias. That way a user has the freedom to choose whether to use whitespaces or not. Personally, I like spaces in my display instead of underscores or nameswithoutanything ;-)

Would it be acceptable to close this PR and just add this functionality to the version with the new object model and link it to this [#95](https://github.com/Arksine/moonraker/issues/95) feature request? BTW, the current version is already fully backwards compatible and will detect what responses to send. I am currently testing with the 3 main versions 1.24, 3.3 and 3.4.0.

Arksine:
That is fine, I would only ask that changes be separated into commits that clearly define the changes made, ie: a commit for adding virtual folders, a commit for adding 3.x support, etc.  I'm guessing that 3.x support is a sizable change.

OasisOfChaos:
> In this example, folders must be prefixed by "-" and be on one line. We can identify an alias for a macro with an @. Aliases must either contain no whitespace or be quoted. Honestly it might be better to just restrict it to no whitespace. If no alias is available we fall back on the Macro's primary name.

Unfortunately, this is not going to work with the current ConfigHelper class (and the underlying RawConfigParser. Your solution is dependent on multi-level parsing for subdirectories and that is not supported (all spaces are stripped in a very early stage (in ```RawConfigParser.get()```) and will require a significant change imo. 

```
macros:
  STATUS
  /my sounds
    @"test beep" PANELDUE_BEEP FREQUENCY=500 DURATION=1
  /my sounds/subsounds
    @"test beep a" PANELDUE_BEEP FREQUENCY=500 DURATION=1
    @"test beep b" PANELDUE_BEEP FREQUENCY=500 DURATION=1
  /calibration
    Z_TILT
    @bedmesh BED_MESH_CALIBRATE
    QUAD_GANTRY_LEVEL
```

This would be parsable while retaining the a folder structure and require no change in the ConfigHelper class. It also has the benefit of using the character that is commonly used for directories. Would this be something you can live with?

Arksine:
You solution looks good.   I agree its more clear for users and easier to parse as it doesn't require recursion.

PR #355, opened 2022-02-02
module: power

matthew-humphrey:
Add separate output_id configuration attribute for TPLink Smartplug device.

Signed-off-by:  Matthew Humphrey <mhumphrey@gmail.com>

### DIFF
--- a/moonraker/components/power.py
+++ b/moonraker/components/power.py
@@ -672,7 +672,12 @@ class TPLinkSmartPlug(PowerDevice):
         super().__init__(config)
         self.timer = config.get("timer", "")
         self.request_mutex = asyncio.Lock()
-        self.addr: List[str] = config.get("address").split('/')
+        addr_and_output_id = config.get("address").split('/')
+        self.addr = addr_and_output_id[0]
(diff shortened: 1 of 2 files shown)


moonraker/components/power.py
```diff
@@ -684,11 +689,11 @@ async def _send_tplink_command(self,
                 'system': {'set_relay_state': {'state': int(command == "on")}}
             }
             # TPLink device controls multiple devices
-            if len(self.addr) == 2:
+            if self.output_id != 0:
```
Arksine, line 692: Is it possible for a smart strip to have an output id of 0?  If so, will it work without specifying the child id?

I'm wondering if we should default `output_id` to None and check for a None type rather than zero.  If the ID can't be zero, or if it defaults to 0, then we can leave it as it is.

matthew-humphrey: I agree that seems safer.

Arksine:
Thanks.  I like the idea, and I agree with allowing the previous option to remain. We should probably add a warning if its detected, however I can do that later.  We only need to confirm that disallowing output id 0 on a Kasa strip will work correctly.

PR #358, opened 2022-02-04
database: transition to asynchronous transactions

Arksine:
As things stand today all lmdb database transactions are done synchronously.  This allowed us to wrap the database in a class that implements Python's mapping protocol and treat it mostly like a `dictionary`.  Unfortunately I have discovered that database transactions can block the asyncio event loop.  Typically this occurs after heavy disk I/O, such as a large file upload or system package update.  In my observations the loop blocks about 5 seconds then recovers, however this will be system dependent as its the Kernel that is actually blocking.  While 5 seconds of blocking isn't likely going to cause major issues, it could potentially sever the unix connection to Klippy since Moonraker can't drain the read buffer on the socket while its blocked.  The impact on clients here is that all status updates accumulated during the time blocked would be pushed at once over the websocket.

This PR runs all transactions in thread if the event loop is running.  If the DB transaction blocks then Python should release the GIL and the main thread will continue to be responsive.  It should be noted that the event loop starts after all components have been instantiated, so DB transactions inside of `__init__()` still run synchronously. 

All database operations now return a Future object.  If the event loop is not running (ie: inside a component's `__init__()` method) method, then the Future's result can be immediately queried via the `result()` method.  If the event loop is running then the developer must do one of the following to get the result:
1) `await` the future
2) set a callback via `Future.add_done_callback()`

Some additional high level implementation notes:
- It isn't necessary to await write operations.  If you want to insert or delete items you can call those methods synchronously and they will be queued.  Each transaction holds a mutex so it isn't possible for transactions to intertwine.  That said, it wouldnt be ideal to synchronously call write operations in loop.  Use one of the new `batch` operations if possible, otherwise await the the result of the write before performing the next one when looping.
- The ability to access the DB synchronously (in `__init__()`) is available for a transition period.  Removing this ability would have required more refactoring to existing modules that depend on the database than I wanted to do at this time.  However, it is possible that I make that move at a future date.  For new components I would suggest:
  - wrapping the DB Namespace in `__init__()` (this is currently a requirement, it isn't possible to call `wrap_namespace` when the event loop is running)
  - performing any initial DB transactions in `component_init()`
- If you using the DB to store a small amount of persistent data I would recommend keeping a local cache of that data and synchronizing with the database when you need to update an entry.  This way you can access the data synchronously.  See the changes to `authorization.py` for an example of this.

I have been testing these changes locally, however given the impact of these changes I want to write some unit tests before I merge this.

While this has no impact on users or client devs, Moonraker devs that wish to use the database will need to take note of the changes. @FrYakaTKoP @jordanruthe I believe you are the two most likely devs that need to be aware of these changes.   I don't want to break `timelapse` with this, if its using the DB I'll wait to merge until the necessary changes are ready.


Arksine:
@FrYakaTKoP I thought you might be using it.  I won't merge until we have everything straightened out on your end.

With regard to your first question, if you are in `__init__()` it shouldn't be necessary to do any checks, you can just call `.result()` on the return value.  Most of the database methods, including `get_item()`, always return a Future.  Before the server starts (ie: `__init__`) that future is resolved immediately before its returned, thus you can call `result()` without an error.   After the server starts you must await the Future to get the result. 

That said, my guess is that you aren't storing much data in the database and are using it for persistent storage.  I would recommend taking the approach I used `authorization.py`, keep a local cache and just sync with the database when you need to do a write.  So for example, it could be something like this:

```python
# Timelapse init
def __init__(self, config):
    #...other init stuff
    self.timelapse_db = database.wrap_namespace("timelapse")
    # The `as_dict()` method is the only method in the namespace wrapper
    # that does not return a future.  It can't be called after the server starts.
    self.cache = self.timelapse_db.as_dict()
    self.config.update(self.cache.get("config", {}))
    # ..etc
```

When you want to read you can directly access the cache, its a pure dict object.  When you want to write, sync with the database.
```python
self.cache['some_key'] = "some value"
self.timelapse_db['some_key'] = self.cache['some_key']
```
If you dont want to use the namespace you could use the database methods:
```python
self.cache['some_key'] = "some value"
self.database.insert_item("timelapse", "some_key", "some_value)
```

FrYakaTKoP:
The check is just so it does not break compatibility with older moonraker versions, but yeah i maybe remove it after sometime when most people have updated their moonraker.

About the local cache: That's exactly what i was already doing. Only read the db once and keep the settings in a dict object. Only do write operations to the timelapse namespace while the component is running.

However timelapse reads the webcam settings of mainsail which are in a different namespace and needs to be read later on. With your examples it should be fairly easy to adapt the rest of the code, so thanks for saving me some time there.

Again i will change it this weekend and report once timelapse is ready.

Did you talk to @jordanruthe ? I just made a quick search on the klipperscreen repo: It doesn't seems like he is using the moonraker database at all.

Arksine:
Ah, that makes sense.  I'm not sure your code would accomplish that though, you either need to check the attribute of the object or check the instance.  So for example, the below code would work:

```python
if isinstance(dbconfig, asyncio.Future):
    self.config.update(dbconfig.result())
else:
    self.config.update(dbconfig)
```
or
```python
if hasattr(dbconfig, "result"):
    self.config.update(dbconfig.result())
else:
    self.config.update(dbconfig)
```
The first is the safest method IMO.

FWIW, I took a look at your code and it looks like your reads after the server starts occur inside of async methods.  I wouldn't bother keeping a cache for this, I would just check to see if the return value of `get_item()` is a future and await it if inside the async method.   This would probably require a bit of refactoring to `getwebcamconfig()` since its called from both an async method and `__init__()`.  Up to you though.

With regard to KlipperScreen, even if it accesses the database its doing so through the websocket API, so it won't be affected by these changes.  That said, Jordan may be working on components for Moonraker that do access the db, so I wanted him to be aware of these changes as well.

FrYakaTKoP:
@Arksine  
I did merge my changes to master. Timelapse should now work with both versions of moonraker, so as said before you are free to merge when you are ready. 

Thanks again for the help you provided!

Arksine:
This has been merged, it took me a bit to write comprehensive tests for the database.  I did manual testing for the `history`, `file_manager`, `authorization`, and `update_manager` modules, however we should be on the lookout for hidden bugs that I may have missed.  I need to write tests for all of them (and the entire codebase), however that is going to take quite a bit of time and I wanted to resolve the event loop blocking issue as soon as possible.

PR #361, opened 2022-02-10
Adding power option for SmartThings switches

qtemp:
See the following for access token generation: https://developer-preview.smartthings.com/docs/advanced/authorization-and-permissions/

The device guids for use in the configuration can be found by doing a GET against https://api.smartthings.com/v1/devices with your 'bearer' token.


qtemp:
Couldn't get link below 80 characters.  Could remove but I think it's useful information.

PR #362, opened 2022-02-11
Add support for smartthings switch

qtemp:


Arksine:
Hi, I see you are fighting the linter and static type checker.  If you use VS Code it can help identify where the issues are at.  FWIW, docs aren't linted.

That said, to save you some additional frustration, I plan on deprecating of the "http" implementations in the power module, replacing with a generic http type.  Now that Moonraker supports templates and secrets it should be possible to support nearly every type of http device with one implementation.  We can add sample configurations for all supported types to the documentation.

qtemp:
Thanks Eric.

Yes, I had read your changes to the contributions readme and realized my
confusion on what was erroring. I made a couple more minor changes based on
testing, got everything fixed and ready but I decided to wait till morning.

Any idea on when the templated method will be available?

Thanks,

Mitch

PR #363, opened 2022-02-11
power: add support for smartthings switch

qtemp:
Added support for smartthings switch in power.py.
Added documentation to configuration.md.

I understand that changes to implementation of power over HTTP will change in the near future.  So, feel free to reject based on that.

Signed-off-by: Mitch Gallman <mitchell.gallman@gmail.com>

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -881,6 +881,40 @@ state_response_template:
 # not necessary to query after a command
 query_after_command: False
 ```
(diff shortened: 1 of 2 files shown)


Arksine:
Thanks.  The implementation looks good, and given the effort you put into this I'm going to merge.

There is one small change I am going to make, and that is to render the `token` option as a template.  This allows users to keep their token in a `secrets` file stored outside of their config.

PR #364, opened 2022-02-11
Update example-home-assistant.yaml

tispokes:
- Less rows and still a graph for actual (bed/heater) temp.
- Changed some icons. Also by context.
- removed file extension (e.g. ".gcode") from filename

Done printing:
image

Printing in progress:
image

### DIFF
--- a/docs/example-home-assistant.yaml
+++ b/docs/example-home-assistant.yaml
@@ -1,10 +1,15 @@
-#  Example Home Assistant configuration file for a Voron V0.
+#  Example Home Assistant configuration file for a Artillery Sidewinder X1
 #  Credit to GitHub users @Kruppes and @pedrolamas
-#
+#  expanded by @tispokes
+camera:
+  - platform: mjpeg
+    still_image_url: http://192.168.178.66/webcam/?action=snapshot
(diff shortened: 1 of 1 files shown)


Arksine:
I don't have a problem with adding an extended/alternate example, however if the repo is going to carry more than one example we probably need to identify the file by the printer, it `homeassistant-artillery-sidewinder-x1.yaml`.  I can handle renaming the original.

Also, commits should be formatted and signed as described in the [contributing documentation](https://moonraker.readthedocs.io/en/latest/contributing/#git-commit-format).  You can rename the file, push a commit with the correct format and signature, and I'll handle squashing them into one commit.  Thanks.

tispokes:
Due it is not specific to a swx1 nether the origin yaml is specific to a voron I would not agree to the naming convention.
Instead I prefer to rename the vars in the `yaml`-file(s) to something more generic. But due it is only a example and not a real template, it does visualize the usage very good as it is. But sure, just my opinion.

Maybe the keyword "extended" is wrong, as non native speaker I was the first word, which came to my mind.

As the origin file hasn't a appropiate copyright header, I don't know how to handle this task properly.
PR message edited.

PR #369, opened 2022-02-16
Allow Slic3r forks & v2 methods

supermerill:
Allow Superslicer forks to be detected, Allow Slic3r v2 to take the same path as prusaslicer

Allow superslicer and slic3r to contains more character before the space, allowing compliant forks to be detected.
If Slic3r has its version to be at least '2', then it can use the same methods as superslicer/prusaslicer

Note that I can't test it. Hopefully I didnt make a dumb mistake.

### DIFF
--- a/moonraker/components/file_manager/metadata.py
+++ b/moonraker/components/file_manager/metadata.py
@@ -304,7 +304,7 @@ class PrusaSlicer(BaseSlicer):
     def check_identity(self, data: str) -> Optional[Dict[str, str]]:
         aliases = {
             'PrusaSlicer': r"PrusaSlicer\s(.*)\son",
-            'SuperSlicer': r"SuperSlicer\s(.*)\son",
+            'SuperSlicer': r"SuperSlicer[^\s]*\s(.*)\son",
             'SliCR-3D': r"SliCR-3D\s(.*)\son"
         }
         for name, expr in aliases.items():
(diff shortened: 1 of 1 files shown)


Arksine:
Thanks Merill.  The PR failed lint because some of the statements were terminated with semicolons. I know the feeling, when you get used to writing in C/C++ adding them is natural.

As to the contents of the PR, generally I think its a good idea to be selective about the slicer identity.  If someone creates a fork, changes some of the metadata, and it fails parsing then it would be hard to track down what happened from my end.   One thing forks could do is either prefix or append a unique identifier to the version rather than append it to `SuperSlicer`.  Put differently, I think it best to put the onus on the fork developer for support here rather than on Moonraker.

I wasn't aware there was a version 2 Slic3r.  I'd like to learn more about it, do you have a link?  I'm okay with adding support, but given that it seems to be based on PrusaSlicer I'm thinking it needs its own class that inherits from it.

Finally, all commits should be formatted and signed as described in the [contributing documentation](https://moonraker.readthedocs.io/en/latest/contributing/#git-commit-format).

supermerill:
>  I think its a good idea to be selective about the slicer identity.

Why not, I was wondering how to make a fork compatible.
I already have a fork with a key like "SuperSlicer-VertexEdition" (it's a skinned version with some defaults changed).
It's maybe better to output the gcode with `by SuperSlicer 2.4.58 Vertex edition`  instead of `by SuperSlicer-VertexEdition 2.4.58` , so it has the same compatibility without any work done by the other software...

> I wasn't aware there was a version 2 Slic3r. I'd like to learn more about it, do you have a link? I'm okay with adding support, but given that it seems to be based on PrusaSlicer I'm thinking it needs its own class that inherits from PS.

It's mostly a merge between superslicer and slic3r, continuing from the current superslicer codebase. I made some cosmetic changes to have the names & colors of slic3r, and it's currently sitting on the slic3r/merill-merge branch.
It needs more tests & doc to go further.

> Finally, all commits should be formatted and signed as described in the [contributing documentation](https://moonraker.readthedocs.io/en/latest/contributing/#git-commit-format).

oups, sorry

I think I can close this until the slic3rv2 is out.

Arksine:
Ah, so these are your forks?   I'll leave it up to you, if you want you can append it to the version, otherwise I'm willing to make the adjustment.  FWIW, I do think we should capture the fork type and report it in the metadata.  That way we can see specifically which version of SuperSlicer users are having.  If necessary I can report an addition value in the metadata for this.  That said, I already report the version, so something like:
```
by SuperSlicer 2.4.58-VertexEdition
```
would get captured and reported by Moonraker.

It sounds like Slicer V2 is yours as well.  In that case, just let me know when its close to ready and I'll add something for it (or you can open a new PR if you wish).

PR #371, opened 2022-02-16
Implemented config-based webcam settings for Cura

sgermain06:
Now reading configs from moonraker.conf file, octoprint_compat section. Values are set to the defaults that were previously hardcoded and reflected in the sample moonraker.conf file.

### DIFF
--- a/moonraker/components/octoprint_compat.py
+++ b/moonraker/components/octoprint_compat.py
@@ -48,7 +48,7 @@ class OctoprintCompat:
             'flipH': config.getboolean('flipH', False),
             'flipV': config.getboolean('flipV', False),
             'rotate90': config.getboolean('rotate90', False),
-            'streamUrl': config.getboolean('streamUrl', '/webcam/?action=stream'),
+            'streamUrl': config.get('streamUrl', '/webcam/?action=stream'),
             'webcamEnabled': config.getboolean('webcamEnabled', True),
         }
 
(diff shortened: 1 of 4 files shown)


moonraker/components/octoprint_compat.py
```diff
@@ -43,6 +43,15 @@ def __init__(self, config: ConfigHelper) -> None:
             'software_version')
         self.enable_ufp: bool = config.getboolean('enable_ufp', True)
 
+        # Get webcam settings from config
+        self.webcam: Dict[str, Any] = {
+            'flipH': config.getboolean('flipH', False),
+            'flipV': config.getboolean('flipV', False),
+            'rotate90': config.getboolean('rotate90', False),
+            'streamUrl': config.get('streamUrl', '/webcam/?action=stream'),
+            'webcamEnabled': config.getboolean('webcamEnabled', True),
```
Arksine, line 52: Moonraker's configuration options are standardized to use snake case.  They should be something like:
 ```
flip_h: 
flip_v
rotate_90:
stream_url:
webcam_enabled:
```
Its fine for the dict keys to use camel case since that is what the response requires.

Arksine:
Thanks.  The in general this looks good outside of the case issue.  The commits need to be formatted and signed off as described in the [contributing documentation](https://moonraker.readthedocs.io/en/latest/contributing/#git-commit-format).

Arksine:
Thanks!

PR #377, opened 2022-02-23
notifier: create the new notifier component 

pataar:
# Notifier component
This component will be a bridge between moonraker and https://github.com/caronc/apprise. This way users can easily add all kind of notification services to their printer. 

Each notifier can subscribe to events.

## Event types
- completed
- started
- error
- canceled

## Configuration syntax

```ini
[notifier my_personal_discord]
events: started,completed
url: https://discord.com/api/webhooks/{id}/{secret}}
body: Printer update: '{event_name}'

[notifier print_start]
url: tgram://{bottoken}/{ChatID}
events: started
body: Your printer started printing '{event_args[1].filename}'

### DIFF
--- a/moonraker/components/notifier.py
+++ b/moonraker/components/notifier.py
@@ -1,0 +1,152 @@
+# Notifier
+#
+# Copyright (C) 2022 Pataar <me@pataar.nl>
+#
+# This file may be distributed under the terms of the GNU GPLv3 license.
+
+from __future__ import annotations
+
+import apprise
+import logging
+
+# Annotation imports
+from typing import (
+    TYPE_CHECKING,
+    Type,
+    Optional,
+    Dict,
+    Any,
+    List,
+)
+
+if TYPE_CHECKING:
+    from confighelper import ConfigHelper
+    from . import klippy_apis
+
+    APIComp = klippy_apis.KlippyAPI
+
+
+class Notifier:
+    def __init__(self, config: ConfigHelper) -> None:
+        self.server = config.get_server()
+        self.notifiers: Dict[str, NotifierInstance] = {}
+        self.events: Dict[str, NotifierEvent] = {}
+        prefix_sections = config.get_prefix_sections("notifier")
+
+        self.register_events(config)
+
+        for section in prefix_sections:
+            cfg = config[section]
+            try:
+                notifier = NotifierInstance(cfg)
+
+                for event in self.events:
+                    if event in notifier.events or "*" in notifier.events:
+                        self.events[event].register_notifier(notifier)
+
+                logging.info(f"Registered notifier: '{notifier.get_name()}'")
+
+            except Exception as e:
+                msg = f"Failed to load notifier[{cfg.get_name()}]\n{e}"
+                self.server.add_warning(msg)
+                continue
+            self.notifiers[notifier.get_name()] = notifier
+
+    def register_events(self, config: ConfigHelper):
+
+        self.events["started"] = NotifierEvent(
+            "started",
+            "job_state:started",
+            config)
+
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by Arksine** [comment]
I have taken a quick look, I have some initial comments.  First, some general observations:
1) We don't want to add IDE specific files to `.gitignore`.  Generally speaking those belong in the global gitignore on your system.
2) Check the [contributing documentation](https://moonraker.readthedocs.io/en/latest/contributing/) for details on how commits should be formatted.
3) The module will need configuration documentation.

With regard the module itself, I have a few comments that I will leave in a review.

File: moonraker/components/notifier.py [hunk 1]
```diff
@@ -0,0 +1,133 @@
...
+
+                logging.info(f"Registered notifier: '{notifier.get_name()}'")
+
+            except Exception as e:
+                msg = f"Failed to load notifier[{cfg.get_name()}]\n{e}"
+                self.server.add_warning(msg)
+                continue
+            self.notifiers[notifier.get_name()] = notifier
+
+    def register_events(self, config: ConfigHelper):
```
Arksine, line 56: I'm ok with hardcoding events to start, but we might want to try to think of a way of allowing users to define events.  This is something we can explore later though.

moonraker/components/notifier.py
```diff
@@ -0,0 +1,133 @@
...
+                      prev_stats: Dict[str, Any],
+                      new_stats: Dict[str, Any]
+                      ) -> None:
+        try:
+            logging.info(f"'{self.event_name}' event triggered'")
+            await self.notify(self.event_name)
+        except self.server.error as e:
+            logging.info(f"Error while notifiying '{self.event_name}'")
+
+    async def notify(self, body="test"):
```
Arksine, line 104: Should we make the body configurable?  It could be a template that has the event name and any arguments received in the event passed to the context.  Right now it appears to just be the event name.

moonraker/components/notifier.py
```diff
@@ -0,0 +1,133 @@
...
+        self.server.register_event_handler(self.event_name, self._handle)
+
+    def register_notifier(self, notifier: NotifierInstance):
+        self.notifiers[notifier.get_name()] = notifier
+        self.apprise.add(notifier.url)
+
+    async def _handle(self,
+                      prev_stats: Dict[str, Any],
+                      new_stats: Dict[str, Any]
+                      ) -> None:
```
Arksine, line 97: We might to capture `*args` here so we aren't limited to `job_state` events.  Should we choose to make the body configurable, we could then pass `args` to the template's context.

moonraker/components/notifier.py
pataar: I agree though

vertigo235:
Is there not a pause event?

pataar:
> Is there not a pause event?

Not that I'm aware of. @Arksine is there any event we could use?

vertigo235:
The main reason I would like to use this is to be notified when the printer is paused, for M600, or ERCF/MMU2S intervention requirements...

vertigo235:
actually there is a paused state event, I added 

        self.events["paused"] = NotifierEvent(
            "paused",
            "job_state:paused",
            config)

and it works just fine, I'll do a pull request

I think resume is also available, looks like you pulled the events from the History module, which doesn't use them all

pataar:
> actually there is a paused state event, I added
> 
> ```
>     self.events["paused"] = NotifierEvent(
>         "paused",
>         "job_state:paused",
>         config)
> ```
> 
> and it works just fine, I'll do a pull request
> 
> I think resume is also available, looks like you pulled the events from the History module, which doesn't use them all

Thanks for working this out!

PR #380, opened 2022-02-25
notifier: add support for 'attach' property

pataar:
This adds support for the `attach` property that is available to numerous notification services.

### DIFF
--- a/moonraker/components/notifier.py
+++ b/moonraker/components/notifier.py
@@ -17,6 +17,7 @@ from typing import (
     Dict,
     Any,
     List,
+    Union,
 )
 
 if TYPE_CHECKING:
@@ -112,6 +113,14 @@ class NotifierInstance:
         self.server = config.get_server()
         self.name = name_parts[1]
         self.apprise = apprise.Apprise()
+        self.warned = False
+
+        self.attach_requires_file_system_check = True
+        self.attach = config.get("attach", None)
+        if self.attach is None or \
+            (self.attach.startswith("http://") or
+             self.attach.startswith("https://")):
+            self.attach_requires_file_system_check = False
 
         url_template = config.gettemplate('url')
         self.url = url_template.render()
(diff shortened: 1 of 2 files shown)


Arksine:
Thanks.  I have a potential security concern with this option.  From my understanding its possible to attach files on the file system as well.  Put differently, it would possible to configure this to attach the `secrets` file or the database which can contain user credentials.

pataar:
> Thanks. I have a potential security concern with this option. From my understanding its possible to attach files on the file system as well. Put differently, it would possible to configure this to attach the `secrets` file or the database which can contain user credentials.

Hmm good point, however it would be nice to add log files as attachment. Could we create a validation for this? Or rather force urls? What would you think is best?

Arksine:
URLs are ok, and I think any path registered to the file manager is ok.  There isn't a great way to validate this using the file manager at the moment, so there is some work to do on my end.  Also this security issue is only exposed if the user leaves a client open in an unsecured environment.  

Now that I think about it, it would be possible for a malicious actor to define a notification containing secrets in the title and/or body to steal known credentials.  They could also publish secrets to an MQTT broker of their choice.  There isn't anything I can do about that, so I should probably update the documentation outlining this risk  imploring people to be careful (and use unique credentials in `secrets.ini`.

Give me a day or so to implement a validation method in the file manager then we can proceed to merge.  At least we can protect the database and other user files that way.

pataar:
Alright! Sounds good. Let me know when you're ready.

pataar:
> We can add a warning, but lets add a flag so it only warns once.

I added the validation.

Arksine:
Thanks.  As I mentioned above, we can't reliably check a file for access in `__init__()`.  The reason is that some folders are registered after after initialization, specifically the`gcode` folder which is provided by Klipper.  It is also possible for gcode folder to be moved if Klipper is reconfigured and restarted.

Thus we need to check the attachment in `NotifierInstance.notify()` and make decision as to what to do there.

pataar:
Thanks. I moved the validation. I think still notifying, but unsetting the attachment would be the best solution.

moonraker/components/notifier.py
```diff
@@ -139,9 +141,25 @@ async def notify(self, event_name: str, event_args: List) -> None:
...
+            if not self.attach.startswith("http://") and \
+               not self.attach.startswith("https://"):
+                fm = self.server.lookup_component("file_manager")
+                if not fm.can_access_path(self.attach):
+                    if not self.warned:
+                        self.server.add_warning(
+                            f"""Attachment of notifier '{self.name}' is not
+                            valid. The location of the
+                            attachment is not
+                            accessible.""")
```
Arksine, line 155: We don't want to use a multi-line string here, as newlines and leading whitespace is included.  So this would evauate to something like:
```
Attachment of notifier '{self.name}' is not\n                        valid. The location of the\n                        attachment is not\n                        accessible.
```

Arksine:
I was just in the process of adding the comment about the multiline string.  FWIW we can still check if the attachment is a url in `__init__()`, we just can't perform the file check.

Arksine:
Thanks!

PR #407, opened 2022-03-31
feat: added system uptime to proc_stats

KwadFan:
Added system uptime to proc_stats.py

This could be handy in case you want to display the uptime of the host or events that relie on time since last reboot.

Signed-off-by: Stephan Wendel <me@stephanwe.de>

### DIFF
--- a/moonraker/components/proc_stats.py
+++ b/moonraker/components/proc_stats.py
@@ -36,6 +36,7 @@ STATM_FILE_PATH = "/proc/self/smaps_rollup"
 NET_DEV_PATH = "/proc/net/dev"
 TEMPERATURE_PATH = "/sys/class/thermal/thermal_zone0/temp"
 CPU_STAT_PATH = "/proc/stat"
+UPTIME_STAT_PATH = "/proc/uptime"
 STAT_UPDATE_TIME = 1.
 REPORT_QUEUE_SIZE = 30
 THROTTLE_CHECK_INTERVAL = 10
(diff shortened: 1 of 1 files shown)


Arksine:
Thanks.  Python has a method for retrieving uptime:

```python
time.clock_gettime(time.CLOCK_BOOTTIME)
```
I don't mind adding support for this and reporting in the response to `/machine/proc_stats`.  I don't think its necessary to add it to the event as its value is predictable once you have a reference.  If a client wants to stream uptime on its UI that can be done at whatever interval a client desires by simply adding time elapsed to the reference.

KwadFan:
Thank you, for your helpful respond.
I started recently with python. Steep learning curve :)

So, I used time instead of proc/uptime file.

Regards

KwadFan:
Hi again,
after some discussion and thinking twice about it, it is ridiculous to "spam" every second its uptime.

So, I have to think about a better place to get the Host boot time.
Therefor that PR should be closed.
Thank you for taking your time :)

Arksine:
No problem. FWIW,I just pushed  9b5b3cc506a6412343047a2887026315e89c1333 which will report `system_uptime` in the response to the `/machine/proc_stats` request.

If you want to continuously update it on the UI you should be able to request that value as the reference or `start_value`, record the local time when you received it, and use that to calculate the time elapsed for each update.  Then just add the time elapse to the initial uptime.

PR #418, opened 2022-04-12
Update Sample Moonraker.conf

hdtechk:
Was trying to help someone in discord get passed his user password after update locked him out.
he was trying to edit this one instead of the one in Klipper_config the changes he made should have helped but didn't


Arksine:
Thanks.  Since it isn't guaranteed that the configuration file is located at `/home/pi/klipper_config/moonraker.conf` I made a slightly different change and pushed it in commit 59f2736bb66b36fe6d0c7e96aa5edb8d214bc5c6.

hdtechk:
Glade you found a better way.

PR #426, opened 2022-05-01
wled: Add extra control for wled effects

richardjm:
For the "Percent" preset the intensity controls the amount of the bar lit up 0-100
Work around issue when preset is re-enabled if sending intensity or speed while no preset is active

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1487,6 +1487,20 @@ gcode:
                              state=True,
                              preset=preset)}
 
+[gcode_macro WLED_CONTROL]
+description: Control effect values and brightness
+gcode:
+  {% set strip = params.STRIP|default('lights')|string %}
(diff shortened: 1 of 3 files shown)


Arksine:
Thanks.  It appears to me that this adds a new API and changes the behavior of existing APIs, thus it is also necessary to update the API documentation in accordance with the changes.

richardjm:
Do you mean adding a section into web_api.md? I don't believe existing behaviour should be affected as if the arguments aren't specified they default to the values used previously.

Arksine:
The `/machine/wled/control` appears to be a new endpoint.  With regard to new arguments, even if the default behavior does not change, we do need to document all arguments available to endpoints...otherwise developers will not know they are there.

richardjm:
Well I've the documentation and playing with api I've removed the `control` endpoint but added a `toggle` that was missed before.

Arksine:
Thanks.

PR #444, opened 2022-06-02
Fix: update manager klipper host_repo incorrect

Neko-vecter:
update manager klipper host_repo incorrect
Change line 42 from `"host_repo": "arksine/moonraker"` to `"host_repo": "Klipper3d/klipper"`

### DIFF
--- a/moonraker/components/update_manager/base_config.py
+++ b/moonraker/components/update_manager/base_config.py
@@ -39,7 +39,7 @@ BASE_CONFIG: Dict[str, Dict[str, str]] = {
         "requirements": "scripts/klippy-requirements.txt",
         "venv_args": "-p python2",
         "install_script": "scripts/install-octopi.sh",
-        "host_repo": "arksine/moonraker",
+        "host_repo": "Klipper3d/klipper",
         "managed_services": "klipper"
     }
 }


Arksine:
Hi.  While I understand why you viewed this as an incorrect option, it is actually correct.  It applies to "zip" distributions.  While I haven't released one for Moonraker/Klipper, when I eventually do both zip files will be hosted in Moonraker's repo, as the build is done by CI.

Neko-vecter:
> Hi. While I understand why you viewed this as an incorrect option, it is actually correct. It applies to "zip" distributions. While I haven't released one for Moonraker/Klipper, when I eventually do both zip files will be hosted in Moonraker's repo, as the build is done by CI.

I see. But what is the advantage of doing this way.

Arksine:
1) While the zip implementation is there, I have not yet created a zip release as Moonraker hasn't yet reached stable.  Git is workable but is not the best choice for deploying a stable release IMO.  Its prone to errors and most users do not understand git (which is reasonable, git a a developers tool).  Automated recovery  is implemented but iffy as new an unexpected errors may arise.

2) A "zip" release is "packaged" specifically for Moonraker, its not simply a zipped version of the source.  The Github Action used to package a release for Moonraker's update manager does not exist in Klipper's repo, and Kevin has his own method of managing releases.  Thus the zipped packages for both Moonraker and Klipper will need to be hosted here.

3) Since Moonraker depends on Klipper it makes sense to guarantee the release is in sync.  That way it can be assured that the stable releases work correctly together.

With regard to the idea of syncing `git` and `zip` releases, I'm not sure I see the purpose.  If a user wants to update via git they can just stay on the dev or beta channels.

Neko-vecter:
> 1. While the zip implementation is there, I have not yet created a zip release as Moonraker hasn't yet reached stable.  Git is workable but is not the best choice for deploying a stable release IMO.  Its prone to errors and most users do not understand git (which is reasonable, git a a developers tool).  Automated recovery  is implemented but iffy as new an unexpected errors may arise.
> 2. A "zip" release is "packaged" specifically for Moonraker, its not simply a zipped version of the source.  The Github Action used to package a release for Moonraker's update manager does not exist in Klipper's repo, and Kevin has his own method of managing releases.  Thus the zipped packages for both Moonraker and Klipper will need to be hosted here.
> 3. Since Moonraker depends on Klipper it makes sense to guarantee the release is in sync.  That way it can be assured that the stable releases work correctly together.
> 
> With regard to the idea of syncing `git` and `zip` releases, I'm not sure I see the purpose. If a user wants to update via git they can just stay on the dev or beta channels.

the idea just an option in `moonraker.conf` let people to choose which one they want using or sync with.

Also thanks you answer this question. Have a good day! 

By Neko.vecter

Arksine:
Ah, I understand.  Yes, its possible to stay on git by configuring the `channel` option, no one will every be forced onto stable and git updates will remain.  The default value will likely change, but that will be a progression (Moonraker needs to move into `beta` first)

Neko-vecter:
> Ah, I understand. Yes, its possible to stay on git by configuring the `channel` option, no one will ever be forced onto stable and git updates will remain. The default value will likely change, but that will be a progression (Moonraker needs to move into `beta` first)

Since I am doing some localization work. So that why I ask this question. Because there have problem with GitHub access in my country. So we move the repo into `https://gitee.com/miroky/moonraker.git` and make some change let update can using our mirror.

Neko-vecter:
So I think problem on `update manager klipper host_repo incorrect` is solved. 

Have a good day 

Send by Neko.vecter

PR #465, opened 2022-07-17
Following polkit versioning scheme change

compihu:
Polkit changed its versioning scheme starting with version 121 (preceding version was 0.120) causing set-policikit-rules.sh to fail.


compihu:
Accidental set -x remained in the script. Will fix and re-submit

PR #466, opened 2022-07-17
Following polkit versioning scheme change

compihu:
Polkit changed its versioning scheme starting with version 121 (the preceding version was 0.120) causing set-policykit-rules.sh to fail.

### DIFF
--- a/scripts/set-policykit-rules.sh
+++ b/scripts/set-policykit-rules.sh
@@ -45,7 +45,7 @@ add_polkit_rules()
         echo "PolicyKit not installed"
         exit 1
     fi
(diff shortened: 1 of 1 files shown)


Arksine:
Thanks.  The change looks good, however to merge all commits need a signature indicating that you accept the DCO as outlined in the [contributing documentation](https://moonraker.readthedocs.io/en/latest/contributing/#git-commit-format).  It would be fine for you to add the signature to the PR as well, I can modify the commit message.

This is a small change, so I don't mind making it myself if this is something that you aren't comfortable with.

Arksine:
Thanks!

PR #468, opened 2022-07-19
Allow configuring the virtualenv location

kageurufu:
Also allow setting defaults for installation command line
flags from the environment

```sh
MOONRAKER_VENV=$HOME/.venvs/moonraker \
MOONRAKER_REBUILD_ENV=y \
MOONRAKER_FORCE_DEFAULTS=n \
MOONRAKER_DISABLE_SYSTEMCTL=n \
MOONRAKER_CONFIG_PATH=$HOME/printer/moonraker.conf \
MOONRAKER_LOG_PATH=/tmp/moonraker.log \
./scripts/install-moonraker.sh
```

Personally, I set `MOONRAKER_VENV` and `MOONRAKER_CONFIG_PATH` in my .profile

Signed-off-by: Franklyn Tackitt <git@frank.af>

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -136,6 +136,16 @@ particularly for those upgrading:
   This is useful for installations that occur outside of a standard environment
   where systemd is not running.
 
+Additionally, installation may be customized with the following environment
+variables:
+
+- `MOONRAKER_VENV`
+- `MOONRAKER_REBUILD_ENV`
+- `MOONRAKER_FORCE_DEFAULTS`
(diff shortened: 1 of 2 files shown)


Arksine:
Overall I'm ok with the change, but I don't think we should change the default install location for moonraker's env.  If someone uses `install-moonraker.sh` to upgrade the script will create a new one, which isn't where the systemd unit will expect it to be.

kageurufu:
Oops, did not mean to include that bit. Fixed!

Arksine:
Thanks!

PR #475, opened 2022-07-30
Skip dbus_manager init without /proc

lotuspar:
(without /proc/self/stat)

dbus_manager: Skip init if /proc/self/stat isn't readable to prevent FileNotFoundError crash on iOS (and probably other non-Linux environments?)

Signed-off-by: Gianni S (lotuspar0@gmail.com)

### DIFF
--- a/moonraker/components/dbus_manager.py
+++ b/moonraker/components/dbus_manager.py
@@ -34,15 +34,18 @@ class DbusManager:
         self.bus: Optional[MessageBus] = None
         self.polkit: Optional[ProxyInterface] = None
         self.warned: bool = False
-        proc_data = pathlib.Path(f"/proc/self/stat").read_text()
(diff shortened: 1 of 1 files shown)


Arksine:
Thanks.  I can see where its useful to make sure that the file exists, so I implemented this in a slightly different manner in commit 0188b81d4aec29ede081156a7defc3bfab5a6a55.

I suspect that you will encounter other issues when attempting to run on iOS, BSD, etc.  At this time Moonraker isn't intended to be a cross-platform application, Linux is required.  That said, if you have it running in any capacity on iOS then that is good to know.

lotuspar:
The new commit looks good!

Also yes, I've definitely come across a lot of issues with running it on iOS. I might try to chip away at the issues until it works (and maybe that'd give the added benefit of Moonraker supporting more platforms?)

Thanks!

PR #480, opened 2022-08-08
metadata: adds chamber temp support

pedrolamas:
Adds Chamber Temp metadata reading for all slicers with these exceptions:

- Kiri:Moto: chamber temp doesn't seem to be supported by the slicer
- Simplify3D: I don't have a license, so no idea how to add this one


Arksine:
Thanks, looks good.

Issue #484, opened 2022-08-11
Add new "public" root


pedrolamas:
### Is your feature request related to a problem? Please describe

Fluidd and Mainsail have custom theme capabilities, both based on allowing users to create specially named subfolders on the "config" root, and in it put the theme files (css, jpg, etc...)

If one has authentication enabled and the user has yet to authenticate, it is impossible to retrieve any files from the "config" root (we get a 401).

### Describe the solution you'd like

So what I've been thinking is to have a new root called "public", that is read-only for unauthenticated users, and read & write for authenticated ones.

### Describe alternatives you've considered

Alternatively, have a "public" subfolder in the config that allows retrieving files even if the user is unauthenticated.

PR #485, opened 2022-08-11
Adds support for an optional "public" root

pedrolamas:
Proof of concept for #484

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -85,6 +85,12 @@ log_path:
 #   An optional path to a directory where log files are located.  Users may
 #   configure various applications to store logs here and Moonraker will serve
 #   them at "/server/files/logs/*".  The default is no log paths.
+public_path:
+#   An optional path to a directory where public files are located.  Users may
+#   configure various applications to store files here and Moonraker will serve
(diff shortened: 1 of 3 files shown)


Arksine:
Thanks.  Currently I'm modifying how Moonraker manages files and it impacts this very section of code in the file manager.  I'll soon post more information on that, but the short of it is that I will be centralizing Moonraker's file storage into one folder and removing all of the "path" options from the configuration.   It will include module that automates the changes necessary so that it does not break common installations.

With regard to themes or extensions, I think that this is something we need to discuss and come up with a robust solution.  I'd prefer to avoid adding an additional root with full access, this will expose a security vulnerability that my current work intends to close.  I think an interim solution here would be a git repo that contains curated themes, served by the reverse proxy.  A symbolic link could be added to the client folder, Moonraker can be configured to avoid removing that link when performing a client update.

pedrolamas:
Thanks for the update @Arksine!

I think for now we will just apply on Fluidd the same solution Mainsail is using (have the files stored on the "config" root under a hidden folder) and revisit this once you make the changes on the file manager.

PR #493, opened 2022-08-27
webcam: fixed rotation param in api, str -> int

grantrules:
Signed-off-by: Grant Harding <gharding@gmail.com>


Arksine:
Good catch.  Can you include your last name in your signature?  Thanks.

grantrules:
> Good catch. Can you include your last name in your signature? Thanks.

Done

Arksine:
Thanks.

PR #495, opened 2022-08-30
notify: add server endpoints for notifiers to list and test them

pataar:
This adds two endpoints:

### Viewing notifiers

Untitled

### Testing notifiers

Untitled 2

### DIFF
--- a/docs/web_api.md
+++ b/docs/web_api.md
@@ -3375,6 +3375,65 @@ Returns: Test results in the following format
 }
 ```
 
+### Notifier APIs
+The following APIs are available to view and tests notifiers.
+
+#### List Notifiers
+
+HTTP request:
+```http
+GET /server/notifiers/list
+```
+JSON-RPC request:
+```json
+{
+    "jsonrpc": "2.0",
+    "method": "server.notifiers.list",
+    "id": 4654
+}
+```
+
+Returns:
+
+A list of configured notifiers:
+
+```json
+{
+    "notifiers": [
+        {
+            "name": "print_start",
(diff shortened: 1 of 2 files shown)


Arksine:
If there is a use case for listing notifiers in front ends then I don't have an issue adding support for that endpoint.  Moonraker shouldn't register debug endpoints as part of the API though, as it would be invalid for a front end to force an event in general usage.

Generally I just keep local branches when I want to register a test endpoint.  After Moonraker releases its next beta I may look into providing the ability to register debug endpoints.

pataar:
I could imagine a place in the frontend to give insight in the notifiers, comparable to the current update manager. Instead of an update button, they would have a test button.

Although, If this use case is a bit far fetched, feel free to close the PR. It was just an idea that popped up.

Arksine:
Ok.  What I'm thinking is that I will add an option for "debug" endpoints in the future enabled using a command line option.    When that is done we can revisit this.  I think it will be ok to keep the list endpoint as it is, and make the test endpoint a "debug" endpoint after I add support for them.

pataar:
Alright, sounds good. I'll update this PR when you're done with that.

Arksine:
FYI - debug endpoints are now merged.  The `database.py` module has an example of how to use them, the path should always start with `/debug`, ie `/debug/notifier/test`.  Its ok for the list to be a normal endpoint.

Debug endpoints are enabled when `-g` is added to the command line.  This is now accomplished by editing `~/printer_data/systemd/moonraker.env`.

pataar:
@Arksine Thanks, that's done!

Arksine:
Thanks!

PR #496, opened 2022-08-31
octoprint_compat: added job status & operation

grantrules:
Added job status & operation to complete compatibility with Cura-OctoPrintPlugin. Used simplyprint.py as a guide.

Signed-off-by: Grant Harding <gharding@gmail.com>

### DIFF
--- a/moonraker/components/octoprint_compat.py
+++ b/moonraker/components/octoprint_compat.py
@@ -20,6 +20,8 @@ if TYPE_CHECKING:
     from .klippy_apis import KlippyAPI as APIComp
     from .file_manager.file_manager import FileManager
     from .job_queue import JobQueue
+    from components.file_manager.file_manager import FileManager
+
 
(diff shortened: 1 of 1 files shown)


grantrules:
is there a way to return an HTTP error from an endpoint callback?

Arksine:
Thanks.  The purpose `octoprint_compat` plugin is to present the bare minimum API required for upload functionality.  I'm weary of extending Octoprint compatibility beyond that due to the additional maintenance burden involved.  My preference would be for the Cura plugin to use Moonraker's native API if it wants to implement print monitoring.

grantrules:
If that's the case, I may break it out into a plugin. I may also take a look into forking Cura-OctoPrint. There is a moonraker plugin for cura but it is very far from complete.

grantrules:
There now it's a plugin. Maybe it'll be useful for somebody else, too.

https://github.com/grantrules/moonraker-octoprint-enhanced

PR #497, opened 2022-08-31
notify: add ability to notify using a remote method

pataar:
This way you can create macros that can be used to notify services through gcode.

Example:
```ini
# printer.cfg

[gcode_macro NOTIFY_FILAMENT_CHANGE]
gcode:
  {action_call_remote_method("notify",
                             name="telegram",
                             message="Filament change needed!")}
```

The body is overridable. 

Let me know what you think of this. Happy to receive any feedback.

### DIFF
--- a/moonraker/components/notifier.py
+++ b/moonraker/components/notifier.py
@@ -35,6 +35,7 @@ class Notifier:
         prefix_sections = config.get_prefix_sections("notifier")
 
         self.register_events(config)
+        self.register_remote_actions()
 
         for section in prefix_sections:
             cfg = config[section]
@@ -52,6 +53,16 @@ class Notifier:
                 self.server.add_warning(msg)
                 continue
             self.notifiers[notifier.get_name()] = notifier
+
+    def register_remote_actions(self):
+        self.server.register_remote_method("notify", self.notify_action)
+
+    async def notify_action(self, name: str, message: str = ""):
+        if name not in self.notifiers:
+            raise self.server.error(f"Notifier '{name}' not found", 404)
+        notifier = self.notifiers[name]
+
+        await notifier.notify("remote_action", [], message)
 
     def register_events(self, config: ConfigHelper):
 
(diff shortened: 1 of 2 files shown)


moonraker/components/notifier.py
```diff
@@ -53,6 +54,21 @@ def __init__(self, config: ConfigHelper) -> None:
...
+        if name not in self.notifiers:
+            raise self.server.error(f"Notifier '{name}' not found", 404)
+        notifier = self.notifiers[name]
+
+        kapis: APIComp = self.server.lookup_component('klippy_apis')
+        result: Dict[str, Any] = await kapis.query_objects(
+            {'print_stats': None}, default={})
+        print_stats = result.get('print_stats', {})
+
+        await notifier.notify("remote_action", [print_stats, print_stats], body)
```
Arksine, line 70: I don't think we need to include print_stats here.  If Klipper is going to define the body that is what should be sent.  No need to include print stats in the jinja context since its not triggered by a print event.

pataar: Good point. What do you think would be a better fit to pass through?

Arksine: Sorry for the delay, life getting a bit in the way.  

I think the remote method could take a "message" argument and we could pass it through to the context.  So for example:
```python
async def notify_action(self, name: str, message: str = ""):
    if name not in self.notifiers:
        raise self.server.error(f"Notifier '{name}' not found", 404)
    notifier = self.notifiers[name]
    await notifier.notify("remote_action", [], message)
```

Then in the `NotifierInstance` class
```python
async def notify(
        self, event_name: str, event_args: List, message: str = ""
) -> None:
    context = {
        "event_name": event_name,
        "event_args": event_args,
        "event_message": message
    }
```

The user could then render the `event_message` in the template.

pataar: No problem! No need to hurry. Good feedback! I updated the code accordingly.

Arksine:
Looks good, thanks

PR #500, opened 2022-09-12
Git submodules use a .git file instead of a directory

kageurufu:
I'm playing with using git submodules to track my entire printer config,
 and moonraker only works with separate clones as is. Using `.exists()`
 instead of `.is_dir()` allows moonraker to control the submodules.

### DIFF
--- a/moonraker/components/update_manager/git_deploy.py
+++ b/moonraker/components/update_manager/git_deploy.py
@@ -562,7 +562,7 @@ class GitRepo:
 
     async def update_repo_status(self) -> bool:
         async with self.git_operation_lock:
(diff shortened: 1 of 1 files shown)


Arksine:
Do submodules work reliably with this change?  If so, I'm ok with adding it.  I would just need you to add the `Signed-off` line before I can merge.

kageurufu:
Submodules work fine on my printer so far, I can update through moonraker still, and I tested against a normal checkout to make sure that didn't break

Updated the PR with a signoff :)

Arksine:
Thanks.  I thought that submodules required a special procedure, but I suppose that is if you want to update them from the parent.

PR #508, opened 2022-09-18
Feature: Generic sensor support and data logging

mjonuschat:
This feature implements a sensor component that can be used to track/log generic sensors from multiple sources. Each sensor can have properties like unit of measurement, accuracy and a display name that help frontends display the tracked measurements.

Some supported sources could be locally connected I2C, SPI or 1-wire sensors as well as remote sensor providers like MQTT or REST API driven sensor.

Example of the component in use with a ESPHome device connected through MQTT:

```
[sensor.py:component_init()] - Initializing sensor component
[sensor.py:initialize()] - Registered sensor 'Chamber Temperature'
[sensor.py:initialize()] - Registered sensor 'Chamber Humidity'
[sensor.py:initialize()] - Registered sensor 'Power Usage'
[sensor.py:component_init()] - All sensors have been initialized
...
[mqtt.py:__call__()] - MQTT Subscriptions Acknowledged
 Topic: devvm-mjonuschat.sharks-with-lasers.net/moonraker/api/request | Granted QoS 0
 Topic: hank/sensor/chamber_temperature/state | Granted QoS 0
 Topic: hank/sensor/chamber_humidity/state | Granted QoS 0
 Topic: hank/sensor/pzem-004t_v3_power/state | Granted QoS 0
[sensor.py:_on_state_update()] - Context: {'payload': '27.7'}
[sensor.py:_on_state_update()] - Received updated sensor value for Chamber Temperature: 27.7°C
[sensor.py:_on_state_update()] - Context: {'payload': '44.9'}
[sensor.py:_on_state_update()] - Received updated sensor value for Chamber Humidity: 44.9%
[sensor.py:_on_state_update()] - Context: {'payload': '1.60'}
[sensor.py:_on_state_update()] - Received updated sensor value for Power Usage: 1.6W
...
```
Querying the aggregated endpoint to retrieve sensor values:
```
{
  "result": {
    "chamber_temperature": {
      "source": "hank/sensor/chamber_temperature/state",
      "name": "Chamber Temperature",
      "unit_of_measurement": "°C",
      "accuracy_decimals": 1,
      "measurements": [
        0,
        27.7,
        27.7,
        27.7,
        27.7,
        27.7,
        27.7,
        27.7,
        27.7,
        27.7,
        27.7,
        27.7,
        27.7
      ]
    },
    "chamber_humidity": {
      "source": "hank/sensor/chamber_humidity/state",
      "name": "Chamber Humidity",
      "unit_of_measurement": "%",
      "accuracy_decimals": 1,
      "measurements": [
        0,
        44.9,
        44.9,
        44.9,
        44.9,
        44.9,
        44.9,
        44.9,
        44.9,
        44.9,
        44.9,
        44.9,
        44.9
      ]
    },
    "power_usage": {
      "source": "hank/sensor/pzem-004t_v3_power/state",
      "name": "Power Usage",
      "unit_of_measurement": "W",
      "accuracy_decimals": 2,
      "measurements": [
        0,
        1.6,
        1.6,
        1.6,
        1.6,
        1.6,
        1.6,
        1.6,
        1.6,
        1.6,
        1.6,
        1.6,
        1.6
      ]
    }
  }
}
```

Signed-off-by: Morton Jonuschat <mjonuschat+moonraker@gmail.com>

### DIFF
--- a/docs/web_api.md
+++ b/docs/web_api.md
@@ -4915,6 +4915,154 @@ State of the strip.
 }
 ```
 
+### Sensor APIs
+The APIs below are available when the `[sensor]` component has been configured.
+
+#### Get Sensor List
+HTTP request:
+```http
+GET /server/sensors/list
+```
+JSON-RPC request:
+```json
+{
+    "jsonrpc": "2.0",
+    "method":"machine.device_power.devices",
+    "id": 5646
+}
+```
+Returns:
+
+An array of objects containing info for each configured sensor.
+```json
+{
+    "sensors": {
+        "sensor1": {
+            "id": "sensor1",
+            "friendly_name": "Sensor 1",
+            "type": "mqtt",
+            "values": {
+                "value1": 0,
+                "value2": 119.8
+            }
+        }
+    }
+}
+```
+
+#### Get Sensor Information
+Returns the status for a single configured sensor.
+
+HTTP request:
+```http
+GET /server/sensors/list?sensor=sensor1
+```
+JSON-RPC request:
+```json
+{
+    "jsonrpc": "2.0",
+    "method": "machine.device_power.get_device",
+    "params": {
+        "sensor": "sensor1"
+    },
+    "id": 4564
+}
+```
+Returns:
+
+An object containing sensor information for the requested sensor:
+```json
+{
+    "id": "sensor1",
+    "friendly_name": "Sensor 1",
+    "type": "mqtt",
+    "values": {
+        "value1": 0.0,
+        "value2": 120.0
+    }
+}
+```
+
+#### Get Sensor Measurements
+Returns all recorded measurements for a configured sensor.
+
+HTTP request:
+```http
+GET /server/sensors/measurements?sensor=sensor1
+```
+JSON-RPC request:
+```json
+{
+    "jsonrpc": "2.0",
+    "method": "server.sensors.measurements",
+    "params": {
+        "sensor": "sensor1"
+    },
+    "id": 4564
+}
+```
+Returns:
+
+An object containing all recorded measurements for the requested sensor:
+```json
+{
+    "sensor1": {
+        "value1": [
+            3.1,
+            3.2,
+            3.0
+        ],
+        "value2": [
+            120.0,
+            120.0,
+            119.9
+        ]
+    }
+}
+```
+
+#### Get Batch Sensor Measurements
+Returns recorded measurements for all sensors.
+
+HTTP request:
+```http
+GET /server/sensors/measurements
+```
+JSON-RPC request:
+```json
+{
+    "jsonrpc": "2.0",
+    "method": "server.sensors.measurements",
+    "id": 4564
+}
+```
+Returns:
+
+An object containing all measurements for every configured sensor:
+```json
+{
+    "sensor1": {
+        "value1": [
+            3.1,
+            3.2,
+            3.0
+        ],
+        "value2": [
+            120.0,
+            120.0,
+            119.9
+        ]
+    },
+    "sensor2": {
+        "value_a": [
+            1,
+            1,
+            0
+        ]
+    }
+}
+```
+
 ### OctoPrint API emulation
 Partial support of OctoPrint API is implemented with the purpose of
 allowing uploading of sliced prints to a moonraker instance.
@@ -6289,6 +6437,30 @@ for that agent, with its identity info in the `data` field.  When an agent
 disconnects clients will receive a `disconnected` event with the data field
 omitted.  All other events are determined by the agent, where each event may
 or may not include optional `data`.
+
+#### Sensor Events
+
+Moonraker will emit a `sensors:sensor_update` notification when a measurement
+from at least one monitored sensor changes.
+
+```json
+{
+    "jsonrpc": "2.0",
+    "method": "sensors:sensor_update",
+    "params": [
+        {
+            "sensor1": {
+                "humidity": 28.9,
+                "temperature": 22.4
+            }
+        }
+    ]
+  ```
+
+When a sensor reading changes, all connections will receive a
+`sensors:sensor_update` event where the params contains a data struct
+with the sensor id as the key and the sensors letest measurements as value
+struct.
 
 ### Appendix
 
--- a/docs/web_api.md
+++ b/docs/web_api.md
@@ -4927,7 +4927,7 @@ JSON-RPC request:
 ```json
 {
     "jsonrpc": "2.0",
-    "method":"machine.device_power.devices",
+    "method":"server.sensors.list",
     "id": 5646
 }
 ```
@@ -4955,13 +4955,13 @@ Returns the status for a single configured sensor.
 
 HTTP request:
 ```http
-GET /server/sensors/list?sensor=sensor1
+GET /server/sensors/info?sensor=sensor1
 ```
 JSON-RPC request:
 ```json
 {
     "jsonrpc": "2.0",
-    "method": "machine.device_power.get_device",
+    "method": "/server/sensors/info?sensor=sensor1",
     "params": {
         "sensor": "sensor1"
     },
@@ -6455,7 +6455,7 @@ from at least one monitored sensor changes.
             }
         }
     ]
-  ```
+```
 
 When a sensor reading changes, all connections will receive a
 `sensors:sensor_update` event where the params contains a data struct
--- a/moonraker/components/sensor.py
+++ b/moonraker/components/sensor.py
@@ -189,7 +189,7 @@ class Sensors:
             self._handle_sensor_list_request,
         )
         self.server.register_endpoint(
-            "/server/sensors/sensor",
+            "/server/sensors/info",
             ["GET"],
             self._handle_sensor_info_request,
         )
@@ -294,7 +294,7 @@ class Sensors:
         output = {
             key: sensor.get_sensor_measurements()
             for key, sensor in self.sensors.items()
-            if sensor_name is None or key == sensor_name
+            if sensor_name is "" or key == sensor_name
         }
 
         return output
--- a/moonraker/components/sensor.py
+++ b/moonraker/components/sensor.py
@@ -8,13 +8,23 @@
 # available to clients
 from __future__ import annotations
 
-import asyncio
 import logging
+from collections import defaultdict, deque
 from dataclasses import dataclass, replace
-from collections import deque
+from functools import partial
 
 # Annotation imports
-from typing import Any, Dict, List, Optional, Type, Deque, TYPE_CHECKING
+from typing import (
+    Any,
+    DefaultDict,
+    Deque,
+    Dict,
+    List,
+    Optional,
+    Type,
+    TYPE_CHECKING,
+    Union,
+)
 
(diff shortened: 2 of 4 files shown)


Arksine:
Thanks.  I'm open to discussing generic sensor support, however the `aioesphomeapi` dependency will create an issue.  It depends on `noiseprotocol`, which depends on `cryptography`.  Unfortunately the Rust bindings in `cryptography` will cause a problem on SBCs that don't have prebuilt wheels available (ie: Armbian).  The wheel won't build and Moonraker will fail to update.  Initial support for ESPHome devices would probably need to be accomplished via MQTT.  In the future when Moonraker reaches stable we could revisit this, as distribution and updates for stable releases would include prebuilt dependencies.

Separately, I think this can be accomplished though a single component.  Its acceptable to break the component into separate modules if necessary, that could be accomplished by making the component a package (see the `file_manager` and `update_manager` components as examples).   I don't believe we need a separate data store module though, it seems that each base sensor could store whatever data is necessary, then `sensor.py` could expose endpoints for querying the data.

moonraker/components/sensor.py
```diff
@@ -0,0 +1,232 @@
...
+    ) -> Dict[str, Dict[str, Any]]:
+        data = {}
+        for sensor in self.sensors.values():
+            data[sensor.config.id] = {
+                "source": sensor.config.source,
+                "name": sensor.config.name,
+                "unit_of_measurement": sensor.config.unit,
+                "accuracy_decimals": sensor.config.accuracy_decimals,
+                "measurements": list(sensor.values),
+            }
```
Arksine, line 223: With the above recommendations this would probably report a result something like the following:
```python
{
    "bme280": {
        "friendly_name": "bme_280",
        "values": {
            "temperature": 21.4,
            "humidity": "33"
        }
    }
}
```

Likewise, an endpoint that reports the store might look like the following:
```python
{
    "bme280": {
        "temperature": [21.4, 22.0, 21.8]
        "humidity": ["33.1, 33.1, 33.0]
    }
}
```

moonraker/components/sensor.py
```diff
@@ -0,0 +1,232 @@
...
+
+    async def component_init(self) -> None:
+        try:
+            logging.debug("Initializing sensor component")
+            event_loop = self.server.get_event_loop()
+            cur_time = event_loop.get_loop_time()
+            endtime = cur_time + 120.0
+
+            uninitialized_sensors = list(self.sensors.values())
+            while cur_time < endtime:
```
Arksine, line 185: This logic was taken from the `power` module before it was refactored.  There is no longer a loop in `component_init`, instead each sensor's init routine is executed asynchronously.  This init procedure is indefinite, however it does log failures.  This was done for two reasons:

1) The server won't start until `component_init` completes for each component.  We want to avoid delaying it.
2) When the machine first boots this can cause failures because the network isn't up yet.

That said, the `mqtt` implementation of initialize doesn't event need to be async.  It doesn't await anything and init occurs when it the subscription data is received.  The above will only be necessary for implementations that require polling should they be added.

Arksine:
Thanks, the configuration documentation provided me with a more generalized idea of what you are looking to accomplish.  I think this module adds value, and I have added some review comments.

With regard to the API, we probably need a couple of endpoints.  One that returns current device info (including its current value), and one that returns the queue stored values for each device.  I touched on what that might look like in the review comments.  The `info` endpoint should have an option for consumers to specify a specific device.

Additionally, most clients want changes in the value pushed to them over the websocket.  In the review I mentioned having a single timer that updates the queue for each registered sensor.  This timer could also push emit a `sensor_update` event that is registered as a websocket notification.  This event would only include sensors that detected a change, if no sensor values change then it wont be emitted.  A one second timer is probably appropriate.

moonraker/components/sensor.py
```diff
@@ -0,0 +1,232 @@
...
+    def __init__(
+        self, name: str, cfg: ConfigHelper, store_size: int = 1200
+    ) -> None:
+        self.server = cfg.get_server()
+        self.error_state: Optional[str] = None
+
+        self.config = SensorConfiguration(
+            id=name,
+            name=cfg.get("name", name),
+            unit=cfg.get("unit", ""),
```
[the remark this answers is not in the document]
Arksine: After some additional thought, I'm not sure we need a generic `fields` option either.  The template can set the fields for the value.  We could pass a method into the context called `set_result` that takes a dict and assigns the current value.  

The template could also be responsible for rounding the result, which would remove the need for the `accuracy_decimals` (or `precision`) option.

mjonuschat:
@Arksine : I think I've addressed most of the comments / implemented the suggested changes. Have not yet tested this fully, wanted to check in to see that this is going in the direction you were thinking.
In regards to the notifications I have implemented those with the only notify when readings changed but it feels a bit counter-intuitive for sensors where you would expect to get a time-series with predictable intervals between the readings, even if say the temperature didn't change (unless it's expected that the clients "fill the gaps").

mjonuschat:
Tested and verified working with the example configuration and notification have been added

moonraker/components/sensor.py
```diff
@@ -0,0 +1,315 @@
...
+        """
+        Append the last updated value to the store.
+        """
+        for key, value in self.last_measurements.items():
+            self.values[key].append(value)
+
+        # Copy the last measurements data
+        self.last_value = {**self.last_measurements}
+
+        return eventtime + SENSOR_UPDATE_TIME
```
Arksine, line 91: It isn't necessary to return a value here since the timer is now in the main component.

Arksine:
It is close, just a couple of minor issues and the "notification bundling". Additionally the APIs and websocket notification will need documentation, then it will be ready to merge.

moonraker/components/sensor.py
```diff
@@ -0,0 +1,309 @@
...
+    TYPE_CHECKING,
+    Union,
+)
+
+if TYPE_CHECKING:
+    from confighelper import ConfigHelper
+    from websockets import WebRequest
+
+SENSOR_UPDATE_TIME = 1.0
+SENSOR_EVENT_NAME = ""
```
Arksine, line 34: Presumably this should be `sensors:sensor_update`?

Arksine:
Other than the missing sensor event name the changes to the source look good.  You need to rebase your branch against master to resolve the merge conflicts with `api_changes.md`, frankly you don't need to update that if you don't want.  That is mostly for breaking changes.  Very soon I plan to keep a single changelog.

The websocket notification needs to be documented in `web_api.md`.  Since you basically have it documented in `api_changes` you could move it with a few tweaks.

mjonuschat:
Requested changes have been addressed and changes have been rebased. Let me know if you want me to squash the changes before merging.

docs/web_api.md
```diff
@@ -6290,6 +6438,30 @@ disconnects clients will receive a `disconnected` event with the data field
...
+    "method": "sensors:sensor_update",
+    "params": [
+        {
+            "sensor1": {
+                "humidity": 28.9,
+                "temperature": 22.4
+            }
+        }
+    ]
+  ```
```
Arksine, line 6458: I believe the whitespace before the three backticks is causing an issue with rendering.  Also the parent json object is missing a closing `}`.

docs/web_api.md
```diff
@@ -4915,6 +4915,154 @@ State of the strip.
...
+#### Get Sensor List
+HTTP request:
+```http
+GET /server/sensors/list
+```
+JSON-RPC request:
+```json
+{
+    "jsonrpc": "2.0",
+    "method":"machine.device_power.devices",
```
Arksine, line 4930: This should presumably be `server.sensors.list`.

docs/web_api.md
```diff
@@ -4915,6 +4915,154 @@ State of the strip.
...
+    }
+}
+```
+
+#### Get Sensor Information
+Returns the status for a single configured sensor.
+
+HTTP request:
+```http
+GET /server/sensors/list?sensor=sensor1
```
Arksine, line 4958: Should be `/server/sensors/sensor?sensor=sensor1`?  Looking at that, perhaps the endpoint name should be `/server/sensors/info?sensor=<sensor_name>`.  The `sensors/sensor?sensor=` is a bit much.

docs/web_api.md
```diff
@@ -4915,6 +4915,154 @@ State of the strip.
...
+```
+JSON-RPC request:
+```json
+{
+    "jsonrpc": "2.0",
+    "method": "machine.device_power.get_device",
```
Arksine, line 4964: Presuming we change the endpoint as recommended, this should be `server.sensor.info`.

moonraker/components/sensor.py
```diff
@@ -0,0 +1,309 @@
...
+        self, web_request: WebRequest
+    ) -> Dict[str, Dict[str, Any]]:
+        sensor_name: str = web_request.get_str("sensor", "")
+        if sensor_name and sensor_name not in self.sensors:
+            raise self.server.error(f"No valid sensor named {sensor_name}")
+
+        output = {
+            key: sensor.get_sensor_measurements()
+            for key, sensor in self.sensors.items()
+            if sensor_name is None or key == sensor_name
```
Arksine, line 297: I may have missed this on the previous review, however `sensor_name` defaults to a null string on line 290.  This will result in an empty return value when a front end requests all measurements.

Arksine:
Thanks.  I found some issues with the docs, and apparently I missed a problem with the `measurements` request in my last review.  I also proposed refactoring the name of `/server/sensors/sensor` to `/server/sensors/info`, as the `/sever/sensors/sensor?sensor=` is a bit extreme on the use of the term "sensor".

mjonuschat:
Sorry for the force-push, forgot to add the Signed-off comment to a bunch of the commit messages.

Tested the endpoint changes and the bugfix:

```
pi@moonraker-dev:~ $ curl -s "http://localhost:7125/server/sensors/measurements" | jq
{
  "result": {
    "powermeter": {
      "power": [
        0,
        0,
        0
      ],
      "voltage": [
        115.7,
        115.7,
        115.7
      ],
      "current": [
        0,
        0,
        0
      ],
      "energy": [
        0,
        0,
        0
      ]
    }
  }
}
```

```
pi@moonraker-dev:~ $ curl -s "http://localhost:7125/server/sensors/measurements?sensor=powermeter" | jq
{
  "result": {
    "powermeter": {
      "power": [
        0,
        0
      ],
      "voltage": [
        115.4,
        115.4
      ],
      "current": [
        0,
        0
      ],
      "energy": [
        0,
        0
      ]
    }
  }
}
```

```
pi@moonraker-dev:~ $ curl -s "http://localhost:7125/server/sensors/info?sensor=powermeter"|jq
{
  "result": {
    "id": "powermeter",
    "friendly_name": "powermeter",
    "type": "mqtt",
    "values": {
      "power": 0,
      "voltage": 115.4,
      "current": 0,
      "energy": 0
    }
  }
}
```

moonraker/components/sensor.py
mjonuschat: Sorry, missed updating the condition when mypy complained about the None as sentinel value when retrieving the request param.

Arksine:
Thanks.

PR #511, opened 2022-09-22
metadata: strip delimiting quotes in strings

pedrolamas:
Trim any delimiting quotes from metadata strings.

I know we could just update the regular expressions to do this, but seems easier (and safer!) to just trim the quotes after finding the values.

Here's an example from a Prusa Slicer sliced file showing the problem:

image

### DIFF
--- a/moonraker/components/file_manager/metadata.py
+++ b/moonraker/components/file_manager/metadata.py
@@ -91,7 +91,7 @@ def _regex_find_int(pattern: str, data: str) -> Optional[int]:
 def _regex_find_string(pattern: str, data: str) -> Optional[str]:
     match = re.search(pattern, data)
     if match:
-        return match.group(1)
+        return match.group(1).strip('"')
     return None
 
 # Slicer parsing implementations


Arksine:
Thanks, I went ahead and merged. I don't see any harm in stripping quotes, I wouldn't think that front ends do anything but display this data.

pedrolamas:
I may have "jumped the gun" on this one....

Last week I discussed this change with @meteyou and he asked a great question: "what about multi-material?"

Well, as expected, different slicers have different behaviors:

## Cura

Cura is useless here, so it only shows a single material (the one it is active when we output the metadata)...

image

Stupidly, the "filament weight" can either be a single number when a single material was used, or an array of numbers when multiple materials were in use (thanks a lot, Cura!!!! 😩)

## PrusaSlicer and derivatives

Guess now I know what the quotes are for...

As seen below, multi-material prints in PrusaSlicer and derivatives are output as a list of quoted strings separated by semi-colon.

image

## Ideamaker

@meteyou tested this one, and we can see below that it outputs multiple entries for multi-material, though at the time Moonraker is only reading the first one.

image

----

So bottom line is that this change might make things worse, especially on a multi-material sliced g-code (so on Fluidd and Mainsail we would show something like `PLA";"PETG`)

This is definitely not a big deal, but something to be aware of and consider if any action is to be taken on it.

PR #522, opened 2022-10-15
Suggestion: use term "sudo" instead of "root"

th33xitus:
22-10-15_09-48-15_brave

Currently Moonraker asks for "root request" and the "root password".
That could be misleading in a way, that users might think moonraker wants access to the root user account (UID 0).

This PR rephrases the terms to "Sudo password" and "sudo" respectively. It also fixes another typo.
It might be a bit nitpicking and i am fine with this PR getting declined. 


Arksine:
Looks good, thanks.

PR #525, opened 2022-10-15
Add support for USB hub power control via uhubctl

Jckf:
This adds support for toggling power to certain USB hubs/ports on or off via the command line tool [uhubctl](https://github.com/mvp/uhubctl) (available via apt on Raspbian). This is useful to completely shut down MCUs, LCDs, and other peripherals that are powered over USB.

On my RPi 3 + RAMPS 1.4 printer setup, the following config lets me turn off the entire printer plus its webcam (port 2 controls power to all 4 ports on the RPi):

```ini
[power USB]
type: uhubctl
port: 1-1.2
bound_service: klipper
```

**Disclaimer:** I am not a Python developer, and the entire Klipper ecosystem is new to me as of a few days ago, but I've done my best 😅 Feedback welcome 🙂

Signed-off-by: Jim C K Flaten <jckf@jckf.no>

### DIFF
--- a/moonraker/components/power.py
+++ b/moonraker/components/power.py
@@ -1424,8 +1424,13 @@ class UhubctlDevice(PowerDevice):
         ports = {}
 
         for line in result:
-            if line[:6] == "Current":
+            first_word = line.split(" ")[0]
+
+            if first_word in [ "Current", "New" ]:
                 hub = line.split(" ")[4]
+                continue
+
+            if first_word == "Sent":
                 continue
 
             port, info = tuple(line.strip()[4:].split(": "))
--- a/moonraker/components/power.py
+++ b/moonraker/components/power.py
@@ -1433,7 +1433,7 @@ class UhubctlDevice(PowerDevice):
             if first_word == "Sent":
                 continue
 
-            port, info = tuple(line.strip()[4:].split(": "))
+            port, info = tuple(line.strip()[5:].split(": "))
             ports[hub + "." + port] = info.split(" ")
 
         id = self.hub + "." + self.port
--- a/moonraker/components/power.py
+++ b/moonraker/components/power.py
@@ -1459,8 +1459,6 @@ class UhubctlDevice(PowerDevice):
         else:
             self.state = "error"
 
-        logging.info(f"uhubctl device {self.name} entered {self.state} state")
-
         return
 
     async def set_power(self, state: str) -> None:
--- a/moonraker/components/power.py
+++ b/moonraker/components/power.py
@@ -1427,7 +1427,7 @@ class UhubctlDevice(PowerDevice):
                 hub = line.split(" ")[4]
                 continue
 
-            port, info = tuple(line.strip().[4:].split(": "))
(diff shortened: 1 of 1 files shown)


moonraker/components/power.py
```diff
@@ -1318,6 +1321,90 @@ async def _send_status_request(self) -> str:
...
+        elif result[1] == "off":
+            self.state = "off"
+        else:
+            self.state = "error"
+
+        return
+
+    async def set_power(self, state: str) -> None:
+        await self.refresh_status(self._port_status(state))
+        return
```
Arksine, line 1405: `set_power` and `refresh_status` should handle exceptions similar to how it the gpio implementation handles them.  That way we can catch any error, reset the state to `error`, then re-raise he exception so the front end gets an error.

Arksine:
Thanks.  I took an initial look and found a few things that should be addressed.

moonraker/components/power.py
```diff
@@ -1318,6 +1321,90 @@ async def _send_status_request(self) -> str:
...
 
 
+# USB Hub Control
+# https://github.com/mvp/uhubctl
+# apt install uhubctl
+class UhubctlDevice(PowerDevice):
+    def __init__(self, config: ConfigHelper) -> None:
+        super().__init__(config)
+
+        self.bin_path = config.get("bin_path", "uhubctl")
```
Jckf, line 1331:
```suggestion
```

moonraker/components/power.py
```diff
@@ -1318,6 +1321,90 @@ async def _send_status_request(self) -> str:
...
+        self.hub, self.port = tuple(config.get("port").split("."))
+        self.initial_state = config.getboolean("initial_state", True)
+
+    def _uhubctl(self, args: List[str] = []) -> Tuple:
+        proc = subprocess.Popen(
+            [self.bin_path] + args,
```
Jckf, line 1337:
```suggestion
            ["uhubctl"] + args,
```

moonraker/components/power.py
```diff
@@ -1318,6 +1321,90 @@ async def _send_status_request(self) -> str:
...
+
+        id = self.hub + "." + self.port
+
+        if id in ports:
+            return ports[id]
+
+        return []
+
+    async def init_state(self) -> None:
+        await self.set_power("on" if self.initial_state else "off")
```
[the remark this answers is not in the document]
Jckf: This was added because I always want my printer to be off unless explicitly commanded to be on, but maybe there are other ways to toggle power devices automatically at startup?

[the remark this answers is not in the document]
Arksine: I recently made `initial_state` a global option for power devices (I just haven't updated the docs yet).  It functions as you suggest, the `initial_state` defaults to `None` if not configured, in which case we do not change the state of the device on init.  The change to your code should be relatively simple, remove line 1332 as the base class has already fetched the option, and change `init_state()` to:
```python
async def init_state(self) -> None:
    if self.initial_state is not None:
        await self.set_power("on" if self.initial_state else "off")
```

PR #532, opened 2022-10-21
docs: Correct `minimum_event_time` example

Jckf:
The internal variable is called `min_`, but the config option is called `minimum_`.

Signed-off-by: Jim C K Flaten <jckf@jckf.no>

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1686,7 +1686,7 @@ pin: gpiochip0/gpio26
 #      ^!gpiochip0/gpio26
 #      ~!gpiochip0/gpio26
 #   This parameter must be provided
-min_event_time: .05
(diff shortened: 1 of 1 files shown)


Arksine:
Thanks, I merged the commit.

The bias flags were not implemented in linux kernel until version 5.5, libgiod subsequently [added support in version 1.5](https://git.kernel.org/pub/scm/libs/libgpiod/libgpiod.git/tree/README).

If you look at Klipper's linux MCU source the bias flags are not set [unless they are defined](https://github.com/Klipper3d/klipper/blob/master/src/linux/gpio.c#L150).  The primary difference is that Klipper will just ignore the pullup setting, whereas Moonraker will raise an exception since it can easily detect this condition in the Python source.

PR #550, opened 2022-11-16
metadata: add support for CrealitySlicer

JoeyGrajciar:
CrealitySlicer is basically same as Cura
Only difference is the identifying regex

Signed-off-by: Jozef Grajciar <jozef.grajciar@gmail.com>


Arksine:
Thanks.  In general I'm ok with adding support for popular slicers, however I'm concerned that this one is violating the GPL given that its a fork of Cura.  I can't find its source and I do not believe it would be appropriate to add support for it if they are in violation.

danorder:
The question id really ask is it really worth risking supporting generally obsolete software. but further is worth risking it for a brand that freely violates open gpl every chance they get. Notably they  still are not listed by klippers partners docs either.  so it would seem they've  still done little.

PR #552, opened 2022-11-21
WLED: Fix for next version and add segment parameter

numanair:
The upcoming 0.14 version of WLED has some API changes that are incompatible with the current WLED_ON command. Removing the "on" call fixes this.

Additionally, I have added support for applying settings (brightness, intensity, speed) to individual WLED Segments. This enables multiple independent lights on one WLED controller. The parameter ID is zero-indexed, and if it is not provided the setting is applied to all segments on the WLED "strip".

I do wonder if ID is the best name for this parameter, as it is not self explanatory imo. It is just what WLED calls it.

I tried to rebase my commits as I initially left off the sign-off message. I am not sure if I need to do another step there as I don't know git very well.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -2677,12 +2677,15 @@ gcode:
   {% set brightness = params.BRIGHTNESS|default(-1)|int %}
   {% set intensity = params.INTENSITY|default(-1)|int %}
   {% set speed = params.SPEED|default(-1)|int %}
+  {% set segment_id = params.SEGMENT_ID|default(-1)|int %}
+  # Optionally specify a WLED segment with the SEGMENT_ID parameter, which starts at 0.
 
   {action_call_remote_method("set_wled_state",
                              strip=strip,
                              brightness=brightness,
                              intensity=intensity,
-                             speed=speed)}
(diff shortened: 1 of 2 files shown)


Arksine:
Thanks.   One thing we would need to be sure of is that these changes don't break compatibility with previous versions of WLED.  

Since this PR includes multiple, unrelated changes to the WLED module its important to break those out into separate commits.  One commit for the compatibility fixes, one commit for the addition of the `id` parameter.

JRHeaton:
I think `id` is too ambiguous a parameter name. Technically, `index` is an ID as well as `preset`, but they're specifically named what type of ID they are. I am more in favor of calling this `segment` to be in line with the others.

Additionally, I'm willing to make that quick parameter name change, fork this, and rebase to the 2 commits so we can get this in. Let me know if it would be permissible to create a new PR that fork to this repo, and credit you you for coding it. Let me know!

If I don't hear back from you, I may fall back to implementing it myself and PRing it, though keep in mind it will be basically what you've coded, as there's not much to change for the `on` flag, and and not really any variation beyond syntactically that a different implementation of setting segment id would have.

PR #554, opened 2022-11-22
machine: use virtualization.virt_identifier property

pedrolamas:
I believe this this get should use `virt_identifier`, not `virt_id`.

Before:

image

After:

image


Arksine:
Thanks.

Issue #434, opened 2022-05-20
The ability to limit the rate MQTT messages.


alexrmay91:
### Is your feature request related to a problem? Please describe

When setting up objects to publish MQTT messages (specifically temperatures), messages are sent extremely rapidly and with a huge amount of decimals. Documentation states that object state is only sent if the value has changed, but with temperatures to so many decimal places, it will always change.

### Describe the solution you'd like

I would like to be able to limit the rate for specific objects. Something like the following:

     status_objects:
       webhooks
       toolhead=position,print_time,homed_axes
       extruder=temperature
         rate: 1000 #ms between mesages

Alternatively, a way to round floats to a smaller number of decimals may limit the rate at which messages get published.

### Describe alternatives you've considered

Using REST would be an alternative because I can specify the rate queries occur, but I prefer push communication to polling communication.

contagon:
I'd like to add a +1 for this feature. After anything more than a couple of minutes publishing, I'm getting thousands of messages which seems unnecessary.

dkneisz:
I'd like to push this topic. I want to include my printer in my openHAB with MQTT. Right now this is sending lots of messages, so it's unusable, because the data will also end up in a database. In Octoprint you have the option to set a threshold for the temperature. It would be nice to have at least that option here to to limit the messages sent. Currently I get around 4 to 5 messages for the various temperatures every second. In that time the temperature has changed by 0.02 °C. That is more or less useless information.

PR #570, opened 2022-12-18
mqtt: add option to publish states to individual mqtt topics

mw-white:
This change allows the user to choose the strategy for publishing klipper states to MQTT.  The original strategy where all state updates are published to a common topic is still the default, but can be turned off using the "publish_combined_status" config option.

The newly added strategy is publishing individual state updates to separate mqtt topics.  It is disabled by default, and can be enabled with the "publish_split_status" config option.

Both strategies can be used together.

Example - when the "heater_bed" attribute "temperature" is published:

publish_combined_status (the existing strategy) publishes a payload similar to:
{"eventtime": timestamp, "status":{ "heater_bed": {"temperature": 55.00}}}
to {instance_name}/klipper/status.  There may be other values in the status key.

publish_split_status would publish a payload similar to:
{"eventtime": timestamp, "value": 55.00}
to {instance_name}/klipper/states/heater_bed/temperature

The newly added strategy is useful for interfacing with other systems as the
information is always available as the last published change.  For example, by using this
strategy, the HomeAssistant example should be able to be redone without needing
to poll the REST API (at least for the sensors), and could determine the printer's
online/offline status properly using the {instance_name}/moonraker/status payload.

I will follow up with an updated HomeAssistant example if this is merged.

Signed-off-by: Matt White <m.w.white@gmail.com>

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1486,8 +1486,7 @@ status_objects:
 #   A newline separated list of Klipper objects whose state will be
 #   published.  There are two different ways to publish the states - you
 #   can use either or both depending on your need.  See the 
-#   "publish_combined_status" and "publish_split_status" options for
-#   details.
+#   "publish_split_status" options for details.
 #
 #   For example, this option could be set as follows:
 #
@@ -1509,12 +1508,14 @@ status_objects:
 #
 #   If not configured then no objects will be tracked and published to
 #   the klipper/status topic.
-publish_combined_status: True
-#   Defaults to True. All Klipper object state updates will be published to
-#   a single mqtt state with the following topic:
-#     {instance_name}/klipper/status
 publish_split_status: False
-#   Defaults to False. All Klipper object state updates will be published to
+#   Configures how to publish status updates to MQTT.  
+#
+#   When set to False (default), all Klipper object state updates will be
+#   published to a single mqtt state with the following topic:
+#     {instance_name}/klipper/status
+#
+#   When set to True, all Klipper object state updates will be published to
 #   separate mqtt topics derived from the object and item in the following
 #   format:
 #     {instance_name}/klipper/state/{objectname}/{statename}
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1484,8 +1484,11 @@ instance_name:
 #   The default is the machine's hostname.
 status_objects:
 #   A newline separated list of Klipper objects whose state will be
-#   published in the payload of the following topic:
-#      {instance_name}/klipper/status
+#   published.  There are two different ways to publish the states - you
+#   can use either or both depending on your need.  See the 
+#   "publish_combined_status" and "publish_split_status" options for
+#   details.
+#
 #   For example, this option could be set as follows:
 #
 #     status_objects:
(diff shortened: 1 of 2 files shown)


pedrolamas:
I haven't tested this yet, but it got my attention as I'm an avid Home Assistant user and wrote part of the MQTT last will and testament implementation here.

I'm not sure what you mean with "information is always available" because currently, these messages are not sent with `retain=True`, so any client that did not receive them, will not be able to get it afterwards.

AFAIK, there is no state maintained in the MQTT server without message retention!

mw-white:
@pedrolamas You're right, I had just been testing with MQTT Explorer but forgot that it wouldn't be retained.  I will update this PR - it's not as useful without retain.

Thanks!

Arksine:
Thanks.  I'm unsure of the additional benefit added with this change.  MQTT clients have access to APIs that query printer state so message retention should not be necessary, nor should it be necessary to poll the REST API.

mw-white:
@Arksine with this option, MQTT clients would be able to passively consume printer status without needing to send a query for a specific piece of information.  For HomeAssistant (my use case) the REST interface would be fine if the printer was on all the time, but when the printer is powered off, obviously the REST queries time out and HA throws errors.  By using MQTT's pub/sub model and checking the Moonraker status topic, we can identify if the printer is online/offline and get the sensor information.  For example:

```
mqtt:
  sensor:
    - name: "V24 Extruder Temperature"
      state_topic: "v24/klipper/state/extruder/temperature"
      value_template: "{{ value_json.value }}"
      unit_of_measurement: '°C'
      availability_topic: "v24/moonraker/status"
      availability_template: "{{ value_json.server }}"
      payload_available: "online"
      payload_not_available: "offline"
      device_class: temperature
```

With the combined status payload, I can use the value_template to pull out individual items, but the issue is that each value is only in the payload when it changes.  HomeAssistant sees the payload update and tries to pull the updated value, but it's not there - it ends up being 'unavailable' even though the printer is up and running.

Thanks!

pedrolamas:
I can relate to the above, it's actually part of the reason why I'm trying to move away from Rest API in Home Assistant to [using MQTT as much as possible](https://github.com/pedrolamas/home-assistant-config/blob/49d0bf3141fd4b2437d25e51d9e658b5b8885bcf/config/integrations/mqtt.yaml#L6-L33), though I have experienced the same issue caused by lack of message retention.

Arksine:
Ok, sounds reasonable.  It seems redundant to have two separate options, I think we could simply have the `publish_split_status` option and publish the selected strategy.

mw-white:
No problem, I can update the PR to just use the one option.   I'm away from home for a couple of days; I will do it when I get back Tuesday.

mw-white:
OK - code (and docs) updated to just use the publish_split_status option to toggle between behaviour.

Quick question @Arksine - would you be OK with a follow up PR that added an option to throttle MQTT updates?  The default 0.25s rate is nice for a responsive GUI, but for interfacing with a home automation platform to show on a dashboard, it's a bit overwhelming.  It would address #434

Arksine:
Thanks.  With regard to your question on "throttling", I'm not sold that it belongs in the server.  The front end should be able to configure the display frequency.

pedrolamas:
I was going to test the PR changes but as it is now merged will just update my instance and check how it behaves directly! 🙂

pedrolamas:
Update: just tested with my printer LED's and after a few minor changes in Home Assistant, it worked very well!

Great job @mw-white! 😁

PR #574, opened 2022-12-22
job_queue: add "reset" parameter to post endpoint

### LINKED ISSUE #382: Allow re-ordering of job queue entries

The current job queue implementation appears to be a static FIFO list.  It would be more useful if items could be re-ordered.  For example, I might have 5 jobs pending in the queue, but realize that the 5th item should be printed before some others.  

Ideally, moonraker would support taking any item in the queue and moving it before any other item in the queue. 

As a simple alternative, just the ability to move any item in the queue to the top of the queue (instead of to a selected position) would still be useful for those "oh, I need to print this next" type situations.

### Additional information

My perception of moonraker is only as seen via mainsail.  I apologize if this makes my descriptions seem odd.

pedrolamas:
Adds a new optional `reset` Boolean parameter to the POST endpoint of the `job_queue` component, allowing to clear the job queue before adding the new filenames.

Resolves #382

### DIFF
--- a/docs/web_api.md
+++ b/docs/web_api.md
@@ -2665,7 +2665,8 @@ Content-Type: application/json
         "job1.gcode",
         "job2.gcode",
         "subdir/job3.gcode"
-    ]
+    ],
+    "reset": false
 }
 ```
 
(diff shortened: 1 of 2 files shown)


pedrolamas:
I've noticed that this is not working consistently (sometimes it does not put files in the same order I sent), so I'm marking as a draft so I can understand what is going on...

pedrolamas:
Seems issue is in the client side, so restoring PR back to "ready for review".

Arksine:
Looks good, thanks.

PR #575, opened 2022-12-25
metadata.py: Pull first layer dimension metadata from cura gcode

JeremyRuhland:
Cura slicer emits metadata about the dimensions of the first layer, this PR makes that data available through the API.

This is one half of a larger project to enable adaptive bed mesh in klipper without a cura post processing script.


JeremyRuhland:
Aaaaaand I just discovered someone implemented this automatic adaptive bed mesh feature using an entirely different method (exclude object) like 2 weeks ago rendering this entire thing moot. I see no real need to merge it into upstream.

PR #590, opened 2023-01-20
Adding OctoEverywhere.com as a default allowed service.

QuinnDamerell:
I'm adding support for Moonraker / Mainsail / Fluidd to OctoEverywhere.com! My plugin is almost ready for early testing; things are looking great.

This simple PR allows the "services" UI to show the OctoEverywhere service.

### DIFF
--- a/moonraker/components/machine.py
+++ b/moonraker/components/machine.py
@@ -62,7 +62,8 @@ DEFAULT_ALLOWED_SERVICES = [
     "moonraker-telegram-bot",
     "moonraker-obico",
     "sonar",
-    "crowsnest"
+    "crowsnest",
+    "octoeverywhere"
 ]
 CGROUP_PATH = "/proc/1/cgroup"
 SCHED_PATH = "/proc/1/sched"


Arksine:
Thanks.  The change looks good, however the commits [need a signature](https://moonraker.readthedocs.io/en/latest/contributing/#git-commit-format).  You can squash and reword the commits yourself or add it to the PR and I'll take care of it.

One thing you should keep in mind is that this change will only affect installations that do not currently have `moonraker.services` in the data path (ie: typically fresh Moonraker installations).  You will probably want your installer to check to see if this file exists, and if so check to see if `octoeverywhere` is in it.  Alternatively you can direct your users to add it manually via ssh.

QuinnDamerell:
@Arksine sorry for the delay, I have been heads down working on the plugin! Thanks for the note about existing installs; like you suggested, I already have logic in the plugin that will add it. 

I have updated the commit with the new signature; let me know if anything else needs to be changed!

Arksine:
Looks good, thanks.

QuinnDamerell:
Thank you!

PR #592, opened 2023-01-20
Add webcam_rotation parameter to octoprint_compat

zanhecht:
Adds a new `webcam_rotation` parameter to the `[octoprint_compat]` settings of the config file which allows setting 0, 90, 180, or 270 degree rotations, instead of separate `flip_h`, `flip_v`, and `rotate_90` paramters, to match main Moonraker webcam settings.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1928,18 +1928,18 @@ separate from `moonraker.conf`.  This allows users to safely distribute
 their configuration and log files without revealing credentials and
 other sensitive information.
 
-!!! Note:
-    This section no long has configuration options.  Previously the
+!!! Note
+    This section no longer has configuration options.  Previously the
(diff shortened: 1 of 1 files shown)


Arksine:
Thanks.  I don't believe your conversion between Moonraker's rotation option and the "flip" options are correct, as the flip options effectively rotate the image about different axes.  With that being said it probably doesn't make sense to introduce an additional rotation option into the `[octoprint_compat]` section.  It might make sense to allow `[octoprint_compat]` to use a configured `[webcam]`, however at the moment I don't think its worth the trouble.  My hope is that eventually applications will implement Moonraker's APIs so there is no need to configure `[octoprint_compat]`.

Also, for future reference all commits should follow the [contirbuting guidelines](https://moonraker.readthedocs.io/en/latest/contributing/#git-commit-format).

PR #596, opened 2023-01-24
Added ability to control hue device groups in power.py

cmerkle:
Added the ability to control entire Hue device power groups in addition to controlling Hue devices individually. This can be advantageous if you have several Hue lights in the area of the printer that you would like to control all together.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1152,6 +1152,11 @@ device_id:
 #   The device id of the light/socket you want to control.
 #   An explanation on how you could get the device id, can be found here:
 #   https://developers.meethue.com/develop/get-started-2/#turning-a-light-on-and-off
+device_type: light
+#   Set to light to control a single hue light, or group to control a hue light gorup. 
+#   If device_type is set to light, the device_id should be the light id, 
+#   and if the device_type is group, the device_id should be the group id.
+#   The default is "light.
 
 ```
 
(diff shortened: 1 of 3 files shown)


Arksine:
Thanks.  I'm ok with adding support for "Hue Groups", however this implementation looks like its nearly a direct copy and past of the original hue implementation.  Perhaps rather than creating a new "type" we could add an `is_hue_group` boolean option to the original implementation, then set the url and parse the response based on this option.

Additionally the documentation will need to be updated to reflect changes and additions.

cmerkle:
> Thanks.  I'm ok with adding support for "Hue Groups", however this implementation looks like its nearly a direct copy  of the original hue implementation.  Perhaps rather than creating a new "type" we could add an `is_hue_group` boolean option to the original implementation, then set the url and parse the response based on this option.
> 
> 
> 
> Additionally the documentation will need to be updated to reflect changes and additions.

Yea I was thinking about maybe just adding a config option for the existing hue device called "type" which can be either "light" or "group". I can try to implement that when I get a chance. Am I responsible for updating the docs? Sorry, I'm new to all this, this is my first PR.

Arksine:
Looks good.  The `device_type` needs to be documented, then this will be ready to merge.

cmerkle:
@Arksine Docs have been updated!

PR #597, opened 2023-01-27
Filament manager refactor

jmccaull:
Module Name: Spool Manager

Component to manager spools and track usage
This is forked from https://github.com/Arksine/moonraker/pull/123/ and attempting to address the comment: #https://github.com/Arksine/moonraker/pull/123#issuecomment-808113520

I have not testing anything but figured getting feedback earlier would be good. I may still make some adjustments, like removing 187/188 disabling the spool when too much is used. In its current state this would just track the usage and let whatever UI handle making sure there is enough for the next print, etc. I have also yet to get to the last bullet point regarding configuration

Signed-off-by: John Mccaulley jmccaull@gmu.edu

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -482,7 +482,34 @@ Enables print history tracking.
 [history]
 ```
 
-### `[paneldue]`
+## `[spool_manager]`
+Enables spool manager. Spool manager templates are optional predefined sets of
+values that can be referenced with the template property when creating a new spool.
+The spool manager publishes four endpoints: "/spool_manager/spool and
+"/spool_manager/spool/list supporting GET, POST and DELETE, "/spool_manager/spool/active"
+supporting GET and POST, and "/spool_manager/templates" supporting GET. The active
+spool is the one to which usage will be tracked. Unused/spent spools can be marked hidden,
+and can optionally be omitted from the list response.
+
+```ini
+# moonraker.conf
+
+[spool_manager]
+sync_rate_seconds: 900
+#   The time in seconds between saving filament usage. Updates will happen on the next
+#   status update from klipper. Default is 900.
+
+[spool_manager template pla]
+vendor: vendor
+material: PLA
+density:  1.24
+diameter: 1.75
+filament_weight: 1000
+spool_weight: 200
+cost: 24.99
+```
+
+## `[paneldue]`
 Enables PanelDue display support.  The PanelDue should be connected to the
 host machine, either via the machine's UART GPIOs or through a USB-TTL
 converter.  Currently PanelDue Firmware Version 1.24 is supported.  Other
@@ -1594,7 +1621,7 @@ instance_name:
 status_objects:
 #   A newline separated list of Klipper objects whose state will be
 #   published.  There are two different ways to publish the states - you
-#   can use either or both depending on your need.  See the 
+#   can use either or both depending on your need.  See the
 #   "publish_split_status" options for details.
 #
 #   For example, this option could be set as follows:
@@ -1618,7 +1645,7 @@ status_objects:
 #   If not configured then no objects will be tracked and published to
 #   the klipper/status topic.
 publish_split_status: False
-#   Configures how to publish status updates to MQTT.  
+#   Configures how to publish status updates to MQTT.
 #
 #   When set to False (default), all Klipper object state updates will be
 #   published to a single mqtt state with the following topic:
--- a/moonraker/components/spool_manager.py
+++ b/moonraker/components/spool_manager.py
@@ -1,0 +1,403 @@
+# Filament Manager for printer
+#
+# Copyright (C) 2021 Mateusz Brzezinski <mateusz.brzezinski@gmail.com>
+#
+# This file may be distributed under the terms of the GNU GPLv3 license.
+from __future__ import annotations
+import asyncio
+import datetime
+import logging
+import time
+import math
+from typing import TYPE_CHECKING, Dict, Any, List
+
+if TYPE_CHECKING:
+    from typing import Set, Optional
+    from database import NamespaceWrapper
+    from moonraker.websockets import WebRequest
+    from .klippy_apis import KlippyAPI as APIComp
+    from confighelper import ConfigHelper
+
+SPOOL_NAMESPACE = "spool_manager"
+MOONRAKER_NAMESPACE = "moonraker"
+ACTIVE_SPOOL_KEY = "spool_manager.active_spool_id"
+MAX_SPOOLS = 1000
+
+
+class Validation:
+    def validate(self) -> Set[str]:
+        failed = filter(lambda f: self.__getattribute__(f) is None,
+                        self._required_attributes)
+        return set(failed)
+
+    _required_attributes: Set[str] = set()
+
+
+class Spool(Validation):
+    _required_attributes: Set[str] = {"diameter", "filament_weight",
+                                      'material'}
+
+    def __init__(self, data={}):
+        self.name: str = None
+        self.hidden: bool = False
+        self.color: str = None
+        self.vendor: str = None
+        self.material: str = None
+        self.density: float = None
+        self.diameter: float = None
+        self.filament_weight: float = None
+        self.used_length: float = 0
+        self.spool_weight: float = None
+        self.first_used: float = None
+        self.last_used: float = None
+        self.cost: float = None
+        self.comment: str = None
+
+        self.update(data)
+
+    def update(self, data):
+        for a in data:
+            if hasattr(self, a):
+                setattr(self, a, data[a])
+
+    def used_weight(self) -> float:
+        used_weight = 0.0
+        if self.diameter and self.density:
+            r = self.diameter / 2
+            density_mm = self.density / 1000
+            used_weight = math.pi * r * r * self.used_length * density_mm
+        return used_weight
+
+    def serialize(self, include_calculated: bool = False):
+        data = self.__dict__.copy()
+        if include_calculated:
+            data.update({'used_weight': self.used_weight()})
+        return data
+
+
+class SpoolManager:
+    def __init__(self, config: ConfigHelper):
+        self.server = config.get_server()
+
+        self.templates = self._parse_materials_cfg(config)
+        self.sync_rate_seconds = config.getint(option="sync_rate_seconds",
+                                               default=900)
+        self.extruded_lock = asyncio.Lock()
+
+        database = self.server.lookup_component("database")
+        database.register_local_namespace(SPOOL_NAMESPACE)
+        self.db: NamespaceWrapper = database.wrap_namespace(SPOOL_NAMESPACE,
+                                                            parse_keys=False)
+        self.moonraker_db: NamespaceWrapper = database.wrap_namespace(
+            MOONRAKER_NAMESPACE, parse_keys=False)
+
+        self.handler = SpoolManagerHandler(self.server, self)
+
+    async def on_exit(self) -> None:
+        await self.track_filament_usage()
+
+    def _parse_materials_cfg(self, config: ConfigHelper) -> \
+            Dict[str, Dict[str, Any]]:
+        template_names = config.get_prefix_sections('spool_manager template')
+        logging.debug("template names: %s", template_names)
+        templates: Dict[str, Dict[str, Any]] = {}
+        for template_path in template_names:
+            config_helper = config[template_path]
+            template: Dict[str, Any] = {}
+
+            vendor = config_helper.get('vendor', None)
+            if vendor:
+                template['vendor'] = vendor
+
+            material = config_helper.get('material', None)
+            if material:
+                template['material'] = material
+
+            density = config_helper.get('density', None)
+            if density:
+                template['density'] = float(density)
+
+            diameter = config_helper.get('diameter', None)
+            if diameter:
+                template['diameter'] = float(diameter)
+
+            filament_weight = config_helper.get('filament_weight', None)
+            if filament_weight:
+                template['filament_weight'] = float(filament_weight)
+
+            spool_weight = config_helper.get('spool_weight', None)
+            if spool_weight:
+                template['spool_weight'] = float(spool_weight)
+
+            cost = config_helper.get('cost', None)
+            if cost:
+                template['cost'] = float(cost)
+
+            template_name = str(template_path).upper() \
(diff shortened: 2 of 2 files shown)


jmccaull:
I've reworked the config section to match previous review comments. Instead of linking spools to a template I think it makes more sense to save the template values with the spool. This allows updating the template without updating all spools created with it or removing the template completely. To make updating multiple easier, I'm adding support to the list path for post/delete.

jmccaull:
Module Name: Spool Manager

Component to manage spools and track usage
This is forked from https://github.com/Arksine/moonraker/pull/123 and attempting to address the outstanding comments

Allows configuring templates to add spools:
[spool_manager template pla]
vendor: vendor
material: PLA
density:  1.24
diameter: 1.75
filament_weight: 1000
spool_weight: 200
cost: 24.99

This exposes 4 endpoints: "/spool_manager/spool and
"/spool_manager/spool/list supporting GET, POST and DELETE, "/spool_manager/spool/active"
supporting GET and POST, and "/spool_manager/templates" supporting GET.

example curl for each operation:

curl --location --request POST 'http://localhost:7125/spool_manager/spool' \
--header 'Content-Type: application/json' \
--data-raw '{"name": "templateTest", "template": "PLA"}'

curl --location --request GET 'http://localhost:7125/spool_manager/spool?id=000001'

curl --location --request DELETE 'http://localhost:7125/spool_manager/spool?id=000001'

curl --location --request GET 'http://localhost:7125/spool_manager/spool/active'

curl --location --request POST 'http://localhost:7125/spool_manager/spool/active' \
--header 'Content-Type: application/json' \
--data-raw '{"id": "000001"}'

curl --location --request POST 'http://localhost:7125/spool_manager/spool/list' \
--header 'Content-Type: application/json' \
--data-raw '{"spools" : [{"name": "templateTest", "template": "PLA"}, {"name": "templateTest", "template": "PLA"}]}'

curl --location --request GET 'http://localhost:7125/spool_manager/spool/list'

curl --location --request DELETE 'http://localhost:7125/spool_manager/spool/list' \
--header 'Content-Type: application/json' \
--data-raw '{"ids": ["000003", "000005"]}'

curl --location --request GET 'http://10.0.0.248:7125/spool_manager/templates'

Signed-off-by: John Mccaulley jmccaull@gmu.edu

jmccaull:
After using this for a little it seems like on exit method might not be getting called, testing it out now. I'm also considering adding a parameter for a sync period, if a status update gets processed and the time limit is exceeded it will write the current usage or something

hvraven:
Hi @jmccaull,
thanks for working on this again. I just skimmed through the code to understand how it works and (with a lot of Python experience, but without any experience on the moonraker codebase) it looks good. I will try to give it a test run as soon as possible.

Only thing I am wondering whether it makes sense to add an extra endpoint to  update the amount of used filament. Basically everything what `track_filament_usage` is doing once it got the extruded length. That would open up a simple path to multiple instances sharing the spool manager, where one master has the actual spool data and the others only track their own active spool & usage, passing everything else to the endpoint of the configured master. It would also work without the extra endpoint, but it would eliminate possible race conditions when multiple ones are updating the same spool in parallel (yes, in theory they shouldn't).

Arksine:
Time permitting I intend to review this the next few days.  FWIW the primary reason I haven't implemented this myself was brought up by @hvraven.  I think most people that have a use for this functionality share spools across multiple printers.  In that case a `spool manager` would be better implemented as an external agent that provides centralized management across multiple instances of Moonraker.  That said, I'm not opposed to merging a module such as this, however users will need to be aware that it is intended to manage spools for a single printer.

> Only thing I am wondering whether it makes sense to add an extra endpoint to update the amount of used filament.

 As mentioned above, if the goal is to manage spools across multiple printers then I don't think its a good candidate for a Moonraker component.  I have had previous discussions with others about sharing data across multiple Moonraker instances, including setting a Moonraker "master" instance.  The result of that conversation was that the challenges and complexity involved were not worth it.  The best way to do this would be to include a shared database option that can be used in place of (or perhaps in addition to) LMDB.  Its something I plan to look into in the future, but it is currently low on my list of priorities.

jmccaull:
@Arksine I honestly see two different components completely as the use cases are pretty distinct. I think it makes the most sense to build a dedicated filament server and have a component that communicates with it. In that scenario the only things in common here are subscribing to the EPOS and calculating used filament. You probably wouldn't need all the template configuration or endpoints related to spools. It would probably want to have logic to checkout and return spools, etc.

jmccaull:
Maybe if people really want the server I could see doing a server component and a client component. In that case this would have a bit more in common with the server but to me this seems like it might be beyond the scope of Moonraker?

hvraven:
This code already implements two parts, they are just combined. One part is tracking the filament usage and storing an id of the currently active spool. The other part is all the information about that spool (and other available spools). The first part needs to be in moonraker, the second part could be implemented anywhere. A simple entry in the config could be used to point the filament tracking code to the endpoint to send the updates to.

That's why I suggested to have a dedicated endpoint for this task to allow for atomic updates of the filament usage. Then sharing the database between multiple instances would not be a problem. Moonraker updates the usage from time to time by calling the endpoint, tying everything together must be done in the user interface.

Then the question is whether moonraker should provide this database as well or not. I personally think there would be benefit to have a (at least basic) implementation in moonraker to provide a fully working  system. Then one can still provide an alternate spool managing service with possibly more features. But ultimately that's a question of project scope  and call of to the project maintainers.

Arksine:
> Maybe if people really want the server in Moonraker I could see doing a server component and a client component. In that case this would have a bit more in common with the server but to me this seems like it might be beyond the scope of Moonraker?

I think it is out of the scope of Moonraker.  I wouldn't want to maintain a "filament spool server" in this repo.

> I honestly see two different components completely as the use cases are pretty distinct. I think it makes the most sense to build a dedicated filament server and have a component that communicates with it. In that scenario the only things in common here are subscribing to the EPOS and calculating used filament. You probably wouldn't need all the template configuration or endpoints related to spools. It would probably want to have logic to checkout and return spools, etc.

A filament service doesn't necessarily need to have a component counterpart.   It could operate similar to Obico, where it connects to Moonraker's websocket and identifies itself as an `agent`.  This provides a few more endpoints that allow it to register methods that front ends may call and send notifications that front ends receive.  The server would have access to Moonraker's full JSON-RPC API.  All of the subscription and tracking logic would remain in the server.

> This code already implements two parts, they are just combined. One part is tracking the filament usage and storing an id of the currently active spool. The other part is all the information about that spool (and other available spools). The first part needs to be in moonraker, the second part could be implemented anywhere. A simple entry in the config could be used to point the filament tracking code to the endpoint to send the updates to.

Implementation isn't quite so simple.  If the master goes down or otherwise becomes unavailable we have a problem.    Its a challenge that can be solved, but one that is better suited for a dedicated service.  Adding a basic solution for spool sharing will almost certainly result in tracking errors, which users will submit as bug reports.

jmccaull:
That makes sense. I think @hvraven 's point may have been that the code is split into two classes. I would describe it as one class that interacts with the database and another that interacts with Moonraker. I've mostly kept it the way @mateuszbrzezinski had it but had a I started from scratch I might have reversed the way the classes interact as the handler taking in the manager is somewhat a circular dependency, making breaking it up more difficult. I can try refactoring that but that is going to be a much longer effort

jmccaull:
on exit and periodic sync seem to be working well now

jmccaull:
The only issue I'm having at this point is I can't think of a way to get this working with a klipper menu. I could add something like a remote method call to allow changing the active filament from a menu, but I'm not sure I can get the filament ids loaded into a menu. Anyone see a path forward with jinja or something?

hvraven:
@Arksine I took a look into the `agent` interface you mentioned and thought about how to implement such a server.  The flow I think of is roughly:
* Moonraker tracks filament usage plus an identifier which spool is active (just an id without any meaning to moonraker)
* Usage is reported periodically / on certain events (start, stop, pause, active spool change)
* All tracking is done in the spool manager
* Control of the spool manager (checking and managing spools) either directly through a custom UI (directly calling the spool manager API) or indirectly through moonraker using the extension API. 

This will require only minimal changes to moonraker to support the tracking of the filament usage (or is this information already available through some of the events). I am also not sure how to best report the usage. For moonraker implementation it is probably easiest to emit a new event with the information. For the server implementation a API call made by moonraker would be better.

In general the biggest problem I see at the moment is the required construction of a websocket connection from the spool manager to moonraker. As by design there is only a single spool manager, but multiple moonraker instances it would be better if they would connect to the spool manager. That's why I think the API call to the server would be better, as it does not require a persistent connection and allows for completely stateless tracking of the filament usage. In case the call fails there is nothing to do for moonraker (except logging the error, possibly notifying the user somehow that it failed).
The agent connection to moonraker could still be performed, but would be entirely optional and only needed by the UIs to list/manage the spools through moonraker.

Arksine:
> I took a look into the agent interface you mentioned and thought about how to implement such a server. The flow I think of is roughly:

The agent would handle spool tracking, it can subscribe to the same data Moonraker can.  By centralizing in this manner you will never end up with a situation where two instances are erroneously tracking filament on the same spool.  The entire process can be implemented with any changes to Moonraker.

> In general the biggest problem I see at the moment is the required construction of a websocket connection from the spool manager to moonraker. As by design there is only a single spool manager, but multiple moonraker instances it would be better if they would connect to the spool manager.

I'm not sure why this is a problem.   The persistent connection allows front ends to know with certainty that the spool manager is running and connected.  If it goes down the UI can alert the user as it won't show up in Moonraker's list of identified agents.

At the moment I want to avoid Moonraker communicating with other instances of Moonraker.  That involves maintaining both the client and server logic in the component, IMO its scope creep.  That said, I plan on introducing 3rd party extensions this year.  If the agent idea isn't appealing then this alternative could be implemented and maintained outside of Moonraker.

hvraven:
> The agent would handle spool tracking, it can subscribe to the same data Moonraker can. By centralizing in this manner you will never end up with a situation where two instances are erroneously tracking filament on the same spool. The entire process can be implemented with any changes to Moonraker.
Listening to all events from klipper appeared to a bit excessive to me, but I'll give it a shot. That certainly would be the easiest solution. 
> I'm not sure why this is a problem. The persistent connection allows front ends to know with certainty that the spool manager is running and connected. If it goes down the UI can alert the user as it won't show up in Moonraker's list of identified agents.
I am primarily concerned about start up. Is there a mechanism to have moonraker inform agents that it just has started and they can connect now? Otherwise the spool manager would have to poll all configured clients regularly to check if any of them became available.

> At the moment I want to avoid Moonraker communicating with other instances of Moonraker. That involves maintaining both the client and server logic in the component, IMO its scope creep. That said, I plan on introducing 3rd party extensions this year. If the agent idea isn't appealing then this alternative could be implemented and maintained outside of Moonraker.
I never intended this kind of communication. It would only between moonraker and the spool manager. 3rd party extensions might be interesting if tracking inside the manager does not work reliably.

Arksine:
> Listening to all events from klipper appeared to a bit excessive to me, but I'll give it a shot. That certainly would be the easiest solution.

You don't need them all.  Moonraker sends notifications for `klippy_ready`, `klippy_shutdown`, and `klippy_disconected` which you would likely need.   From Klipper you would probably just need the toolhead position and possibly the `print_stats` data if you want to be informed when a print starts/stops/etc.  All that said, presuming we get this PR merged you could take advantage of its tracking.  Downside is that it requires configuration in each instance of Moonraker, where as handling tracking in the agent requires none.

> I am primarily concerned about start up. Is there a mechanism to have moonraker inform agents that it just has started and they can connect now? Otherwise the spool manager would have to poll all configured clients regularly to check if any of them became available.

Correct, a simple reconnecting websocket would be required.  Presumably Obico and KlipperScreen do this and it is effectively how Moonraker connects to Klipper's Unix Domain Socket.    An alternative is that I could implement support for outbound connections where Moonraker is the websocket client.   As above, this would require some configuration in each instance of Moonraker.

With regard to this PR, I'm currently in the process of reviewing it are some issues to work through and  discuss.  I think much of it may be some leftover issues from the previous PR, such as missing annotations.

Donkie:
I've only used the Klipper/Moonraker setup for a short time and I was stumped to see that there isn't much support for keeping track of filaments. I found this and the parent PR which both look promising, but as discussed here I think it's a bit weird that this isn't being developed as a completely separate web app which Moonraker and its clients could hook into. 

Throwing together a simple project with an internal database and a REST API that just keeps track of materials and spools isn't so hard. Why isn't that approach being considered instead? Moonraker could then receive an update which hooks into this external service's API and sends filament usage updates etc.

This would both solve the problem of allowing multiple printers, and hell even completely different ecosystems, to use the same database. Additionally, the scope issues that are discussed here which I agree with would also be solved.

If no one else would be interested I could take charge in creating an external service like this, if it would be of any value.

jmccaull:
@Donkie please feel free, I don't have a driving need for anything that works with more than one printer. The only reason it would need to be a component is for the database.
@Arksine I should be able to spend a bit more time on this soon. Honestly I've been teaching myself python while working on this so any direction would be appreciated.

Donkie:
I've setup a basic proof-of-concept project now that tries to fulfill the needs: https://github.com/Donkie/Spoolman 

Please see if this is what you're after and if this is something I should continue working on :)

hvraven:
Hi @Donkie ,

thanks for starting a dedicated software :+1: (also thumbs up for the software stack). I had a look at the API and it looks sensible to me. Only comment I have there is that I think the use spool filament call should always update the spool, even if it is already empty. In my experience the amount tracking and initial weight of the spool is not that precise.

Now only some method is missing to connect to various klipper instances, allowing to select the spool currently in use and perform automatic updates while printing.

Donkie:
@hvraven thank you for your feedback! Sounds like we should be counting upwards (filament usage) then instead of counting downwards (remaining weight), seems sensible?

Perhaps we can continue discussions about the SW itself in its own github to prevent polluting this discussion? :)

Also what does @Arksine think about this direction?

Donkie:
Moonraker now supports interaction with Spoolman which fulfills the same goals, so I believe this PR could be closed.

https://moonraker.readthedocs.io/en/latest/configuration/#spoolman

jmccaull:
Thanks for the heads up, I was already testing out spoolman on my local but hadn't gotten around to closing this. Nice job by the way!

PR #603, opened 2023-02-01
use systemctl poweroff instead of halt

darrellenns:
"systemctl halt" doesn't shut down power (it halts but remains powered on). "systemctl poweroff" is better suited, as it actually powers the machine off.


Arksine:
Thanks.  I ended up having to make this an option in commit 8010f1dda2dca86d2e31a81878876981d3f9a7db.  Some distros don't support the `poweroff` action and thus `systemctl poweroff` will do nothing.  That said, the default action is to power off, so it effectively accomplishes the goal of this pull request.

PR #617, opened 2023-02-19
metadata: Add support for Simplify3D v5

### LINKED ISSUE #571: Simplify3D V5 Support

Metadata isnt read or just not showed by mainsail 

Add Metadata support

_No response_

### Additional information

here is an example gcode: 
[Support_test.zip](https://github.com/Arksine/moonraker/files/10255003/Support_test.zip)

If you need more infos just ask :)

kleinweby:
This PR adds changed to support the new/changed Metadata for Simplify3D Version five.

For most part this are simple changes where only the "property name" changed slightly.

A bot more involved is the new temperature format, as this is a block of properties that which are grouped together and must be parsed as group.

Closes #571.

### DIFF
--- a/moonraker/components/file_manager/metadata.py
+++ b/moonraker/components/file_manager/metadata.py
@@ -17,6 +17,7 @@ import tempfile
 import zipfile
 import shutil
 import uuid
+import io
 from PIL import Image
 
 # Annotation imports
@@ -571,6 +572,8 @@ class Simplify3D(BaseSlicer):
     def check_identity(self, data: str) -> Optional[Dict[str, str]]:
         match = re.search(r"Simplify3D\(R\)\sVersion\s(.*)", data)
         if match:
+            self._version = match.group(1)
+            self._is_v5 = self._version.startswith("5")
             return {
                 'slicer': "Simplify3D",
                 'slicer_version': match.group(1)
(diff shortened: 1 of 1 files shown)


Arksine:
Could you attach a sample gcode file sliced with Simplify3D V5 so I could take a look at it?  Thanks.

moonraker/components/file_manager/metadata.py
```diff
@@ -68,12 +69,15 @@ def _regex_find_ints(pattern: str, data: str) -> List[int]:
             pass
     return []
 
-def _regex_find_first(pattern: str, data: str) -> Optional[float]:
+def _regex_find_first(pattern: str,
+                      data: str,
+                      group: int = 1
+                      ) -> Optional[float]:
```
Arksine, line 75: I don't think the addition of a `group` here is necessary, after all this helper is intended to find the first captured group and convert it to a float.  In places where you need to group an expression you should be able use non-capturing groups.  For example:
```
(?:Plastic\sweight|Material\sWeight)
```

kleinweby: You're right. I've completely forgot that non-capturing groups exist. Thanks, I've changed the code to use them.

kleinweby:
Here is a sample gcode from Simplify3D v5:

[topRegister.gcode.zip](https://github.com/Arksine/moonraker/files/10787379/topRegister.gcode.zip)

There is also one in #571 which works as well with these changes.

Arksine:
Thanks.  I went ahead and merged this, however I may look to simplify the temperature search code.  It appears to me that the following block is of data is consistent between the two files:
```
;   temperatureController,Primary Extruder
;   enableTemperatureController,1
;   temperatureNumber,0
;   temperatureType,extruder
;   stabilizeAtStartup,1
;   temperatureSetpoints,1|210
;   useIdleCooldown,0
;   cooldownTemperature,150
;   cooldownRequiredIdleTime,180
;   cooldownReheatTime,60
```

Presuming the order doesn't change, we can use a single search with the `re.MULTILINE | re.DOTALL` flags set to find the set point.

PR #650, opened 2023-05-05
notifier: Add `body_format` as an option

cwgthornton:
This lets you format the messages through apprise a little nicer


Arksine:
Looks good, thanks.

PR #656, opened 2023-05-21
data_store: check for numeric temperature

pedrolamas:
This is a follow up on https://github.com/Klipper3d/klipper/pull/6210

Right now, Moonraker will throw an error if/when it receives a null temperature (due to the forced `round` operation)

All this PR does right now is check for a non-None value before doing the `round` so it does not throw.


pedrolamas:
Closing this down as it is not necessary anymore given the latest changes!

PR #670, opened 2023-06-07
Rework load_system_module

### LINKED ISSUE #669: load_system_module constantly fails on system with non-standard path layout
### What happened

I use NixOS. It stores Python modules under a unique path starting with `/nix/store/`.
When moonraker wants to load a system module, in particular for GPIO pins (AFAICT it is not used for anything else), it calls [`load_system_module` function](https://github.com/Arksine/moonraker/blob/master/moonraker/utils/__init__.py#L202). This function runs a loop over `SYS_MOD_PATHS` declared as `glob.glob("/usr/lib/python3*/dist-packages") + glob.glob("/usr/lib/python3*/site-packages")` — i.e. looks for packages installed for any Python version.
I think this complex logic (instead of just `import`) is a workaround to look for other Python version's module variant.
But on NixOS there is no `/usr/lib/python3*` and even no `/usr/lib` directory. So `SYS_MOD_PATHS == []` and the loop does not run at all.
As a result, even when `libgpiod` is properly installed and can be just `import`ed, moonraker complains that it was `Unable to import module gpiod` and hence cannot use GPIO pins.

### Client

Fluidd

### Browser

Chrome

### How to reproduce

1. Install Moonraker on a system with non-standard filesystem layout
2. Make sure that `gpiod` can be imported from plain Python shell
3. Enable any GPIO pin in Moonraker config
4. Observe Moonraker unable to import `gpiod`.

### Additional information

Possible fixes:
1. Append some dummy path like `/dev/null` to the end of `SYS_MOD_PATHS` list;
2. First try to just `import` the module, and run all that machinery only if simple `import` fails.

MarSoft:
This fixes #669.

New logic:
First try to import the module directly which will likely succeed.
If it doesn't then try workarounds.
Using `try`…`finally` makes sure that the inserted entry will be removed from `sys.path` even if some unexpected (not `ImportError`) exception happens.

### DIFF
--- a/moonraker/utils/__init__.py
+++ b/moonraker/utils/__init__.py
@@ -200,20 +200,20 @@ def verify_source(
     return checksum, checksum == orig_chksum
 
 def load_system_module(name: str) -> ModuleType:
+    try:
+        return importlib.import_module(name)
+    except ImportError:
+        pass
     for module_path in SYS_MOD_PATHS:
         sys.path.insert(0, module_path)
         try:
-            module = importlib.import_module(name)
+            return importlib.import_module(name)
         except ImportError as e:
             if not isinstance(e, ModuleNotFoundError):
                 logging.exception(f"Failed to load {name} module")
+        finally:
             sys.path.pop(0)
-        else:
-            sys.path.pop(0)
-            break
-    else:
-        raise ServerError(f"Unable to import module {name}")
-    return module
+    raise ServerError(f"Unable to import module {name}")
 
 def get_unix_peer_credentials(
     writer: asyncio.StreamWriter, name: str


Arksine:
Thanks.  Moonraker is generally intended to run in an isolated virtualenv, thus the initial import will fail on most systems.  That is why this method exists, the distribution folder needs to be prefixed to the path for the import to succeed.

The bigger issue is that this code could introduce a regression on some systems.  There is a PIP package named "gpiod", which is different from the package delivered with libgpiod.  The proposed implementation would likely prefer it if a user has it installed,   thus the wrong package would be imported.

I agree with the `finally` addition, however removing the break will definitely introduce a regression as the load will always fail.

I went ahead and pushed a corrected version in commit 1a5cf565328395e05f8879808d12861c94a1172b.  It should perform a direct import without introducing the regressions outlined above.

MarSoft:
Removing `break` would not break anything, because I've also replaced `module = importlib.import_module(name)` with `return importlib.import_module(name)` — i.e. as soon as the module is successfully imported (without an exception thrown), we return it from the function and the loop breaks itself. At the same time, due to how `try...finally` works, the `finally` clause is executed right before we return from the function, making sure that `sys.path` is cleaned up.

OTOH, this would not convert `ImportError` into `ServerError`.

Anyway, I'm happy with any working solution. Thank you for quick reply!

PR #675, opened 2023-06-27
data_store: make additional fields configurable

sddev0:
Prior to this PR, the temperature_store records these sensor fields:
- temperature
- target
- power
- speed

There is a need to create time series records for new combined sensors like VOC-, particle- or humidity sensors with their special measurements.
This PR introduces a new configuration "temperature_store_additional_fields" which extends the existing default fields. It is not possible to replace or reduce that default by configuration, so there is no break in compatability.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -239,6 +239,13 @@ temperature_store_size: 1200
 #   that this value also applies to the "target", "power", and "fan_speed"
 #   if the sensor reports them.  The default is 1200, which is enough to
 #   store approximately 20 minutes of data at one value per second.
+temperature_store_additional_fields:
+    voc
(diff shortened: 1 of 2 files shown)


sddev0:
I have to rethink the architecture. It works for a custom sensor which is registered in the „heaters“ object. But bme280 sensors behave differently: They are represented by one temperature_sensor in the „heaters“ node which onl has temp and another „bme280_sensor“ which exists in the root. With my approach is it not possible to add the humidity and pressure values, as they don’t exist in the temperature_sensor.
I see two alternatives: let the temperature store add non-heater sensors or add another „store“ for non-heater sensors. What do you think?

PR #688, opened 2023-07-19
[Bug] Fix for TPLinkSmart Strip Index Changing.

mista-funky:
There is a feature in the Kasa app that will allow you to change plug index (basically for naming). However, in the power.py there are 2 different methods for retreiving children devices (plugs). If you have the index reordered in Kasa then you will get inconsistent results since those methods for retrieval are not the same. This PR unifies them properly and now they work as intended. Simple change.

Signed-off-by: Rick Marino <rcmarino@gmail.com>

### DIFF
--- a/moonraker/components/power.py
+++ b/moonraker/components/power.py
@@ -809,10 +809,9 @@ class TPLinkSmartPlug(PowerDevice):
             # TPLink device controls multiple devices
             if self.output_id is not None:
                 sysinfo = await self._send_tplink_command("info")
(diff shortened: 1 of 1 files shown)


mista-funky:
The more correct way to pull the children index is used in async def _send_info_request(self). However, as you can see if you don't unify the way we're getting / setting the children plugs there will be issues with properly powering those devices.

Arksine:
Thanks.  The change looks fine to me, however the commit needs to be signed off.  Please refer to the [contributing documentation](https://moonraker.readthedocs.io/en/latest/contributing/) for details.

Issue #689, opened 2023-07-21
Make unix socket path configurable


mapero:
### Is your feature request related to a problem? Please describe

Hi,

thanks for implementing moonraker.

I am currently working on a snap integration to make this run together with klippy in an ubuntu core environment.
However I stumpled over an issue with regard to the unix socket that moonraker is creating in the data directory.

Since the path length for unix socket in linux is limited to 108 chars, and the snap environment is very strict where you can store things. I run into this issue.

```
2023-07-21T18:41:14Z klipper.moonraker[34089]: [eventloop.py:_wrapper()] - Error Running Callback
2023-07-21T18:41:14Z klipper.moonraker[34089]: Traceback (most recent call last):
2023-07-21T18:41:14Z klipper.moonraker[34089]:   File "/snap/klipper/x19/moonraker/eventloop.py", line 73, in _wrapper
2023-07-21T18:41:14Z klipper.moonraker[34089]:     await ret
2023-07-21T18:41:14Z klipper.moonraker[34089]:   File "/snap/klipper/x19/moonraker/server.py", line 183, in server_init
2023-07-21T18:41:14Z klipper.moonraker[34089]:     await self.start_server()
2023-07-21T18:41:14Z klipper.moonraker[34089]:   File "/snap/klipper/x19/moonraker/server.py", line 188, in start_server
2023-07-21T18:41:14Z klipper.moonraker[34089]:     await extm.start_unix_server()
2023-07-21T18:41:14Z klipper.moonraker[34089]:   File "/snap/klipper/x19/moonraker/components/extensions.py", line 122, in start_unix_server
2023-07-21T18:41:14Z klipper.moonraker[34089]:     self.uds_server = await asyncio.start_unix_server(
2023-07-21T18:41:14Z klipper.moonraker[34089]:   File "/usr/lib/python3.10/asyncio/streams.py", line 113, in start_unix_server
2023-07-21T18:41:14Z klipper.moonraker[34089]:     return await loop.create_unix_server(factory, path, **kwds)
2023-07-21T18:41:14Z klipper.moonraker[34089]:   File "/usr/lib/python3.10/asyncio/unix_events.py", line 306, in create_unix_server
2023-07-21T18:41:14Z klipper.moonraker[34089]:     sock.bind(path)
2023-07-21T18:41:14Z klipper.moonraker[34089]: OSError: AF_UNIX path too long
```

[...]

PR #690, opened 2023-07-21
allow configuration of unix path

mapero:
The pull requests allows setting the location for the moonraker unix socket. The maximum path length for unix sockets is limited to 108, so it might be necessary to change it in certain situations.

https://github.com/Arksine/moonraker/issues/689


Arksine:
Thanks.  I saw the issue before I noticed this PR and implemented something similar in commit 1fadae885ed1eb8c2447249d1c03a218db42c0c8.  The option name is different, I want to avoid confusing this option with Klipper's API Server option.

Issue #713, opened 2023-09-21
Webcam-API: Add/Expose UUID of the webcams


Clon1998:
### Is your feature request related to a problem? Please describe

Currently, as a Webcam-API user, I am forced to use the name of a webcam to identify and modify it.
I think we chatted a while back and both I and some other devs already suggested that Moonraker also exposes the Webcam's UUID and offer a new Endpoint that allows us to use the UUID to identify a webcam entry rather than the name of a webcam. The reason behind that is, that the name can be tricky to use and requires some additional checks, especially if the user of a UI renames a webcam.

If I remember correctly the UUID is already available/used in the moonraker-db.

### Describe the solution you'd like

1. Expose UUID of webcams (If they are not added via the config/read only)
2. Add a new endpoint that allows editing via uuidread-only

### Describe alternatives you've considered

Currently if a user modifies a webcam I am forced to implement it like that:

1. Call - Remove old webcam using the old/cached name 
2. Call - Add updated webcam values,

PR #714, opened 2023-09-21
Webcam: Added a uuid/id to the webcams to make it easier to use the API

Clon1998:
Hey, in #713, I suggested that we should use ID/UUID when addressing webcams instead of their names. I went ahead and implemented that. The implementation also still supports the addressing of webcams via names rather than the id.

I tested everything and updated the docs, so let me know if you spot any issues I missed. 🚀

Signed-off-by: Patrick Schmidt <ps.patti1998@gmail.com>

### DIFF
--- a/docs/web_api.md
+++ b/docs/web_api.md
@@ -3745,6 +3745,7 @@ A list of configured webcams:
 {
     "webcams": [
         {
+            "id": "59b7d256-0db9-47bf-bf5f-b9e4a6f61488",
             "name": "testcam3",
             "location": "door",
             "service": "mjpegstreamer",
@@ -3762,6 +3763,7 @@ A list of configured webcams:
             "source": "config"
         },
         {
+            "id": "2986b491-2bc8-4cad-9dae-c5a500ba6511",
             "name": "tc2",
             "location": "printer",
             "service": "mjpegstreamer",
@@ -3779,6 +3781,7 @@ A list of configured webcams:
             "source": "database"
         },
         {
+            "id": "cd68e494-b7c4-4ef0-a54e-61b8661bc9b6",
             "name": "TestCam",
             "location": "printer",
             "service": "mjpegstreamer",
@@ -3804,6 +3807,7 @@ A list of configured webcams:
 HTTP request:
 ```http
 GET /server/webcams/item?name=cam_name
+GET /server/webcams/item?uuid=cd68e494-b7c4-4ef0-a54e-61b8661bc9b6
 ```
 JSON-RPC request:
 ```json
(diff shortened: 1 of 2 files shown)


Arksine:
Thanks.  I recall this discussion and did look into it, concluding that exposing the UUID is not the best approach at this time.  I looked into an implementation similar to yours, however there are challenges in doing so:

1) Changing the `self.webcams` dict to key off of the ID instead of the name will break other components that depend on the `webcam` component.  Specifically the `simplyprint` component in Moonraker depends on `webcam`.  Additionally I think `timelapse` does externally.  It isnt worth the refactor in these components to make this change.

2)  The changes as implemented would allow for webcams with duplicate names.  This could result in unexpected behavior when attempting to lookup a webcam by name.  Webcam names must be unique as it isn't feasible to configure a UUID for webcams added to `moonraker.conf`.

3) The UUID was kept for compatibility with legacy frontends.  It was the best option for Mainsail to reliably generate unique keys, however a UUID is not the best option when ID assignment can be managed by Moonraker.  That said, I'm not sure we should risk breaking legacy clients at this time to change it.

IIRC the primary issue was renaming a webcam, as it requires two API calls.  I have a local branch that has resolved that particular problem, I haven't gotten around to fully testing it but I'll try to do so asap.

Clon1998:
> Thanks.  I recall this discussion and did look into it, concluding that exposing the UUID is not the best approach at this time.  I looked into an implementation similar to yours, however there are challenges in doing so:
> 
> 
> 
> 1) Changing the `self.webcams` dict to key off of the ID instead of the name will break other components that depend on the `webcam` component.  Specifically the `simplyprint` component in Moonraker depends on `webcam`.  Additionally I think `timelapse` does externally.  It isnt worth the refactor in these components to make this change.
> 
> 
> 
> 2)  The changes as implemented would allow for webcams with duplicate names.  This could result in unexpected behavior when attempting to lookup a webcam by name.  Webcam names must be unique as it isn't feasible to configure a UUID for webcams added to `moonraker.conf`.
> 
> 
> 
> 3) The UUID was kept for compatibility with legacy frontends.  It was the best option for Mainsail to reliably generate unique keys, however a UUID is not the best option when ID assignment can be managed by Moonraker.  That said, I'm not sure we should risk breaking legacy clients at this time to change it.
> 
> 
> 
> IIRC the primary issue was renaming a webcam, as it requires two API calls.  I have a local branch that has resolved that particular problem, I haven't gotten around to fully testing it but I'll try to do so asap.

Sure! I was unaware about the dependencies.

Arksine:
After some further thought and review I came up with a solution to add uid support without breaking local components.  I chose `uid` over `uuid` as I wanted the parameter/field to be more generic in the event that I want to move away from uuids in the future.

The change was made in commit a2322607135bf110061d269d021e1ba7b1ad982b.    Feel free to give it a try and let me know if there are any issues or regressions.  Thanks.

PR #717, opened 2023-09-27
Add basic socket activation support

trchen1033:
Hello Eric,
I wrote a proof-of-concept implementation to support systemd socket activation. The motivation is is that I use an Nginx reverse proxy to host Moonraker and Fluidd under the same CORS domain, and I also want Nginx to handle TLS so I can easily integrate certbot.
There are some other advantages with socket activation:
* Can listen to privileged ports without running as root
* Unix socket support (or any arbitrary kind of SOCK_STREAM) comes for free, and we avoid the subtleties around setting up ownership and umask for the newly created socket.
* Seamless service restart

This implementation is only a quick hack, least resistance path towards my use case. I would like to chat with you before jumping into a full implementation. Specifically there are a few issues I want to hear your opinion about:
* Do we want to support TLS with it?
* Do we want socket activation for the JSON-RPC socket as well?
* How should we provide backward compatibility if we add new config option to listen to arbitrary number of sockets?
* How should we handle `Server.get_host_info()`, HTTP `Location` header and generally other things that need to be aware of the server's published address?

### DIFF
--- a/moonraker/app.py
+++ b/moonraker/app.py
@@ -10,17 +10,21 @@ import mimetypes
 import logging
 import json
 import traceback
+import socket
 import ssl
 import pathlib
 import urllib.parse
 import tornado
 import tornado.iostream
 import tornado.httputil
+import tornado.netutil
 import tornado.web
+from contextlib import closing
 from inspect import isclass
 from tornado.escape import url_unescape, url_escape
 from tornado.routing import Rule, PathMatches, AnyMatches
 from tornado.http1connection import HTTP1Connection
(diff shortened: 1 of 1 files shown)


Arksine:
Hi.  Unfortunately I missed this PR and its nearly 1 year old now.  As you mention, there are issues internally with moonraker that would need to be addressed to support systemd socket activation.  At this time I believe that its best to leave things as they are, however I'm open to further discussion.  In the meantime I'll close this PR, however its fine to post in it and go into further detail about a fully integrated proposal.

PR #732, opened 2023-10-26
docs: add Domoticz power section

olivierberthomme:
Example of a Domoticz integration for power module (tested with domoticz release 2023.2)

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1198,6 +1198,58 @@ token: smartthings-bearer-token
 device: smartthings-device-id
 ```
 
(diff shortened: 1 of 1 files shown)


Arksine:
Thanks.  The change looks ok, however the commit [needs to be signed off](https://moonraker.readthedocs.io/en/latest/contributing/#git-commit-format) before merging.

I may consider creating a folder in the docs path and moving the power examples there, however for the time being it is ok for them to be inline.

olivierberthomme:
Replaced by #742

PR #733, opened 2023-10-26
add energy_meter component: energy usage per job

### LINKED ISSUE #445: Power consumption monitoring

No - just a request.

I'd like to also see and log the power consumption per print in the history of i.e. Mainsail. That will require a additional attribute in moonraker.

Plugins for the frontend would be possible, or just read it from the smart power plug. But then there is no automatic logging and automatic price calculation is not possible.

### Additional information

_No response_

PSandro:
fixes #445 

This adds a new component `energy_meter` which is used for starting and stopping a energy measurement. For the actual data, a sensor is to be specified in the config section. The `history` component collects print stats already for jobs, I added the energy measuring there:  On print start, the current value from the sensor's energy counter is saved. When the print stops, this saved value then gets subtracted from the most recent sensor value and voila, we get the energy used for the print.

Here's a sample configuration used for a ShellyPlus1PM:
```
[mqtt]
address: 192.168.230.1
username: {secrets.mqtt.username}
password: {secrets.mqtt.password}
enable_moonraker_api: True

[sensor powermeter]
type: mqtt
name: powermeter
state_topic: shellies/shellyplus1pm-4855199da2e4/status/switch:0
state_response_template:
  {% set notification = payload|fromjson %}
  {set_result("power", notification["apower"]|float)}
  {set_result("voltage", notification["voltage"]|float)}
  {set_result("current", notification["current"]|float)}
  {set_result("energy", notification["aenergy"]["by_minute"][0]|float * 0.000001)}

[energy_meter]
sensor: powermeter
field: energy
```

### DIFF
--- a/moonraker/components/energy_meter.py
+++ b/moonraker/components/energy_meter.py
@@ -17,7 +17,7 @@ from typing import (
 
 if TYPE_CHECKING:
     from ..confighelper import ConfigHelper
-    from .sensor import BaseSensor, Sensors
+    from .sensor import BaseSensor
 
 
 @dataclass(frozen=True)
@@ -39,8 +39,7 @@ class DeltaMeasurement:
 class EnergyMeter:
     def __init__(self, config: ConfigHelper) -> None:
         self.server = config.get_server()
-        sensors: Sensors = self.server.load_component(config, "sensor")
-        self.sensors = sensors.sensors
+        self.sensors = self.server.load_component(config, "sensor")
         self.config = EnergyMeterConfiguration(
             sensor=config.get("sensor"),
             field=config.get("field", "energy"),
@@ -93,7 +92,7 @@ class EnergyMeter:
                 "Can not start another measurement when currently active."
             )
             return None
-        sensor: Optional[BaseSensor] = self.sensors.get(self.config.sensor)
+        sensor: Optional[BaseSensor] = self.sensors.get_sensor(self.config.sensor)
         if sensor is None:
             logging.warning(
                 "Could not start measurement: sensor '%s' is missing",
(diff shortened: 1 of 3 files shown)


kondas:
Awesome! Works flawless using generic power meter, example config:

```
[sensor powermeter]
type: mqtt
name: Power
state_topic: zigbee2mqtt/pow-printi
state_response_template:
  {% set notification = payload|fromjson %}
  {set_result("energy", notification["energy"]|float)}

[energy_meter]
sensor: powermeter
field: energy
```

PSandro:
Thanks for testing @kondas !
image
This is how the energy consumption information is then included in the history api endpoint.

I think for now this is ready for review. I was thinking about an edgecase where moonraker restarts and the state of energy_meter is lost - but in this case we might have a gap in measurement anyways and it kinda makes sense for the measurement reporting to 'fail'.

kondas:
Bumping this, because i'd really like to see this feature merged

Anywherexu:
[moonraker.log](https://github.com/Arksine/moonraker/files/13785537/moonraker.log)
**Hello, this is a machine translation I don't know if I can express it. Thank you for your understanding.**
**I added the sensor from mqtt. Homeassistant is receivable, but fluidd cannot be displayed. There is an error in the moonraker .log file. Don't know how to fix it. Ask for guidance.**

Arksine:
I know it was a long time coming, but I finally found some time to complete support for adding auxiliary fields to the `job history`.  Front ends will need to implement it, however you can manually call the API and view the data.  Each field is presented with a consistent schema, which should make it easier for frontends to determine how to display the data.

The [documentation](https://moonraker.readthedocs.io/en/latest/configuration/#sensor) explains how to configure one or more "history fields" for a sensor.  In addition, a sample that should work with tasmota devices is provided.

Arksine:
Since support for this is now available I will close this PR.  Thanks.

PR #740, opened 2023-11-03
[FIX] Update metadata.py to work with Ideamaker

Manu512:
metadata : fix extract to work with Ideamaker

Correction of the regex (Replacing the = with one: and adding the filament index for the regex to work) 
Correction du regex (Remplacement du = par un : et ajout de l'index du filament pour que le regex fonctionne)

Signed-off-by: Emmanuel WALDNER <emmanuel.waldner@gmail.com>

### DIFF
--- a/moonraker/components/file_manager/metadata.py
+++ b/moonraker/components/file_manager/metadata.py
@@ -775,7 +775,7 @@ class IdeaMaker(BaseSlicer):
         if filament:
             return sum(filament)
         return None
-        
+
     def parse_filament_type(self) -> Optional[str]:
         return (
             _regex_find_string(r";Filament\sType\s.\d:\s(.*)", self.header_data) or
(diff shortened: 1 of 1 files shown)


Arksine:
Thanks.  Presumably IdeaMaker now officially supports this metadata, whereas previously it was necessary for the user to add it in their start gcode.  I'm ok with the corrections, however I'm not sure that Moonraker should drop support for the legacy entries immediately.  We could do something like the following:
```python
def parse_filament_type(self) -> Optional[str]:
    return (
        _regex_find_string(r";Filament\sName\s.\d:\s(.*)", self.header_data) or
        _regex_find_string(r";Filament\stype\s=\s(.*)", self.header_data)
    )
```

The above would prefer the new style, then fallback to the legacy expression if it isn't found.  We could modify the expression to handle either case, however doing it this way will make it easier to remove the deprecated expression the future.

Arksine:
Thanks.

PR #751, opened 2023-11-19
power: refactor component in order to dynamically import plugins

cedbossneo:
I think it's a good idea to make moonraker modulable in the same way as klippy does.

By adding new files in the extras/power directory, they are dynamically imported so we can have a plugin mechanism


Arksine:
Moonraker's component system behaves in a similar (not identical) manner to that of Klippy's extras.  That said, neither are intended to be used to support 3rd party plugins.  Developers may choose to use them this way, but internal code changes to Moonraker could break the extension.

I don't think this PR is a good candidate for the main repo.  While not officially supported, it is possible to create a 3rd party power `component`.  The 3rd party code would load the `power` component and call its `add_device()` method.  In the future I plan to consider official 3rd party plugin support, and this is the method in which a 3rd party power device would be added.

In addition, the power component supports generic http devices.  Going forward I will tend to avoid merging new HTTP device implementations and instead direct users to configure the device using the generic type.

cedbossneo:
Ok, good to know, so i should create another repository that add the TPLink Tapo support (that was the mail goal of my two PRs).

Arksine:
If you want you could create something like [moonraker-timelapse](https://github.com/mainsail-crew/moonraker-timelapse/).   The install script creates a symbolic link to the python module in Moonraker's component folder and installs required dependencies, such as pycryptodome.

PR #758, opened 2023-11-27
Added support for systems without gpio or not suppoting the "python3-libgpiod" lib

galbb12:
Hi, it's my first contribution ever so please don't judge me too much,
I use the project on an android tv box which I installed armbian on and the installer couldn't install the "python3-libgpiod" lib and failed, So i added a check on the "install-mooneraker.sh" script to check if the "python3-libgpiod" apt package is installable if it is it runs the install process normally and if it isn't it skips the installation of the "python3-libgpiod" package by installing the packages from the file "system-dependencies-without-gpio.json" I added instead of the  "system-dependencies.json" file.

### DIFF
--- a/scripts/install-moonraker.sh
+++ b/scripts/install-moonraker.sh
@@ -47,7 +47,7 @@ install_packages()
     if [apt-cache search grep | grep -q "python3-libgpiod"]; then
      system_deps="${SRCDIR}/scripts/system-dependencies.json"
     else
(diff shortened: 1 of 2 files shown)


Arksine:
Thanks.  The change itself looks ok, however the install script is intended for standard Debian distros and derivatives.  One goal is to keep the installer as straight forward as possible so it can be used as a template for custom installations.  Therefore I don't think its appropriate to merge this change.

At some point in the future I will likely write a more robust installer in Python intended to support more distros and obscure derivatives.

galbb12:
Ok I understand, I am closing the request.

PR #761, opened 2023-12-04
Explain stream types

WayneManion:
Add a list of allowed stream types. I thought this was valuable because the stream types are not explained in this document. I had to look in the fluidd codebase to find a list of valid values. The valid values are not the same as the options shown in the fluidd webUI.

https://github.com/fluidd-core/fluidd/blob/999ccddd1a39d6037ec6a8f34b689b0216ea592d/src/components/settings/cameras/CameraConfigDialog.vue#L86

Signed-off-by: Wayne Manion treowayne@gmail.com

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -355,7 +355,7 @@ enabled: True
 service: mjpegstreamer
 #   The name of the application or service hosting the webcam stream.  Front-
 #   ends may use this configuration to determine how to launch or start the
-#   program.  The default is "mjpegstreamer".
+#   program.  The default is "mjpegstreamer". Other valid values include: 'mjpegstreamer-adaptive', 'hlsstream', 'webrtc-camerastreamer', 'ipstream', 'iframe', and 'webrtc-go2rtc'.
 target_fps: 15
 #   An integer value specifying the target framerate.  The default is 15 fps.
 target_fps_idle: 5


Arksine:
@bakatrouble is correct.  This available stream types are determined by the frontend and not constrained by Moonraker.  It is probably best to simply direct users to the front end's documentation as is noted in the paragraph preceding the example.

WayneManion:
I get what you're saying. It isn't Moonraker's job to tell what cameras the front ends can support.

[Here is fluidd's documentation on cameras](https://docs.fluidd.xyz/features/cameras). 

[Here is Mainsail's camera docs. ](https://docs.mainsail.xyz/overview/features/webcam)

Neither one of them documents anything about how to specify the camera settings through moonraker.conf. I don't think a user should have to dig through source code to find the list of strings needed for a particular type of camera to work. Maybe the solution is more thorough documentation of fluidd and Mainsail.

Arksine:
I took a more detailed look at this and agree that Moonraker's documentation needs to provide more guidance here.  The frontend documentation explains how to configure webcams through their interfaces, but abstracts the underlying values.

I have updated the [webcam documentation](https://moonraker.readthedocs.io/en/latest/configuration/#webcam) to include tables for known `icon` and `service` values.

WayneManion:
That is awesome! Thanks for the thorough and helpful documentation.

Arksine:
Thanks.  I'll go ahead and close this as resolved.

PR #791, opened 2024-01-15
Fix if spoolman spool is missing

Zeanon:
Currently if the last selected spool gets deleted from the spoolman database, the spoolman component fails to load. This fix sets the selected spool to None if that happens so the component can load

### DIFF
--- a/moonraker/components/spoolman.py
+++ b/moonraker/components/spoolman.py
@@ -195,7 +195,14 @@ class SpoolManager:
             url=full_url,
             body=body,
         )
-        response.raise_for_status()
+
+        if (response._code == 404
+                and self.spool_id is not None
+                and dict(response.json()).get("message")
+                == ("No spool with ID %d found." % self.spool_id)):
+            await self.set_active_spool(None)
+        else:
+            response.raise_for_status()
 
         return response.json()
 


Arksine:
Hi.  A 404 returned by the proxy request would not cause a failure to load the Spoolman component.  The front end is misreporting it as such because it believes that the proxy endpoint doesn't exist due to the 404, when its Spoolman is returning 404.  While this is understandable, I think that frontends need to treat errors as if they were returned by Spoolman when using this endpoint.  Moonraker reports a list of loaded components and failed components in its `/server/info` endpoint. 

I don't think its a good idea to change the active spool in the proxy endpoint.  The message string isn't a reliable source for comparison, if the error response changes just one character we will always evaluate to False.   In addition, having Moonraker report the message in place of a 404 error represents an API change that some clients may not be able to deal with.

There may be some things we can do better to help alleviate this issue, but regardless of what we do changes will need to be made at the frontend to accommodate it.

Arksine:
I just opened #792 with the aim of resolving this issue.  I makes some changes that should reduce the chance that Moonraker's spool id does not exist in the Spoolman Database.  In addition it makes an API change that will allow frontends to deal with the ambiguous 404 error, however it will require implementation on their side.

> But out of curiosity: the component failed to load message you get in mainsail, is that from mainsail or moonraker?

Based on the behavior described, I'm reasonably certain that the component hasn't actually failed.  The ambiguous 404 returned when a spool doesn't exist translates to `Method not found`, so Mainsail thinks that the component isn't loaded when it is.

Arksine:
This should now be resolved on Moonraker's side.  The major known frontends have changes complete or in process.

PR #804, opened 2024-01-31
Bump preprocess_cancellation to 0.2.1

kageurufu:
0.2.1 ensures that the exclude_object markers are the first
 non-comment g-code in the files, fixing issues with klipper adaptive
 meshing


Arksine:
Thanks.

PR #812, opened 2024-02-15
Allow configuring the update refresh window

kdomanski:
This enables customizing the time of day within which periodic refresh occurs.

Previously docs said it happens between 12am-4am which is misleading, because the code effectively allowed refresh until 4:59:59 AM. The default window is now '0-5', which is practically the same as before.

The comment about checking every 24h hours has been fixed as well. This hasn't been true since commit c4fa76f21792943ea512d24326259f01d2fa7e89

### DIFF
--- a/moonraker/components/update_manager/update_manager.py
+++ b/moonraker/components/update_manager/update_manager.py
@@ -48,8 +48,6 @@ if TYPE_CHECKING:
 
 # Check To see if Updates are necessary each hour
 UPDATE_REFRESH_INTERVAL = 3600.
-# Perform auto refresh no later than 4am
-MAX_UPDATE_HOUR = 4
 
 def get_deploy_class(
     app_type: Union[AppType, str], default: _T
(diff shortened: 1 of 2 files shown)


moonraker/components/update_manager/update_manager.py
```diff
@@ -126,6 +124,18 @@ def __init__(self, config: ConfigHelper) -> None:
         # Auto Status Refresh
         self.refresh_timer: Optional[FlexTimer] = None
         if auto_refresh_enabled:
+            self.refresh_window = config.getintlist('refresh_window', [0, 5],
```
Arksine, line 127: This should be moved outside of the condition.  This way a user can toggle auto refresh on or off without removing "refresh_window" in the configuration. 

In addition it guarantees that the `self.refresh_window` attribute is set.  While it shouldn't be reachable when auto refresh is disabled, we avoid potential future bugs if there is a refactor that makes it reachable.

Arksine:
Thanks.  Overall it looks like a good addition, I just had one comment.

moonraker/components/update_manager/update_manager.py
kdomanski: Right. I forgot about the warning when an unused option is set. Fixed it now.
I could not decide whether the validation belongs up there with `config.getintlist` or right above the `if auto_refresh_enabled:` section. Let me know if you want it moved again.

Arksine (approved):

Arksine:
Thanks.

Issue #504, opened 2022-09-17
Support Prometheus-format metrics output


wywywywy:
### Is your feature request related to a problem? Please describe

It's a useful & common feature nowadays to support exporting metrics in Prometheus format, so that stats can be put into dashboards (e.g. through Grafana) or used for future analysis/debugging.

### Describe the solution you'd like

Add a new component `Metrics` that registers the `/metrics` endpoint, which returns all available metrics. Enabled/disabled through a new setting.

The output will be something like...

```
moonraker_print_print_duration{printer="Voron 2.4"} 16.12345
moonraker_heater_bed_temperature{printer="Voron 2.4"} 65.06
moonraker_heater_bed_power{printer="Voron 2.4"} 0.23456
moonraker_heater_bed_target{printer="Voron 2.4"} 65
...etc
```

### Describe alternatives you've considered

It could also be done as a standalone separate Prometheus exporter, not as part of Moonraker. I have previous done one for OctoPrint here - https://github.com/wywywywy/octoprint_exporter. But in terms of the user experience it'll be much cleaner to be done as part of Moonraker (and subsequently Klipper).

scross01:
I created a standalone Prometheus exporter that scrapes the Moonraker API endpoints for suitable metrics. https://github.com/scross01/prometheus-klipper-exporter. However I do agree with @wywywywy that building a Prometheus `/metrics` endpoint directly into Klipper would be the better solution.

PR #816, opened 2024-02-19
POC: Prometheus metrics module

kdomanski:
This is a POC to address #504 . It is meant to spawn feedback, but is still neither complete nor satisfactory to me.

- For now the only notable metrics that are exported are statuses of any heaters or temperature sensors available
- No apikey/auth support
- Experience in instrumenting software tells me, that it would be better to instrument printer objects directly in Klippy and then only add Moonraker-specific metrics from here. Need to think about this.
- I'm not sure whether this is race-free. Could a status update callback land after klippy shutdown/disconnect callback?

### DIFF
--- a/moonraker/components/prometheus.py
+++ b/moonraker/components/prometheus.py
@@ -5,29 +5,38 @@
 # This file may be distributed under the terms of the GNU GPLv3 license.
 
 from __future__ import annotations
+from asyncio import gather
+import logging
 from prometheus_client import (
     exposition, registry,
-    Info
+    Info, Gauge
 )
+from prometheus_client.metrics import MetricWrapperBase
 
 from ..common import (
+    KlippyState,
     RequestType,
     TransportType,
     WebRequest
 )
 
+from .klippy_apis import KlippyAPI
+
 from typing import (
-    TYPE_CHECKING
+    TYPE_CHECKING,
+    Dict,
+    Any
 )
 
 if TYPE_CHECKING:
     from ..confighelper import ConfigHelper
 
-class PrometheusClient:
+class PrometheusExporter:
     def __init__(self, config: ConfigHelper) -> None:
         self.server = config.get_server()
         app_args = self.server.get_app_args()
 
+        # Not saved in the object, because it should not be changed or cleared.
         i = Info('moonraker_instance', '')
         i.info({
             'version': app_args['software_version'],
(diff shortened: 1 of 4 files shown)


kdomanski:
Screenshot 2024-02-19 095435
Screenshot 2024-02-19 095452

moonraker/components/prometheus.py
```diff
@@ -0,0 +1,157 @@
...
+
+    async def _handle_metrics_endpoint(self, web_request: WebRequest) -> bytes:
+        """Writes metrics in response to the scrape.
+
+        Usually some properties of the response depend on request headers.
+        To make request headers available here, a serious refactoring would be needed.
+        Instead, we make some assumptions:
+        - Response will be in the "normal" format and not openmetrics
+        - No filtering by name will be done
+        - gzip won't be used (also sparing the CPU cycles in return for bandwidth)
```
Arksine, line 152: It wouldn't be difficult to provide the `WebRequest` access to the request headers if needed.  This hasn't been required to date so I haven't done so.

As far as gzip compression, a reverse proxy could handle that if its desired.  I don't think it needs to be in Moonraker.

moonraker/components/prometheus.py
```diff
@@ -0,0 +1,157 @@
...
+        # clear metrics to stop providing metrics when their value is simply unknown
+        self.server.register_event_handler(
+            "server:klippy_shutdown", self._clear_metrics
+        )
+        self.server.register_event_handler(
+            "server:klippy_disconnect", self._clear_metrics
+        )
+        self.server.register_event_handler(
+            "server:klippy_disconnected", self._clear_metrics
+        )
```
Arksine, line 73: Moonraker doesn't send a `klippy_disconnected` event.  It does register a `notify_klippy_disconnected` websocket notification that piggybacks off of the `server:klippy_disconnect` event.

Arksine:
Thanks, I have a few questions/comments based on your initial feedback.

> No apikey/auth support

Is this a limitation in Prometheus?  Does it support any type of authentication?

> Experience in instrumenting software tells me, that it would be better to instrument printer objects directly in Klippy and then only add Moonraker-specific metrics from here. Need to think about this.

I'm not sure Kevin would be enthusiastic about merging instrumentation directly.   Its effectively repackaging the data, sending instrumented objects to Moonraker would duplicate the serialization/deserialization effort, presumably with additional overhead.  Part of the motivation for Klipper's API server is to farm this kind of work to applications outside of Klippy.

> I'm not sure whether this is race-free. Could a status update callback land after klippy shutdown/disconnect callback?

Status updates can still be received after shutdown, but not after a disconnect.

moonraker/components/prometheus.py
kdomanski: > As far as gzip compression, a reverse proxy could handle that if its desired. I don't think it needs to be in Moonraker.
You are absolutely right. These notes basically address all the parts of prometheus client's default handler that I stripped in order to get this integrated.

kdomanski:
> > No apikey/auth support
> 
> Is this a limitation in Prometheus? Does it support any type of authentication?

No, this is my mistake. I forgot that Moonraker api keys work simply as bearer tokens. That works just fine with Prom.
 
> > Experience in instrumenting software tells me, that it would be better to instrument printer objects directly in Klippy and then only add Moonraker-specific metrics from here. Need to think about this.
> 
> I'm not sure Kevin would be enthusiastic about merging instrumentation directly. Its effectively repackaging the data, sending instrumented objects to Moonraker would duplicate the serialization/deserialization effort, presumably with additional overhead. Part of the motivation for Klipper's API server is to farm this kind of work to applications outside of Klippy.

And this is exactly why I think we'll be pondering this for a while. I understand and share the desire to minimize any extraneous stuff in Klippy itself. But I just don't think this particular thing can be farmed out effectively.

In Klippy, there's access to the actual objects in question, type information, introspection, and the like. It should be possible to do most of the instrumentation from the Prometheus module itself (with a little bit of introspection) and not litter every other module manually with instrumentation code.

But doing it in Moonraker means inferring all that information which Klippy already has, based on an implicit API (status updates), which doesn't have a clear schema. Any change in Klippy must be followed by a change in Moonraker and a bespoke converter from status JSON to metrics for each component type. That's overhead too. And unlike status updates, which are strictly scheduled, the Prometheus metrics are only serialized during scrape.
In the end, I'd rather not argue about performance and/or overhead without having actual measurements thereof.
 
> > I'm not sure whether this is race-free. Could a status update callback land after klippy shutdown/disconnect callback?
> 
> Status updates can still be received after shutdown, but not after a disconnect.

I failed to articulate my question clearly. What I mean is: if a status updates comes right before a disconnect, will their handlers still execute in order every time, or is there a risk that after clearing the metrics (on disconnect, on shutdown, etc.) a wayward update handler still runs and writes a value from the old session?

PR #831, opened 2024-03-13
docs: corrects canbus peripherals response example

pedrolamas:
The current example shows `can_uuids` as an array of strings, when in fact this is an array of objects with `uuid` and `application` properties.


Arksine:
Thanks.

PR #834, opened 2024-04-02
Support Hubitat for power device

kevdliu:
This PR adds support for setting up a Hubitat connected smart switch as a power device. It requires the following config parameters:
- `device_id`: ID of the smart switch device
- `maker_id`: ID of the Maker API application. Maker API is the Rest API server for Hubitat. 
- `token`: Access token for Maker API
- `status_delay`: Delay between controlling the smart switch and refreshing its state

I have my Hubitat hub set up with a self-signed certificate, so I also needed to disable certificate verification for it to work. This PR also adds an universal `validate_ssl_cert` config parameter that disables ssl cert verification for power device requests.

### DIFF
--- a/moonraker/components/http_client.py
+++ b/moonraker/components/http_client.py
@@ -80,7 +80,8 @@ class HttpClient:
         retry_pause_time: float = .1,
         enable_cache: bool = False,
         send_etag: bool = True,
-        send_if_modified_since: bool = True
+        send_if_modified_since: bool = True,
+        validate_ssl_cert: bool = True
     ) -> HttpResponse:
         cache_key = url.split("?", 1)[0]
         method = method.upper()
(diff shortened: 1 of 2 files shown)


Arksine:
Thanks.  I'm no longer merging pull requests for additional HTTP implementations. Instead I recommend using the [Generic HTTP type](https://moonraker.readthedocs.io/en/latest/configuration/#generic-http-devices) to support devices not  natively available.  Frankly, there are too many vendors offering http switches and its not feasible to maintain all of them in Moonraker.  This was the motivation for creating the generic type.

With regard to SSL, I think that the client should always validate the certificate when https is configured.  When using self signed certificates you should install your own root certificate on the machine hosting Moonraker.  See #799 for some discussion on this topic.

kevdliu:
Ah that's unfortunate but totally understandable. I appreciate the in-depth explanation. I'll look into installing my CA cert on the device. Thanks!

PR #839, opened 2024-04-08
data_store: allow users to specify extra fields to track

SanaaHamel:
Allow users to specify additional sensor fields to track in the data-store.
Use case: Provide history support for sensors, e.g. humidity/pressure/gas.


Arksine:
Thanks.  I'm leery of configurable options here, as it could make the API unpredictable and cause problems for frontends.   We could potentially add direct support for some of these items like humidity, pressure, etc.  An alternative would be to track configurable items in a second structure and report them in a different endpoint.

PR #840, opened 2024-04-12
Notifier Print Progress Enhancements

thankyousam:
Notifier.py: Enable Layer Change progress messages through Notifier

Expanded the stub of existing code in notifier.py to generate a layer change notification at user specified % layer progress points during the print. This allows the user to receive notifications at set progress points. Only the current layer and total layer count is available in the print state data, so this doesn't necessarily represent linear progress.

Before this change it was only possible to automatically generate a notification at the start or end of the print, or if something went wrong and caused an error or pause.

Additional inclusion of a timestamp variable in all Notifier event data to allow the event time to be included in the notification text.

See the additional Notifier sections in the configuration.md documentation for usage.

Files updated:
 moonraker/common.py
 moonraker/components/job_state.py
 moonraker/components/notifier.py
 docs/configuration.md

Signed-off-by: David Osbourn <thankyousam@gmail.com>

### DIFF
--- a/moonraker/common.py
+++ b/moonraker/common.py
@@ -109,6 +109,7 @@ class ExtendedEnum(Enum):
         return self._name_.lower()  # type: ignore
 
 class JobEvent(ExtendedEnum):
+    LAYER_CHANGED = 0
     STANDBY = 1
     STARTED = 2
     PAUSED = 3
--- a/moonraker/components/notifier.py
+++ b/moonraker/components/notifier.py
@@ -10,6 +10,7 @@ import apprise
 import logging
 import pathlib
 import re
+from datetime import datetime  # To allow timestamp to be generated
 from ..common import JobEvent, RequestType
 
 # Annotation imports
@@ -41,7 +42,13 @@ class Notifier:
                     if job_event == JobEvent.STANDBY:
                         continue
                     evt_name = str(job_event)
-                    if "*" in notifier.events or evt_name in notifier.events:
+                    # add extra AND condition (and evt_name != 'layer_change')
+                    # to * to exclude layer_change from being automatically picked
+                    # up by the * otherwise it triggers on EVERY layer change
+                    if ("*" in notifier.events and evt_name != 'layer_changed') or \
+                            evt_name in notifier.events:
+
+                        logging.info(f"Job Event loading: {evt_name}")
                         self.events.setdefault(evt_name, []).append(notifier)
                 logging.info(f"Registered notifier: '{notifier.get_name()}'")
             except Exception as e:
@@ -53,6 +60,9 @@ class Notifier:
         self.register_endpoints(config)
         self.server.register_event_handler(
             "job_state:state_changed", self._on_job_state_changed
+        )
+        self.server.register_event_handler(
+            "job_state:layer_changed", self._on_job_layer_changed
         )
 
     def register_remote_actions(self):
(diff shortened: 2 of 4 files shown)


moonraker/components/notifier.py
```diff
@@ -74,6 +84,36 @@ async def _on_job_state_changed(
...
+                current = int(new_stats["info"]["current_layer"])
+                # calculate if we've hit a % trigger point above the min height
+                # this isn't nice to read because of the 88 char line limit
+                if layer_trigger != 0 and total >= minimum_layers and current > 0 and \
+                    current != total and \
+                    round(current/total/layer_trigger, 5)//1 != \
+                        round(((current-1)/total)/layer_trigger, 5)//1:
+
+                    logging.info(f"Layer change notification at \
+                    {round(current/total/layer_trigger,5)//1*layer_trigger*100}%")
```
nmaggioni, line 114: Continuing over a new line with a backslash includes the indentation as actual characters, producing spaced out output like `Layer change notification at                     25.0%`.
```suggestion
                    logging.info(
                        "Layer change notification at "
                        f"{round(current/total/layer_trigger,5)//1*layer_trigger*100}%"
                    )
```

moonraker/components/notifier.py
```diff
@@ -123,6 +163,16 @@ def __init__(self, config: ConfigHelper) -> None:
...
         self.attach = config.gettemplate("attach", None)
         url_template = config.gettemplate("url")
+        self.layer_trigger = float(config.get("layer_trigger", 0))
+        self.minimum_layers = int(config.get("minimum_layers", 0))
+        if self.layer_trigger < 0 or self.layer_trigger > 1:
+            # layer_trigger out of range. Default to 0 and send logging info
+            logging.info(
+                f"Layer change trigger out of range for Notifier {self.name}. "
+                + f"Expected decimal between 0 and 1, found {self.layer_trigger}. "
+                + "Ignoring layer change trigger."
```
nmaggioni, line 173: Operators can be skipped, strings between brackets will still be concatenated:
```suggestion
                f"Layer change trigger out of range for Notifier {self.name}. "
                f"Expected decimal between 0 and 1, found {self.layer_trigger}. "
                "Ignoring layer change trigger."
```

moonraker/common.py
```diff
@@ -109,6 +109,7 @@ def __str__(self) -> str:
         return self._name_.lower()  # type: ignore
 
 class JobEvent(ExtendedEnum):
+    LAYER_CHANGED = 0
```
nmaggioni, line 112: Wouldn't it be better to append the new event to the bottom of the Enum with a value of `8`?

nmaggioni:
I've been using your changes for a couple of weeks (thanks!) but I can't seem to access the filename in the progress notification like I do with the other notifiers:

```
[notifier progress]
url: # ...
events: layer_changed
layer_trigger: 0.25
body: Progress printing {event_args[1].filename or 'N/A'}: {(event_args[1].info.current_layer / event_args[1].info.total_layer * 100 if event_args[1].info.total_layer is not none and event_args[1].info.total_layer > 0 else 0) | int}% (layer {event_args[1].info.current_layer} of {event_args[1].info.total_layer})
```

The above config produces this notification:

> Progress printing N/A: 30% (layer 4 of 13)

Other kinds of events work as expected instead:

```
[notifier starting]
url: # ...
events: started
body: {event_name|title}{'d' if event_name.endswith('e') else ''} printing {event_args[1].filename}{' (%s)' % event_args[1].message if event_args[1].message|length > 0 else ''}
```

> Started printing <correct_filename_here>.gcode

Dumping the stats during a print shows that the `filename` field is correctly populated: `printer['print_stats'].filename = <correct_filename_here>.gcode`.

I'll add some debugging output and report back.

moonraker/components/notifier.py
thankyousam: Thanks for the code review and feedback. I wasn't used to restricting line lengths, but now I know :). I'll commit these changes.

moonraker/common.py
thankyousam: I did spot this when I was coding it originally. Setting it to 8 would require additional changes a few lines further down in common.py which determines if the printer has finished or aborted. Adding the new state as 0 (zero) didn't disturb that existing code. 

```
    @property
    def finished(self) -> bool:
        return self.value >= 5

    @property
    def aborted(self) -> bool:
        return self.value >= 6
```

We could change it to this *untested* code, but I don't know if that might cause any problems elsewhere.

```
    LAYER_CHANGED = 8

    @property
    def finished(self) -> bool:
        return (self.value >= 5 and self.value != 8)

    @property
    def aborted(self) -> bool:
        return (self.value >= 6 and self.value != 8)

```

thankyousam:
> I've been using your changes for a couple of weeks (thanks!) but I can't seem to access the filename in the progress notification like I do with the other notifiers:
> 
> ```
> [notifier progress]
> url: # ...
> events: layer_changed
> layer_trigger: 0.25
> body: Progress printing {event_args[1].filename or 'N/A'}: {(event_args[1].info.current_layer / event_args[1].info.total_layer * 100 if event_args[1].info.total_layer is not none and event_args[1].info.total_layer > 0 else 0) | int}% (layer {event_args[1].info.current_layer} of {event_args[1].info.total_layer})
> ```
> 
> The above config produces this notification:
> 
> > Progress printing N/A: 30% (layer 4 of 13)
> 
> Other kinds of events work as expected instead:
> 
> ```
> [notifier starting]
> url: # ...
> events: started
> body: {event_name|title}{'d' if event_name.endswith('e') else ''} printing {event_args[1].filename}{' (%s)' % event_args[1].message if event_args[1].message|length > 0 else ''}
> ```
> 
> > Started printing <correct_filename_here>.gcode
> 
> Dumping the stats during a print shows that the `filename` field is correctly populated: `printer['print_stats'].filename = <correct_filename_here>.gcode`.
> 
> I'll add some debugging output and report back.

Thanks again for reviewing the code and testing this out.

The layer change event only appears to have the info payload from the printer. The filename only appears to be populated in the state changes, which is why it's there for started, complete, etc.

This is the extent of what I get in event_args[1] in the layer_change event.

```
{'total_duration': 19.789630468934774, 'info': {'total_layer': 10, 'current_layer': 8}}
```

I don't know if we could work backwards up the event stack to see if the filename can be pulled through. I couldn't figure it out at the time. Could it be added into the GCode and pulled through that way like the total and current layer counts?

nmaggioni:
@thankyousam Filename is now populated, thanks!

> Progress printing <correct_filename_here>.gcode: 53% (layer 7 of 13)

PR #843, opened 2024-04-22
spoolman: Include total count header

### LINKED ISSUE #841: Spoolman Proxy: Include total count header in response

Through the Spoolman proxy, I am able to interact with any of the API endpoints of Spoolman. However, the find endpoints take up a special role. They not only return data via the body but also contain the total entity count by including a `X-Total-Count` field within the response header.

Currently, this information is unavailable via the proxy but is required for UI implementations and pagination.

Can we adjust the response of the proxy method to also include the value of the header, if available?

Implementing spoolman pagination without counts...
It can work but is not as good...

### Additional information

_No response_

Clon1998:
Closes #841.

As noted in the linked issue, spoolman uses the `x-total-count` header field in its response to indicate the total count, if appropriate, for a pagination result.

To also include this in the proxy I modified the returned JSON of the V2 response to not only return the `response` but also a `response_header` field with all custom headers (Headers starting with `x-`).

I am open to suggestions or alternative ideas on how to handle this change. I feel that adding it as part of the V2 is the safest and will not break any current implementations.

### DIFF
--- a/moonraker/components/spoolman.py
+++ b/moonraker/components/spoolman.py
@@ -373,6 +373,7 @@ class SpoolManager:
         else:
             return {
                 "response": response.json(),
+                "response_headers": dict(response.headers.items()),
                 "error": None
             }
 


Arksine:
I can see how this would be useful.  I'm wondering if we should convert the full headers object to a dict and return them all.  Are we sure we only need the extended headers?  The entire headers object may not be necessary, but I don't think it would hurt to return them all either.

Also, when ready to merge, don't forget to add your [signed off line](https://moonraker.readthedocs.io/en/latest/contributing/#git-commit-format).  Thanks.

Clon1998:
> I can see how this would be useful.  I'm wondering if we should convert the full headers object to a dict and return them all.  Are we sure we only need the extended headers?  The entire headers object may not be necessary, but I don't think it would hurt to return them all either.
> 
> 
> 
> Also, when ready to merge, don't forget to add your [signed off line](https://moonraker.readthedocs.io/en/latest/contributing/#git-commit-format).  Thanks.

For now I just went with all custom headers, but as you mentioned including all headers in the result won't hurt either.
I'll adjust the change and also adapt the PR within your guidelines. Sorry about that.

Arksine:
Thanks.

Issue #842, opened 2024-04-22
How to add Authenticator


SlimRG:
### Is your feature request related to a problem? Please describe

I have static ip (https://slimhome.tech/#/login) and fluiid.
So, my pass was brootforced by me by 45 days.
It's a secure problem, so I want to fix it.

### Describe the solution you'd like

After entering login/pass show page to write google auth pin or with QR code to create it if no key inserted before.

### Additional information

https://www.rfc-editor.org/rfc/rfc6238
https://github.com/googleapis/google-auth-library-nodejs

SlimRG:
Adding a little, but I don't know more about your code.
I know python only a little(
But you can add

```
    def __init__(self, config: ConfigHelper) -> None:
        """
        Initialize with a storage mechanism for user-specific secrets.
        user_secret_storage should be a dictionary or a database that maps usernames to secret keys.
        """
        self.user_secret_storage = user_secret_storage

    def generate_secret(self) -> str:
        """
        Generate a new secret key that can be saved for a user.
        Typically, this would be done during the setup process of 2FA.
        """
        return pyotp.random_base32()

    def get_uri_for_qr(self, username: str, issuer_name: str) -> str:
        """
        Generate a URI that can be encoded into a QR code and scanned by Google Authenticator.
        issuer_name is the name of the organization providing the 2FA.
        """
        secret = self.user_secret_storage.get(username)
        if not secret:
            raise ValueError("User does not have a secret key set up.")
        return pyotp.TOTP(secret).provisioning_uri(username, issuer_name=issuer_name)

    def verify_code(self, username: str, code: str) -> bool:
        """
        Verify the TOTP code entered by the user.
        Returns True if the code is correct, False otherwise.
        """
        secret = self.user_secret_storage.get(username)
        if not secret:
            raise self.server.error("User does not have a secret key set up.")
        
        totp = pyotp.TOTP(secret)
        return totp.verify(code)
```

SlimRG:
https://github.com/Arksine/moonraker/pull/844

PR #844, opened 2024-04-23
Add TOTP Authorization to moonraker auth. algo

SlimRG:
https://github.com/Arksine/moonraker/issues/842


SlimRG:
Sorry, git system is new for me...

Arksine:
Thanks.  At this time I don't think support for 2FA is a good candidate for this repo.  Moonraker is not intended to run facing the wan directly.   It may be something I consider in the future, but I think I would prefer adding support for OAuth and allowing the provider to handle 2FA.

If you need to add additional security  and cannot use a VPN I would recommend looking into authentication options provided by reverse proxies, etc.

SlimRG:
We can't use VPN in many countries, such as Russia (if you don't have licence), if we don't want to have problems with law (I had this problems by local WireGuard).
Google Authenticator is used in banking services (I know, as I work there). It's quiet simple and secure way to improve security. 
I have in use some printers and want to connect them from office next to my home. Why you don't want to add this secure layer?

Arksine:
> We can't use VPN in many countries, such as Russia (if you don't have licence), if we don't want to have problems with law (I had this problems by local WireGuard).

I understand.   This is why I suggested [using a reverse proxy](https://stackoverflow.com/questions/72434853/ways-of-authentication-with-nginx)  to provide additional authentication if you want expose Moonraker to the wan.

> I have in use some printers and want to connect them from office next to my home. Why you don't want to add this secure layer?

At this time I don't think 2FA adds significant value to Moonraker.   Moonraker is not  intended to run directly facing the wan.  The overwhelming majority of users will not want to enable 2FA to their local installation, in fact the majority do not even have user authentication enabled.    I have not reviewed your submission in detail, however a cursory glance indicates that it would need significant refactoring before its eligible for merging.   The [contribution guidelines](https://moonraker.readthedocs.io/en/latest/contributing/) need to be followed before I provide a detailed review.

In addition, I think it may be possible to add OpenID/OAuth support which would accomplish a similar goal, with the bonus of allowing users to log in with existing accounts.

> P.S. Localhost is not secure enough too, as lots of people use phone number as wifi pass. Also, MGTS provider sets routers with default pass MGTS98765 and without possiblility to change this password.

Users who don't secure their local network and/or routers would be unlikely to user authentication, much less two factor authentication.  They would have bigger vulnerabilities than anything exposed by Moonraker.

>  As I think, 2FA additional secure layer will be better and simpler way, than using LDAP.

LDAP is already integrated and well tested.

PR #846, opened 2024-04-29
Implemented control for Tuya smart plugs

fer662:
I bought a Tuya smart plug to make my printer power off after prints if left unused but didn't find support for it, so I added it.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -949,6 +949,39 @@ type: tplink_smartplug
 address: 192.168.1.123
 ```
 
+#### Tuya Smartplug Configuration
+
+You need to follow the steps described in [the tinytuya github repository](https://github.com/jasonacox/tinytuya)
+to set up a Tuya developer account and obtain the information required to configura a tuya smart plug
+
+The following options are available for `tuya_smartplug` device types:
+
+```ini
+# moonraker.conf
+
+device_id:
+# Unique identifier for the Tuya device
+address:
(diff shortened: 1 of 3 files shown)


Arksine:
Thanks.  Unfortunately the `tinytuya` module is not compatible with Moonraker, as it includes blocking code and attempts to install the `cryptography` dependency which breaks some installations.  See the [contributing documentation](https://moonraker.readthedocs.io/en/latest/contributing/) for additional details.

Generally speaking I prefer to avoid adding new http integrations directly in the `power` component, as there are simply too many devices, manufacturers, and protocols.  Instead I recommend using the [generic http](https://moonraker.readthedocs.io/en/latest/configuration/#generic-http-devices) when possible.  This may not be possible with a Tuya plug, in which case third party software like Home Assistant may be an alternative.  A Tuya implementation in Moonraker would likely need to be done from scratch, with non-blocking code and without depending on modules like cryptography that cannot be built on low-resource SBCs.

fer662:
I understand what you mean, but this works fine for my use case, which is basically the printer commits suicide and turns itself off (including the raspberry) a few mins after getting cold enough. It's not too important to me if the blocking calls are handled properly at that point. Thought i'd leave a reference to my fork in case that's what someone else needs, even if it doesn't reach the standards for merging. The plugs that are supported are much harder to find in many places.

Arksine:
Thanks.  I'm going to go ahead and close this PR since it cannot be merged, however its reference is still available in the feature request, so users should be able to find it.

PR #870, opened 2024-06-17
spoolman: add support for aditional headers 

LobodPl:
Adds additional_headers parameter, to define extra headers passed with all spoolman requests.

Use case: 
Spoolman has no auth and api is all public. 
You use your printers in different locations, so you put Spoolman behind ZeroTrust (CloudFlare for example) to use only 1 instance.
You will need to pass extradata along with all spoolman requests to get through ZeroTrust Gateway.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -3141,6 +3141,9 @@ server: http://192.168.0.123:7912
 sync_rate: 5
 #   The interval, in seconds, between sync requests with the
 #   Spoolman server.  The default is 5.
+additional_headers:
+#   Additional headers for spoolman connection. Usefull for connecting to for example CloudFlare Zero Trust.
+#   Dict "name=value" list.  The default is empty.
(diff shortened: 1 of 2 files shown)


Arksine:
Thanks.  I'm a bit weary of adding this functionality directly to Moonraker.   It seems like it would be possible to proxy requests to Spoolman and add any additional necessary headers there, which would be the preferable way of handling unique setups that don't apply to the majority of users.

LobodPl:
> Thanks. I'm a bit weary of adding this functionality directly to Moonraker. It seems like it would be possible to proxy requests to Spoolman and add any additional necessary headers there, which would be the preferable way of handling unique setups that don't apply to the majority of users.

Hello, good point of view. For me my change is as rare as use of the spoolman itself, but right now I'm also sceptic if it should be a part of moonraker itself.

PR #885, opened 2024-07-12
Read MQTT clientId from config 'client_id' param, with default value …

antlu65:
Read MQTT clientId configuration option from moonraker.conf. Default value is socket hostname.


Arksine:
Support for a configured MQTT `client_id` was added in commit b8ba6c0d2ba4f686c9dbf1e7469fd2988e248123.  Thanks

PR #888, opened 2024-07-22
Add support for 'png begin' and 'png end'

geekykayaker-anon:
I have an Ender3 V3 KE, with testing I've found that the display will show thumbnails if the text "png begin" is used, but NOT if the text "thumbnail begin" is used.

I've changed my slicer to use "png begin" but now moonraker can't detect the images...

Can we add support for both specification of png thumbnails?

### DIFF
--- a/moonraker/components/file_manager/metadata.py
+++ b/moonraker/components/file_manager/metadata.py
@@ -208,7 +208,7 @@ class BaseSlicer(object):
     def parse_thumbnails(self) -> Optional[List[Dict[str, Any]]]:
         for data in [self.header_data, self.footer_data]:
             thumb_matches: List[str] = re.findall(
-                r"; thumbnail begin[;/\+=\w\s]+?; thumbnail end", data)
+                r"; (thumbnail|png) begin[;/\+=\w\s]+?; (thumbnail|png) end", data)
             if thumb_matches:
                 break
         else:


Arksine:
Hi.  Moonraker supports the standard thumbnail format generated by PrusaSlicer and its derivatives.  These thumbnails are already PNGs.  It sounds like the software driving the display is deviating from the standard and should be corrected to use the normal thumbnail identifier for PNGs.

Further, Creality is not in compliance with Moonraker's GPLv3 license, along with other software in the Klipper ecosystem including Klipper itself.  Until they correct this, my policy to to avoid adding support specific for Creality hardware and/or software to Moonraker.

geekykayaker-anon:
Yeah, I completely understand where you're coming from with this, if Creality aren't following the standard and are non GPLv3 compliant, it seems reasonable to not support them.

That said, it sounds like in principle you're happy to follow the standard, but I'm having trouble finding out what that actually is here...

> The standard thumbnail format generated by PrusaSlicer and its derivatives. 
Do you happen to know if this standard is documented somewhere?

If it's just "What slicers support" then:
SuperSlicer has this Closed MR - https://github.com/supermerill/SuperSlicer/issues/3989
Creality has this MR (that I opened) - https://github.com/Ultimaker/Cura/pull/19419
I can look at adding to PrusaSlicer too if it would help?

It kind of depends if your view is "as the majority of slicers don't support this, we wont" or "no matter how many slicers support this, we're not supporting it until Creality are GPLv3 compliant"
(either would be reasonable)

geekykayaker-anon:
Kind of unrelated to this MR, but as it's come up...

Do you have a statement/document somewhere about how Creality is not in compliance with Moonraker's GPLv3?

I don't know the details (which is why I'm asking) but a cursory look shows that Creality does have a bunch of code published with GPLv3 licences.

Arksine:
> Do you happen to know if this standard is documented somewhere?

I'm unaware of a "documented" standard, there is no RFC for it or anything.  Prusa introduced this method of inlining thumbnails in gcode.  As such, its reasonable to consider their method the defacto standard as the other implementations are derived from it.

> Do you have a statement/document somewhere about how Creality is not in compliance with Moonraker's GPLv3?

Moonraker's source for the K1 is nowhere to be found on Creality's GitHub.  There is some old source for the Sonic Pad, however no source for the latest release on Jan 24 2024 has been supplied.

They have released some source for Klipper, however it is not the complete source.  For example, there is no source for the `pr touch`, instead there are [binaries](https://github.com/CrealityOfficial/K1_Series_Klipper/blob/main/klippy/extras/prtouch_v2.py) loaded by a wrapper in Python.   I don't know the details with regard to Mainsail and KlipperScreen, but it is my understanding that Creality is infringing on those projects as well.

Creality has a history of infringing on the GPL for multiple projects going back to Marlin.  For the purposes of  adding any kind of specific support for Creality software/hardware in this repo the burden is on them to prove they are on compliance.   They will need to comply with the licensing for every project in the Klipper ecosystem and do so over a significant period of time to show that they will remain in compliance.

geekykayaker-anon:
Thanks for this, appreciate the details.

I'm happy to close if you want, I can re-open if/when Prusaslicer chooses to add support for this.

PR #901, opened 2024-08-28
Don't use the global `data` directory

FFY00:
Putting data in the global `data` directory is not recommended, a better place for data generally is to install it as a subpackage. This PR makes that change.

Additionally, it's unclear to me how the `moonraker` config in update manager isn't broken, since `requirements` and `system_dependencies` specify set the base path to the Python package (`moonraker.utils.source_info.source_path`) but the files aren't installed there.

https://github.com/Arksine/moonraker/blob/f735c0419444848b59342a98ad3532eef123ea46/moonraker/components/update_manager/common.py#L24-L43


Arksine:
> Putting data in the global data directory is not recommended, a better place for data generally is to install it as a subpackage. 

Hi.  I must ask, not recommended by whom?  While I agree with this recommendation when it comes to data used by the package itself, this does not apply to the items in Moonraker's scripts path.   They are external helpers, primarily used to help install Moonraker on a system.  In my view the shared data directory is a viable location for these items.   Moving the scripts path in the repo is not an option, as it would break existing workflows that install Moonraker from source.

> Additionally, it's unclear to me how the moonraker config in update manager isn't broken, since requirements and system_dependencies specify set the base path to the Python package (moonraker.utils.source_info.source_path) but the files aren't installed there.

The `path` option is not used when updating python packages.

FFY00:
See https://setuptools.pypa.io/en/latest/userguide/datafiles.html#non-package-data-files.

The shared data directory is leftover from old packaging tooling, and has a set of issues. It's difficult to support and is likely to be deprecated in [sysconfig](https://docs.python.org/3/library/sysconfig.html) at some point in the future, and will likely be missing from newer APIs (source: I am the maintainer :sweat_smile:). Installation tools should still be able to install the files, preserving backwards compatibility, but we'd recommend against its usage.

> Moving the scripts path in the repo is not an option, as it would break existing workflows that install Moonraker from source.

I included a symlink in the PR for this reason.

Arksine:
Thanks for the information.  I'll accept your recommendation, however I would prefer to copy the scripts into the distribution during the build rather than move the `scripts` folder and symlink to it.  This way I can exclude items that do not need to be distributed in the package.  In addition the install script will need an update to correctly locate the scripts folder when Moonraker is installed as a package.  I will create a superseding PR to accomplish this.

FWIW, if the intent is to deprecate the use of the data directory, I would recommend coordinating with PyPA.  According to the [Python Packaging User Guide](https://packaging.python.org/en/latest/specifications/binary-distribution-format/#the-data-directory) the data folder is the correct location for scripts and such.  The recommendation from setuptools seems to apply to files accessed at runtime by the package, which makes sense but does not apply here.  The `update_manager` only accesses items in the `scripts` path when Moonraker is installed from source.

Arksine:
Commit 8df98908ca36f673dcf571997c1e33fd6a4e9ea6 copies selected items from the `scripts` folder into the package during the build.   Since this change required an update to the installer I published a patch release of Moonraker, `v0.9.3`.

PR #903, opened 2024-09-01
Update metadata.py

dobrzann:
Thumbnails support for Creality Print 5.0


Arksine:
Thanks.  My policy is to avoid merging code supporting Creality printers and/or software until they comply with the GPLv3 for all software in the Klipper ecosystem, including Moonraker itself.   I understand this is disappointing for Creality owners that just want to improve the software on their printers.  Unfortunately legal action isn't a viable option, so this is the only small measure I can take in hopes that the feel some pressure to comply with open source licensing.

thijstriemstra:
> My policy is to avoid merging code supporting Creality printers and/or software until they comply with the GPLv3

Excellent!!

Issue #481, opened 2022-08-09
[FR] Pipe through Klipper Estimator


Sineos:
### Is your feature request related to a problem? Please describe

Currently the print time guesses are just that - guesses based on slicer information and (I assume) layer times of the previous layers.

### Describe the solution you'd like

Pipe the gcode transparently through [Klipper Estimator](https://github.com/Annex-Engineering/klipper_estimator) to return accurate print time forecasts to the clients

### Describe alternatives you've considered

Leave it as is. Anyway takes as long as it takes 😉

gauravmm:
I'm happy to do this as a moonraker plugin -- it will fit nicely with the  gcode server plugin I was writing.

Dreistein75:
Is there a chance this can be implemented like the EXCLUDE_OBJECT post-processing? That would be very neat

gauravmm:
The current idea is that this is dynamically calculated (and cached) by a moonraker plugin, so no slicers need to be involved, it can pull the latest Klipper configuration information automatically, and it can recompute the estimate when the Klipper configuration changes.

lbibass:
Bump! Any updates on this?

gauravmm:
I'm almost done writing my PhD dissertation, I'll start work on this in mid-Feb. Sorry about the delay!

lbibass:
> I'm almost done writing my PhD dissertation, I'll ~start~ resume work on this in mid-Feb. Sorry about the delay!

No worries! Congrats on nearly being finished with your PhD! Good luck!

richfelker:
Ping. Any progress?

clinta:
Using klipper estimater as a post processing script in my slicer results in very accurate estimates if I look at the slicer time remaining in mainsail. Makes me wish I could just disable the built in time remaining, cause it's basically pointless.

nelsongraca:
I have written a plugin to run gcode on upload through klipper estimator, still a little rough but it's working.

@Arksine would you be willing to add this in or would you prefer for it to be a plugin?

CharlemagneLasse:
@Arksine this would improve my life so much. Just happens to often that I go to the printer (on the incorrectly estimated time by the slicer) just to figure out that I could have stayed an hour longer at home instead of waiting for the 3D printer in the workshop to finally finish. I don't have any control about the slicer because I am often not the person who sliced and started the print.

@nelsongraca I am not sure how this is handled here but maybe having some RFC/Draft to discuss this further would help

[...]

PR #914, opened 2024-10-22
Add klipper_estimator support

nelsongraca:
Added the ability to pipe gcode through klipper_estimator.

Related feature request: https://github.com/Arksine/moonraker/issues/481

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -3275,6 +3275,24 @@ With the above configuration it is possible to run the `SET_ACTIVE_SPOOL ID=1`
 command to set the currently tracked spool ID to `1`, and the `CLEAR_ACTIVE_SPOOL`
 to clear spool tracking (useful when unloading filament for example).
 
+
+### `[klipper_estimator]`
+
+Enables gcode processing through [klipper_estimator](https://github.com/Annex-Engineering/klipper_estimator)
+to update gcode estimated time.
+
+```ini
+# moonraker.conf
+
+[klipper-estimator]
+executable_path: /tmp/klipper-estimator
+#Where klipper-estimator executable is
+download: False
+#If set to true, it will download the latest version of the executable
+asset_name: klipper_estimator_rpi
+#The asset name to download for your installation, see the list on the releases page: https://github.com/Annex-Engineering/klipper_estimator/releases
+```
+
 ## Include directives
 
 It is possible to include configuration from other files via include
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -3284,7 +3284,7 @@ to update gcode estimated time.
 ```ini
 # moonraker.conf
 
-[klipper-estimator]
+[klipper_estimator]
 executable_path: /tmp/klipper-estimator
 #Where klipper-estimator executable is
 download: False
--- a/moonraker/components/klipper_estimator.py
+++ b/moonraker/components/klipper_estimator.py
@@ -1,0 +1,101 @@
+# KlipperEstimator - Run uploaded gcode files through klipper_estimator
+#
+# Copyright (C) 2024 Nelson Graça <graca.nelson@gmail.com>
+#
+# This file may be distributed under the terms of the GNU GPLv3 license
+
+import asyncio
+import logging
+import os
+import os.path
+import stat
+from typing import cast
+from moonraker.components.http_client import HttpClient
+
+
+class KlipperEstimator:
+    def __init__(self, config) -> None:
+        self.server = config.get_server()
+        self.name = config.get_name()
+        self.ke_exec: str = config.get("executable_path", "/tmp/klipper-estimator")
+        self.download = config.getboolean("download", False)
+        self.tag = config.get("tag", "latest")
+        if self.tag != "latest":
+            self.tag = f"tags/{self.tag}"
+        self.asset_name = config.get("asset_name", "klipper_estimator_rpi")
+        self.file_manager = self.server.lookup_component('file_manager')
+        self.gc_path = self.file_manager.get_directory()
+        hostname = self.server.get_host_info()["hostname"]
+        port = self.server.get_host_info()["port"]
+        self.url = f"http://{hostname}:{port}/"
+        self.server.register_event_handler(
+            "file_manager:filelist_changed",
+            self._handle_filelist_changed
+        )
+        if self.download:
+            asyncio.create_task(self._download())
+
+    async def _download(self) -> None:
+        client: HttpClient = self.server.lookup_component("http_client")
+        response = await client.github_api_request(
+            f"repos/Annex-Engineering/klipper_estimator/releases/{self.tag}")
+        response.raise_for_status(
+            "Failed to get latest Klipper Estimator info.")
+        data = cast(dict, response.json())
+        tag_name = data["tag_name"]
+        try:
+            current_version = open(f"{self.ke_exec}.version", "r").read()
+        except Exception as e:
+            logging.info("Cant determine current version")
+            logging.info(e)
+            current_version = "9999"
+        if current_version != tag_name or not os.path.isfile(self.ke_exec):
+            for asset in data["assets"]:
+                if asset["name"] == self.asset_name:
+                    down_url = asset["browser_download_url"]
+                    content_type = asset["content_type"]
+                    logging.info(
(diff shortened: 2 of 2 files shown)


docs/configuration.md
```diff
@@ -3275,6 +3275,24 @@ With the above configuration it is possible to run the `SET_ACTIVE_SPOOL ID=1`
...
+
+### `[klipper_estimator]`
+
+Enables gcode processing through [klipper_estimator](https://github.com/Annex-Engineering/klipper_estimator)
+to update gcode estimated time.
+
+```ini
+# moonraker.conf
+
+[klipper-estimator]
```
CharlemagneLasse, line 3287:
```suggestion
[klipper_estimator]
```

The rest seems to work like a charm. Thank you

nelsongraca: missed this one, thanks

nelsongraca:
sorry for the ping @Arksine but is there any chance this will be merged?

moonraker/components/klipper_estimator.py
```diff
@@ -0,0 +1,101 @@
...
+                logging.info(f"Running klipper-estimator in {path}")
+                await self._run_estimator(full_path)
+
+    async def _run_estimator(self, path) -> None:
+        path = path.replace("\"", "\\\"")
+        cmd = " ".join(
+            [self.ke_exec, "--config_moonraker_url", self.url, "post-process",
+             f"\"{path}\""])
+        logging.info(f"Running {cmd}")
+        timeout = 10.
```
CharlemagneLasse, line 90: During my experimentations, I've noticed that the rpi4 is often not able to process the gcode files in 10 seconds. I've now increased it for me to 120s inside the code. Not sure if it would be better to make it configurable from the outside.

nelsongraca: I think it's worth adding as a config, I'll add it.

Arksine:
Thanks, sorry for the delay in responding, time has been somewhat constrained until recently.

I had started some work on this and recently finished it, which is now available in the [analysis](https://moonraker.readthedocs.io/en/latest/configuration/#analysis) module.  It takes a different approach as it doesn't apply a post-process to the file.  The primary issue with a post-process approach that triggers on a `filelist_changed` event is that it can potentially modify the file while Klipper is attempting to print.

With the `[analysis]` module Klipper Estimator will automatically perform an analysis as a post-metadata extraction step when `enable_auto_analysis` is set to true in the configuration.  The `estimated_time` field in the metadata will be replaced with the `total_time` returned by Klipper Estimator.  The `analysis` module also adds a few more features:

- It attempts to auto detect the platform.  If autodetection fails the user can configure it. 
- The initial download places the binary in the `<data_folder>/tools/klipper_estimator` directory.
- There is an option to register `klipper_estimator` with the `update_manager` to handle updating.
- There are API endpoints to perform analysis and dump the config
- Klipper Estimator uses a config file rather than Moonraker's URL for peforming estimates.  The user can specify the KE config file, however the default is to use a dumped configuration.  This makes Klipper Estimator available even if Klippy is not connected to Moonraker.  Moonraker will dump the default config if it doesn't exist and can be configured to dump it every time Klipper reports ready.

freakydude:
@Arksine I read your changeset and configured it with
```
[analysis]
enable_auto_analysis : true
```

The `estimated_time' field now differs from the estimated time in the slicer. So it looks to me like this is working fine so far. 
I used the KlipperEstimator before by adding it as a post-progress step in the slicer. It did not change the metadata but updated the `M73` gcode commands within the file.

Does the current solution do this? And if not, is it planned?

The reason I ask is that frontends like KlipperScreen calculate the remaining time with one of your methods -> https://moonraker.readthedocs.io/en/latest/external_api/introduction/?h=progress#basic-print-status. And if they use `display_status.progress` it will be different and not as accurate as KlipperEstimator.

Thanks for your work 
Kind regards

CharlemagneLasse:
> @freakydude That would require a post process. I think it would be better to perform it in the slicer. At this time I don’t have plans to add support for post-processing in Moonraker as there are potential issues with inotify.

But aren't you already doing postprocessing for ` preprocess_cancellation`? After I tried to switch from nelsongraca's working implementation to the official one from you, I noticed various problems. Not sure how they are related:

* klipperscreen time is wrong (maybe because m73 are no longer updated)
* fluidd is no longer showing the slicer time during print
* in fluidd, i no longer have the possibility to press "load current file" when opening the webpage/dashboard when a print was started 

Of course, I needed to update moonraker and co to get this builtin analysis code. It can therefore be that some other non-analysis change broke these things

Arksine:
> But aren't you already doing postprocessing for  preprocess_cancellation? After I tried to switch from nelsongraca's working implementation to the official one from you, I noticed various problems. Not sure how they are related:

The processing done for `preprocess_cancellation` occurs during the metadata processing step, before the file is moved to folders watched by inotify.    It was also meant as a temporary measure to support a critical feature in Klipper.   Most modern slicers now natively support Klipper objects.   

I don't use KlipperScreen, but I'll test to see if I can repro some of your issues in Fluidd.  Some of what you report doesn't sound related.

Arksine:
I did decide to change the behavior of the `enable_auto_analysis` option.  It will now result in a `post-process` just before metadata is extracted from the file, presuming that the file hasn't been previously processed by Klipper Estimator.

@pedrolamas @meteyou  This change includes a new metadata field, `file_processors` (the [specification](https://moonraker.readthedocs.io/en/latest/external_api/file_manager/#gcode-metadata-spec) has been updated to refect this).  This field will contain an array of application names that have processed the file.  Right now it can include `klipper_estimator` and/or `preprocess_cancellation`.   The `[analysis]` section has to be present in the config for `metadata.py` to identify if a file has been processed by Klipper Estimator, but `enable_auto_analysis` does not have to be set to `true`.  This makes it possible for Moonraker identify files for users who prefer to use Klipper Estimator manually or in a slicer.

I chose this method in case there are further post-processing apps that should be added in the future.   The implementation is generic, the component hosting the app (ie: `[analysis]`) registers a configuration that is passed to `metadata.py` each time its invoked.  The configuration contains the path to the executable, its arguments, its name, the regex used to ID it, etc.

I also added a [an endpoint](https://moonraker.readthedocs.io/en/latest/external_api/integrations/#post-process-a-file) that allows for a manual post-process.

freakydude:
@Arksine Sounds great – I'll check this out soon. Thanks for reconsidering.

pedrolamas:
> * fluidd is no longer showing the slicer time during print
> * in fluidd, i no longer have the possibility to press "load current file" when opening the webpage/dashboard when a print was started

Unrelated, but for the record, that was probably my fault in a botched Fluidd code refactor I did, and not related with the Moonraker changes here...

It should be fixed on Fluidd v1.32.2 though, so please do ensure you update to that version.

pedrolamas:
@Arksine I've now implemented in the latest moonraker `[analysis]` changes in Fluidd and everything seems to be working fine!

CharlemagneLasse:
> > * fluidd is no longer showing the slicer time during print
> > * in fluidd, i no longer have the possibility to press "load current file" when opening the webpage/dashboard when a print was started
> 
> Unrelated, but for the record, that was probably my fault in a botched Fluidd code refactor I did, and not related with the Moonraker changes here...
> 
> It should be fixed on Fluidd v1.32.2 though, so please do ensure you update to that version.

I have  v1.32.2  according to the "Software Updates" page. But still, I see the problem. But there are a couple of things which could be related:

* I have my printjobs in subfolders
* print from subfolder is runnig
* "load current file" button is not available
* opening the folder (under jobs) with the current printing job makes following available:
** "load current file" for the gcode preview
** "Slicer" time in the "Printing" component

Arksine:
@freakydude Moonraker's automated post-process just executes `klipper_estimator`.  If Klipper Estimator processes macros then they will be included.

The discrepancy might be from post-processing the same file twice.  It might be a simple rounding error.  I haven't looked into how Klipper Estimator calculates its estimates, its possible that different runs always produce slightly different results.

Arksine:
I'm going to go ahead and close this since its implemented in `analysis` and discussion seems complete.  Thanks.

nelsongraca:
A little late but time has been limited here, thank you for implementing it in a better way that I could ever do python is not my strength at all.

PR #923, opened 2024-12-05
Update web_api.md

HawkerLin:
the request type is POST in file_manager.py
```python
self.server.register_endpoint(
            "/server/files/metascan", RequestType.POST, self._handle_metascan_request
        )
```


Arksine:
Thanks.  This should be resolved in the API Documentation rework that was recently pushed.

PR #947, opened 2025-01-30
Add Qidi Plus 4

liberodark:
The Goal of this PR is to try to upstream change for Qidi Plus 4.
For add Support of this printer.

More info is coming.

### DIFF
--- a/moonraker/components/authorization.py
+++ b/moonraker/components/authorization.py
@@ -254,6 +254,11 @@ class Authorization:
             "/access/user/password", RequestType.POST, self._handle_password_reset,
             transports=TransportType.HTTP | TransportType.WEBSOCKET
         )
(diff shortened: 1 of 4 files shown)


Neko-vecter:
You need to explain a bit more about what exactly this PR is for. “No description provided” is not acceptable.

Things need include:
- Function of this PR / what this PR implement?
- What change?
- Demo of the change

Neko-vecter:
> The Goal of this PR is to try to upstream change for Qidi Plus 4. For add Support of this printer.

I don't think the printer specific function PR in to the main repo (mainstream) is a good idea. Since only that printer using this function. 
It should be implement in your own repo (fork).

liberodark:
Yep but probably a part would be upstream.

Issue #877, opened 2024-06-24
[Spoolman] Support fetching filament information from spoolman


tbe:
### Is your feature request related to a problem? Please describe

I print some parts repeatedly with whatever filament i have at hand. Currently, i have to re slice every time with PA, Bed Temp and Heater temps.

### Describe the solution you'd like

Spoolman supports `settings_extruder_temp` and `settings_bed_temp` nativly. Also, there is support for custom attributes (`extra`).

Fetching those informations and storing them as saved variables would allow access to them in the GCODE. This way, gcode files, saved on the virtual sd card, can be reused without the need of changing those files, editing configs, or re slicing  everything. 

### Describe alternatives you've considered

Currently, i switch include files, one for each filament type by hand. This process is error prone.

Issue #702, opened 2023-08-15
Get Spoolman active Spool ID inside Klipper


lettore:
### Is your feature request related to a problem? Please describe

It's possible to expose to Klipper as a printer variable the Spoolman spool ID currently active in Moonraker?

### Describe the solution you'd like

From any G-Code Macro we should be able to load current loaded filament data via printer.spoolman.id, printer.spoolman.material, printer.spoolman.extruder_temp etc.
This will be amazing because in any macro you can get the currently loaded filament and adjust the parameters of the print accordingly.

rufinus:
this should be easy a `action_call_remote_method` to the spoolman api would be enough.

Terror-Gene:
First result for "klipper macro active spool". have searched around & cant find a solution.
I would like to add if statements somewhere in my LOAD_FILAMENT / RESUME / START_PROCEDURE macros that check if "active spool = none" & warns to select a spool first.
Since i have it set so UNLOAD_FILAMENT clears the active spool. this feature would save me ever accidentally using a spool without tracking it. Plus the benefit of printing (or loading filament) using the correct temperature for the filament inserted, without needing to re-slice gcode. could also configure bed temp to be set to the layer1 filaments preferred bed temp.
(ideally it would prompt with the spoolman integration to select a filament, but i dont know how one would achieve that with both webui & klipperscreen simultaneously)

evgarthub:
I would like to get active spool id at least to achieve full workflow:
1. remove old spool => unload macro clears current active spool
2. load new spool => prompt for spool (optional and separate feature from current one)
3. slice
4. upload and print
5. START_PRINT macro can be adjusted to make sure `GET /server/spoolman/spool_id` is set to something otherwise PAUSE print

I checked the current code, it looks like we need to register additional action to get the id from api in gcode
https://github.com/Arksine/moonraker/blob/3008a13efb6045a007456cb4f34a64afa6140617/moonraker/components/spoolman.py#L54

Not sure if it's technically possible to make async request with `action_call_remote_method` and use the response of it in macro...

Arksine:
As of right now Klipper does not have the ability to request data from its clients (ie: Moonraker).  The `action_call_remote_method` 
 is fire and forget, it can send data to Moonraker but it cannot receive a response.

Klipper will need to add support for requesting data from its API Server clients before this feature request can be fulfilled.

evgarthub:
Thank you! Good to know! 
[...]

PR #948, opened 2025-01-31
Send Spoolman spool info to klipper

bofh69:
This is a fix for  #702 and #877

## Description

I have a very shallow knowledge of python, klipper's gcode templates and the moonraker codebase so don't hesitate to suggest better ways I can do this.

This changes the spoolman integration so that when the spool id is changed, moonraker will lookup
the new spool's data from spoolman. For every field in the spool's data, it checks if there is a gcode macro
in Klipoper with the name `_SPOOLMAN_SET_FIELD_<fieldname>`. If there is, it will be called
 with `VALUE=<value>` as argument.

For nested fields, the field names are separated by underscore, ie it can call:
`_SPOOLMAN_SET_FIELD_filament_vendor_name VALUE="Funky Filament Factory"`

When the spool id is cleared, it will instead call `_SPOOLMAN_CLEAR_FIELDS`, if available.

## Testing

I've tested it on my own printer because I want to store the filemant.id in a variable.
That works fine. I've tested without these gcode macros, with just one for setting or clearing and with both macros for setting and clearing the fields.

## Backwards compatibility

This should not affect old installs, if they don't configure any `_SPOOLMAN_*` macros, nothing changes.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -3287,6 +3287,22 @@ With the above configuration it is possible to run the `SET_ACTIVE_SPOOL ID=1`
 command to set the currently tracked spool ID to `1`, and the `CLEAR_ACTIVE_SPOOL`
 to clear spool tracking (useful when unloading filament for example).
 
+#### Sending spool info to Klipper
+
+When changing the spool id, spoolman will fetch info about the spool
+from Spoolman and it can send that info to Klipper.
+
+Moonraker does that by running gcode like this:
+"`_SPOOLMAN_SET_FIELD_name VALUE=val`".
+where name is the field's name. For nestled fields, the names are separated by '_'.
+
+Ie if the spool has a filament id of 2, it will run:
+`_SPOOLMAN_SET_FIELD_filament_id VALUE=2`.
+
+When the spool id is cleared, it will run:
+`_SPOOLMAN_CLEAR_FIELDS`.
+
+
 ## Include directives
 
 It is possible to include configuration from other files via include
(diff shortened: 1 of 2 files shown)


Arksine:
Thanks.  As I mentioned briefly in issue #702, I don't think the appropriate way of getting information from Moonraker to Klipper is to hack it in via macros.  They don't have strictly defined behavior and it isn't possible for Klipper to know if the data received is up to date, if an error is encountered, etc.     In addition there may be other data that users would like to  access from Klipper (such as its gcode metadata), and this isn't a pattern I want to maintain throughout various parts of Moonraker.  In summary, I don't think the potential maintenance burden is justified.

Its my view that the linked issues are feature requests for *Klipper*, not Moonraker, as the functionality desired is to access this data in Klipper.  As things stand today Klipper has no way of requesting data from its clients.  There are two ways this could be implemented:
1) Extend Klipper's API server to allow Klipper to request data from a client.
2) Add a Klipper extra that connects to Moonraker's unix domain socket, allowing macros to call Moonraker APIs.  IIRC someone may have done some work on this.

In the interim, you do have a couple of options if you don't want to maintain a fork of Moonraker:
1) You could move this functionality to your own component that listens for the `active_spool_set` event and syncs with your macros.
2) It would be possible to write a small agent that does something similar, as `active_spool_set` is also reported as a JSON-RPC notification.

bofh69:
Thanks for the suggestion!

I made an agent of it instead:
https://github.com/bofh69/spool2klipper

PR #961, opened 2025-02-17
docs: fixes invalid JSON in example

pedrolamas:
Fixed the main JSON example that was actually invalid (was missing the containing `{` and `}`)

Reformatted an example for better readability.

Also marked 2 other examples with the new doc tags


pedrolamas:
I do envy your awesome work documenting everything, so kudos for that!

PR #962, opened 2025-02-17
file_manager: Added additional metadata to aid tool selection on single extruder multi-material prints

moggieuk:
file_manager: Added additional metadata to aid tool selection on single extruder multi-material prints

The added metadata is essential for single extruder multi-material printing.  It is used in new Mainsail functionality to map tools to MMU spools when using Happy Hare (Klipper MMU driver)

Signed-off-by: Paul Morgan (moggieuk@hotmail.com)


norris108 (approved):

Arksine:
Thanks.  I'm fine with adding the new  fields, however the metadata specification in the docs needs to be updated with the new fields.

https://github.com/Arksine/moonraker/blob/a0164e157543a91255a02cc9ce4c3848dc3ee694/docs/external_api/file_manager.md?plain=1#L258-L291

I use VSCode's [markdown table](https://marketplace.visualstudio.com/items?itemName=TakumiI.markdowntable) extension to keep tables formatted.  In addition, Moonraker's docs makes use of my [compact tables](https://github.com/Arksine/markdown-compact-tables) extension.  This allows you to merge rows with long descriptions.  When you see a carrot (`^`) at the end of a row it indicates that the row will be merged with the one above it.  It also allows for nested tables with a `+`, but I don't think that will be necessary here.

moggieuk:
Of course.  Still learning my way around Moonraker.  I'll update the docs..

docs/external_api/file_manager.md
```diff
@@ -273,10 +272,15 @@ GET /server/files/metadata?filename=tools/drill.gcode
 | `first_layer_extr_temp` |  float   | The configured first layer extruder temperature, in Celsius. |
 | `first_layer_bed_temp`  |  float   | The configured first layer bed temperature, in Celsius.      |
 | `chamber_temp`          |  float   | The configured chamber temperature, in Celsius.              |
-| `filament_name`         |   str    | The name of the filament used.                               |
-| `filament_type`         |   str    | The type of filament used, ie: `PLA`.                        |
+| `filament_name`         |   str    | The name(s) of the filament used.                            |
```
Arksine, line 275: For consistency with the documentation it would be preferred for these types to be `string` rather than `str`.   The docs are oriented toward JSON (JavaScript) type descriptions.

docs/external_api/file_manager.md
```diff
@@ -273,10 +272,15 @@ GET /server/files/metadata?filename=tools/drill.gcode
...
+| `filament_color`        |   str    | The colors(s) of the filament used in #RRGGBB format.        |
+| `extruder_color`        |   str    | The slicer defined extruder color(s) for the print.          |
+| `filament_temp`         |   str    | The base temperature(s) of filament used, in Celsius.        |
```
Arksine, line 278: I noticed that this is parsed as a string.  Other temperatures parsed by metadata.py are expressed as a float, is there a particuar reason for this not to be?  Is it because there may be multiple?  Since this is a new addition it could be returned as a list of floats.  I'm open to discussion on which is best.

docs/external_api/file_manager.md
```diff
@@ -273,10 +272,15 @@ GET /server/files/metadata?filename=tools/drill.gcode
...
+| `filament_type`         |   str    | The type(s) of filament used, ie: `PLA`.                     |
 | `filament_total`        |  float   | The total length filament used in mm.                        |
 | `filament_weight_total` |  float   | The total weight of filament used in grams.                  |
+| `single_extruder_multi_material` | int | Identifies a multimaterial print with single extruder.   |
```
Arksine, line 282: I wonder if we can shorten the name of this field and make it more descriptive.  It seems this indicates that the print was sliced with MMU enabled, so I wonder if it would make sense to have this field be named `has_mmu` or `has_se_mmu`.

Arksine:
Thanks.  With the documentation I was able to get a better feel for what is being added, so I had a few comments.  I'm open to leaving the `filament_temp` and `single_extruder_multi_material` alone, however I wanted to provide some feedback and open discussion on what could be IMO minor improvements.

moggieuk:
Happy to change the return types and shorten the `single_extruder_multi_material` name - I used that because that is what Prusa calls it and thus thought that seemed to be the precedence elsewhere.

As for types, yes, temp can be parsed an array of floats,  in fact that is what I do in Mainsail amyway.  Again I was really trying to pick up on conventions and missed the individual temps.  I'm back from my trip so let me update..

Arksine:
> One possible concern is that many of the metadata variables used in the slicer can be scalar or an array. E.g. take the existing filament_type. This is currently a string but is really an array of strings, often with one element. Given that I'm sure, for backward compatibility, you don't want to change the meaning of that(?) I think it would make sense to leave the new filament_color, etc as string rather than to chop into string[]. If I do that it will be different from a couple of existing placeholders...

You are correct, we don't want to break API compatibility so `filament_type` should stay as is.   We also have to consider that other slicers may not return multiple values for these fields. 

I'd be fine with returning `filament_color` as a list, lets rename it `filament_colors` if we do that.  That way the nomenclature indicates it will be an array of values, so the deviation from existing `string` fields makes sense.  Likewise, we should probably rename `filament_temp` as `filament_temps`.

moggieuk:
No problem. Let me update this.

moggieuk:
@Arksine I have updated according to your guidance:
Shortened `single_extruder_multi_material` to `mmu_print`

The new metadata: filament_temps, filament_colors, extruder_colors are not all typed-lists.  I added a convenience method to parse using alternative delimiters.  These metadata names are all plural to indicate lists whilst the two original `filament_name` and `filament_type` remain scalar strings.  This at least gives the opportunity to duplicate with "plural lists" (which would be more appropriate) in the future.

Let me know if there is anything else you would like changed.

Arksine:
Looks good, thanks!

moggieuk:
Thank you!

pedrolamas:
I've just started adding support for this in Fluidd, here is a first draft of it: https://github.com/fluidd-core/fluidd/pull/1617

I really dislike how the Filament Name and Type are now showing up (check the screenshot), so will try to do something "cleaver" around that...

moggieuk:
@pedrolamas yes, that did require a little cleanup in the UI ;-)
```
            // This is just a string so split
            const names = file.filament_name ?? ''
            const processedNames = names
                ? names.split(/[,;]/).map((element) => element.trim().replace(/^["']|["']$/g, ''))
                : []
            td.name = processedNames[toolIndex] || 'Unknown'
 ```
 Perhaps there should be a `filament_names` (plural) metadata.  Same with `filament_types`

pedrolamas:
I applied a similar "fix" in Fluidd to be able to show the data like this (though I am only splitting by `;`, not `,` as I never seen it in use on any slicer):

image

PR #965, opened 2025-02-19
updated depencies to run on Mint LInux

whitenhiemer:
Moonraker used to run on LInux Mint, my son ran into issues last night installing it on his workbench.

Updated system dependencies to allow for Linux Mint install

### DIFF
--- a/scripts/system-dependencies.json
+++ b/scripts/system-dependencies.json
@@ -1,5 +1,17 @@
 {
-    "debian": [
+     "debian": [
+        "python3-virtualenv",
+        "python3-dev",
+        "libopenjp2-7",
+        "libsodium-dev",
+        "zlib1g-dev",
+        "libjpeg-dev",
+        "packagekit",
+        "wireless-tools",
+        "curl",
+        "build-essential"
+    ],
+    "linuxmint": [
         "python3-virtualenv",
         "python3-dev",
         "libopenjp2-7",
--- a/scripts/system-dependencies.json
+++ b/scripts/system-dependencies.json
@@ -1,5 +1,5 @@
 {
-     "debian": [
+    "debian": [
         "python3-virtualenv",
         "python3-dev",
         "libopenjp2-7",


Arksine:
This wont fix the install since your don't likely have a hook set up to trigger `sync_dependencies.py`.  The dependencies look identical to Debian, what is the output from `/etc/os-release` on the system running Linux Mint?

whitenhiemer:
> This wont fix the install since your don't likely have a hook set up to trigger `sync_dependencies.py`. The dependencies look identical to Debian, what is the output from `/etc/os-release` on the system running Linux Mint?

Worked last night, I did copy the dependency list from the Debian section.  The out put from os release reads as "linuxmint", should have taken a screenshot.  Mint is debian based, and uses all of their base repos, so I end up treating it as Debian with a mask.  Once I put in the fix, the install problems went away.

Arksine:
What I need to know is what the `ID_LIKE` variable is.  I suspect its ubuntu.  Checking `install_moonraker.sh`, it should just skip the dependency with the requirement specifier.  

Are you using KIAUH to install?  If so, we may need to come up with a fix.  In an ideal scenario  I want to avoid duplicate entries in `system-dependencies.json`.

whitenhiemer:
> Are you using KIAUH to install? If so, we may need to come up with a fix. In an ideal scenario I want to avoid duplicate entries in `system-dependencies.json`.

Yes, I am using KIAUH to do the install of Moonraker

Arksine:
I have done further research and it was a recent change to KIAUH that is resulting in the error.  As I mentioned earlier, I want to avoid duplicate sections in `system_dependencies.json` if at all possible.  The way this works is that distributions that are derivatives report their parents in the `ID_LIKE` field.   Moonraker's `install-moonraker.sh` script will fall back to the first distro specified in `ID_LIKE` that has a definition in `system_depenedencies.json`.  Linux Mint's `ID_LIKE` field contains both `ubuntu` and `debian`.  Ubuntu is first, so the script will will install Ubuntu's dependencies for Linux Mint.

KIAUH has yet to implement this logic, so it fails because `linuxmint` does not have dependencies defined.  I have notified the developer of this issue, however it is up to him if he wants to implement the fallback or not.

Arksine:
This should now be resolved in KIAUH versions 5 and 6.

Issue #932, opened 2025-01-01
POWER control new type: shell script


Jaj0:
### Is your feature request related to a problem? Please describe

[power printer] config has a lot of different types of supported switches, but the issue is that all of them are very device specific.
But there might be cases that all of available solutions are not working for specific case. Then some universal solution would be appropriate.

Example case: im running klipper+moonraker+mainsail on NanoPi Neo Air, my power relay (that shuts down whole printer with MCU) is connected to one of gpios. As this is not RaspberryPI there are some issues to use gpio type for this config.
But i can switch gpio using alternative WiringNP library, just calling gpio write 7 1.

### Describe the solution you'd like

It would be good to add option to have possibility to execute shell command as power on/power off command. Similar as in linked discussion
`[power PRINTER]
type:command
TurnON: gpio mode 7 output; gpio write 7 1
TurnOFF: gpio write 7 0`

### Describe alternatives you've considered

I've tried using shell macros extension for klipper, but issue is that it only works when MCU is already running which is not possible before powering MCU of printer.

Jaj0:
https://github.com/Arksine/moonraker/discussions/784

Twinki14:
+1, I have a TP Link Tapo smart plug between my printer and power supply, but the plug is no longer natively compatible with Moonraker, this tool https://github.com/python-kasa/python-kasa let's me control them from a CLI - using Klippers shell command extensions I have some level of control on my MCU

The CLI/Library for controlling my smart plug could be wrapped in a HTTP server which I've already made an issue about here https://github.com/python-kasa/python-kasa/issues/1403 but if Moonraker had the ability to just execute the CLI, there would be no need

Jaj0:
In the meantime ive created temporary solution in means of simple http server that takes http requests and calls cli commands:
https://github.com/Jaj0/moonraker-cli-switch-server

LittleNyanCat:
+1, I'm using an Arduino connected via USB controlling and ATX PSU and a python script to tell it to turn on/off.

Will probably just implement it myself soon and PR it, doesn't look too hard.

mdm351981:
+1 I’m using a i2c relay to switch power to printer PSU and being executable via [printer power] would tidy things up a lot.

PR #968, opened 2025-02-22
Support for arbitrary power devices via shell commands

LittleNyanCat:
Allows users to implement support for any device they wish themselves, by running a shell command of their choosing.

Addresses https://github.com/Arksine/moonraker/issues/932.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1460,6 +1460,49 @@ location: 1-1
 port: 2
 ```
 
+#### Arbitrary devices via Shell commands
+
+Control any device as long as you can come up with a way to control it with a shell command.
+
+The most practical way of doing it will likely be with a short Python script, although it doesn't have to be.
+
+```ini {title="Moonraker Config Specification"}
+# moonraker.conf
+
+on_cmd:
+#   Command to be ran to switch the device on. This parameter must be provided.
+off_cmd:
+#   Command to be ran to switch the device off. This parameter must be provided.
+timeout:
+#   Number of seconds after which the subprocess spawned by the command will be 
+#   forcefully terminated. It should not need to be increased in most cases. Default is 1 second.
+```
+
+/// Warning
+These commands will not recognize the tilde (`~`) shorthand for your home directory, and as such, you must always use full paths when referring to files.
+
+`python3 /home/<user>/printer_data/psuctrl.py on` will work
+
+`python3 ~/printer_data/psuctrl.py on` will not.
+///
+
+
+Example:
+
+```ini {title="Moonraker Config Example"}
+# moonraker.conf
+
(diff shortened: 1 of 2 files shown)


Arksine:
Thanks.  Unfortunately this can't be merged into mainline, as configurable script execution exposes potential security issues, particularly on installations in public areas like dormitories an labs.

LittleNyanCat:
In such scenarios your installation should already be protected to not allow anyone to interact with your WebUI/Moonraker's API, if it isn't then that's the user's fault. 

Besides, the scripts are fixed, they don't receive arguments from the API, and if you're worried about someone modifying moonraker.conf for arbitrary command execution, I'd argue that you already have way bigger problems if someone managed to get read/write access to your host's filesystem (or just the printer's config files in general).

Arksine:
> In such scenarios your installation should already be protected to not allow anyone to interact with your WebUI/Moonraker's API, if it isn't then that's the user's fault.

Regardless of whose fault it is, my goal is to limit exposure should a malicious actor gain access to the instance of a default installation (no mods).  Currently the worst case scenario is an unauthorized individual can control the printer and reconfigure software.  Bad, but limited.

> Besides, the scripts are fixed, they don't receive arguments from the API, and if you're worried about someone modifying moonraker.conf for arbitrary command execution, I'd argue that you already have way bigger problems if someone managed to get read/write access to your host's filesystem (or just the printer's config files in general).

Moonraker provides r/w access to the `gcodes` and `config` directories, so users can modify the configuration.  What they cannot do is modify the source code for Moonraker, Klipper, or any software registered with Moonraker, even if for some reason the user symlinked it in one of the writable folders.   While it isn't ideal for unauthorized users to be able to reconfigure software, the actions they can currently take do not include running malicious code.  Configurable scripts change that, so I would respectfully disagree that a malicious actor gaining access in the current situation is a bigger problem.

I understand why some users would like this functionality, however it isn't something I am comfortable with in mainline Moonraker.  There are a few other options:
1) A 3rd party `component` that configures the "switch" and registers itself with the `power` component.  I don't endorse or provide support for 3rd party modifications to Moonraker, but I have nothing against them either.  This could require maintenance if `power.py` changes.
2)  A small  http server bound to the localhost that provides endpoints which will run the desired scripts.  Moonraker could then be configured with the generic `http` type.  This is probably the easiest method, I would imagine most people with the skills to write a shell script can figure out how to create a simple server in Python.
3)  Similar to above, but using an MQTT client that subscribes to topics that will execute the script.   This is a step up in difficulty and requires a broker, so not ideal for everyone.
4)  Its possible to jumper two GPIOs.  One Moonraker writes to, the other is monitored by a script.  The primary issue with this method is that the gpio will turn off when Moonraker restarts.

PR #969, opened 2025-02-26
file_manager: Added per tool weights and total filament change count

Wondro:
This adds a couple more variables that are useful for multi-color prints.


Arksine:
Looks good to me, thanks.

PR #975, opened 2025-03-05
metadata: balanced quoted string split

pedrolamas:
This is a follow up on PR #962 as I just noticed that the split and then parse uses unbalanced regular expressions (so a string like `"bla;bla";"bla"` will produce `['"bla', 'bla"', '"bla"]` instead of `['"bla;bla"', '"bla"]`)

Regex is kind of a personal challenge, so I took a stab in rewriting the pattern with a balanced one!

The new method will accept values with properly balanced double quotes (including escaped double-quotes) or simple values followed by the separators!

Quotes will be stripped and then all spaces will be trimmed in the resulting values.

### DIFF
--- a/moonraker/components/file_manager/metadata.py
+++ b/moonraker/components/file_manager/metadata.py
@@ -71,11 +71,19 @@ def regex_find_ints(pattern: str, data: str) -> List[int]:
 
 def regex_find_strings(pattern: str, separators: str, data: str) -> List[str]:
     pattern = pattern.replace(r"(%S)", r"(.*)")
-    separators = re.escape(separators)
-    delimiters = rf"[{separators}]"
     match = re.search(pattern, data)
     if match and match.group(1):
-        return re.split(delimiters, match.group(1).strip('"'))
+        separators = re.escape(separators)
+        pattern = rf'\s*(")(?:\\"|[^"])*"\s*|[^{separators}]+'
+        parsed_matches: List[str] = []
+        for m in re.finditer(pattern, match.group(1)):
+            (val, sep) = m.group(0, 1)
+            val = val.strip()
+            if sep:
+                val = val[1:-1].replace(rf'\{sep}', sep).strip()
+            if val:
+                parsed_matches.append(val)
+        return parsed_matches
     return []
 
 def regex_find_float(pattern: str, data: str) -> Optional[float]:


pedrolamas:
FWIW, I have been testing this in VSCode in a python Jupyter book, just by dropping the declaration there and then manually running the function.

image

pedrolamas:
I ran a couple of tests with Prusa Slicer and specified filament names and types with spaces and the separators, and the above would parse these without any problem:

image

image

moggieuk:
That is much more reliable and better if users use special characters in names / type definitions. Because slicers don't have any restrictions this looks like a safe way to go.

If this is added why not also add additional array parsing (with plural names) for:
`filament_names`
`filament_types`
That would complete the necessary set of array metadata and remove the need to post process in the various UI's..

The original scalar metadata parsers could remain for backward compatibility.

pedrolamas:
> If this is added why not also add additional array parsing (with plural names) for:
> `filament_names`
> `filament_types`

I tend to agree on that... @Arksine what do you think of this?

Either add 2 new fields (which would mean some duplication) or just assume a breaking change and remove the legacy `filament_name` and `filament_type` and add the new `filament_names` and `filament_types` and ensure that the frontends make the necessary changes to use either one or the other formats (I personally am fine with that from Fluidd side)

Arksine:
I'd rather avoid adding duplicating metadata, and I'd also like to avoid breaking existing metadata.

What we could do instead is special processing for PrusaSlicer derivatives.  If we detect a single filament name, report it as is done now.  If there are multiple filaments perhaps we can report it in a way that is easier to parse, such as a json array.

pedrolamas:
For all matters and purposes, this PR only addresses the currently incorrect parsing of the data (as shown on the examples above)

I used Prusa Slicer as an example but all derivatives have the same output.

> What we could do instead is special processing for PrusaSlicer derivatives. If we detect a single filament name, report it as is done now. If there are multiple filaments perhaps we can report it in a way that is easier to parse, such as a json array.

Unless I am missing something, that is still a breaking change!

Consumers are expecting that `filament_name` is a `string` value, but from what you just said it could then be `string` or `string[]`

Arksine:
I mean an encoded json array.  It would be pretty easy, we do something like the following:

```python
def parse_filament_name():
  result = regex_find_strings(<pattern>)
  if len(result) > 1:
     return json.dumps(result)
  elif result:
    return result[0]
  return None
```

pedrolamas:
(I removed my previous comment as I realize now you were talking about filament name and type)

I guess that is expected as we have JSON that needs to be encoded to be presented in the JSON output! 😁

Arksine:
Given the rules, I wonder if it would be easier and more reliable to use shlex in `regex_find_strings`.  Something like the following snippet:

```python
def regex_find_strings(pattern: str, separators: str, data: str) -> List[str]:
    pattern = pattern.replace(r"(%S)", r"(.*)")
    match = re.search(pattern, data)
    if match and match.group(1):
        lexer = shlex.shlex(match.group(1), posix=True, punctuation_chars=separators)
        return [token for token in lexer if token not in separators]
    return []
```

pedrolamas:
> Given the rules, I wonder if it would be easier and more reliable to use shlex in `regex_find_strings`

That works, though there is a small difference, but it is quite an edge-case that I don't think it is even relevant:

image

pedrolamas:
Seems to work perfectly without the `punctuation_chars` though!

image

pedrolamas:
This seems to fail, even when I set `lexer.quotes = '"'` which is the correct (single quotes are not used as string delimiters here)

image

It separates the `'` from `PETG` instead of keeping as a single token...

My proposed version works fine though! 😁

image

Arksine:
Upon further research, there are other places it will fail.  Given the rules, I thought Prusa might be quoting their names using shell rules.  However `(PLA)` is not quoted in the metadata, and that would fail parsing without adding `(` and `)`.  You also provided another instance of this.  I will go ahead and merge this implementation.

Arksine:
FWIW, my current thought on `filament_name` and `filament_type`:

If its a single item (which is likely to encompass most users), return it as is currently done, with quotes stripped.
If its multiple items return it without quotes stripped and allow the frontend to parse it.

pedrolamas:
> If its a single item (which is likely to encompass most users), return it as is currently done, with quotes stripped. If its multiple items return it without quotes stripped and allow the frontend to parse it.

I guess the only way to know if we have a single value or not is to always use `regex_find_strings`:

- if single value, then return the value as it currently does (the function has already removed the quotes and removed any escapes applied)
  Examples:
  `PLA` -> `"PLA"`
  `"PLA"` -> `"PLA"`
- if multiple value, delimit each value with `"` (and escape any existing `"` inside the value), concatenate all values separated by `;`, and return the result
  Examples:
  `PLA;PLA` -> `"PLA";"PLA"`
  `PLA;'PLA` -> `"PLA";"'PLA"`
  `PLA;"\"PLA"` -> `"PLA";"\"PLA"`

Is this it?

Arksine:
Yes, that was what I was thinking.  However, after more thought, I'm back to json encoding the result.  It doesn't look good double encoded due to the escaped quotes, which threw me off...however if the value includes quotes they are going to be escaped anyway.  Using your examples, it would just look like the following:

`PLA;PLA` -> `["PLA", "PLA"]`
`PLA;'PLA`-> `["PLA", "'PLA"]`
`PLA;"\"PLA"` -> `["PLA", "\"PLA"]`

The frontend can check the first character, if its a `[` then use the json library to decode it.

PR #979, opened 2025-03-12
metadata: filament_name and filament_type as array

pedrolamas:
Following up on #975, this does exactly what we discussed!

For `filament_name` and `filament_type` fields, we now search for multiple strings and then:

- if no value was found, return `None` (current behavior)
- if a single value was found, we return the value (current behavior)
- if more than one value was found, return all values as a serialized JSON blob (new behavior)

I've ran a couple of tests in Fluidd, this seems to work fine!

### DIFF
--- a/moonraker/components/file_manager/metadata.py
+++ b/moonraker/components/file_manager/metadata.py
@@ -429,14 +429,24 @@ class PrusaSlicer(BaseSlicer):
         return None
 
     def parse_filament_type(self) -> Optional[str]:
-        return regex_find_string(
-            r";\sfilament_type\s=\s(%S)", self.footer_data
+        result = regex_find_strings(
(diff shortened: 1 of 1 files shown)


pedrolamas:
For the record, this is what I am using in Fluidd to test this (new code marked):

image

Arksine:
The change looks good to me.  I'll give it bit of time to merge so @moggieuk @Wondro  @meteyou can be aware of the change so it doesn't break their existing implementation.

pedrolamas:
@Arksine considering the lately momentum around MMU support (Happy Hare, Armored Turtle AFC, ...), I believe this is even more important now than ever, do you think we can get this merged soon?

Arksine:
Sorry about that, completely forgot this one was ready.

PR #985, opened 2025-04-02
build: bump pillow for Python >=3.9

Icenowy:
Pillow 10 does not support Python 3.13 and Pillow 11 does not support Python 3.8.

Keep Python 3.8 at Pillow 10 and update to Pillow 11 for newer Python versions.


Arksine:
Thanks.  The change is good, but I think it would be best to go with a new pattern for dealing with dependencies that have python_version specifiers:

```
pillow>=9.5.0, <=11.1.0
```

I used this pattern the last time I bumped Tornado.  Pip will use its own dependency resolver to install/update to the latest version compatible with the current version of Python.  We can do this because I made a minor change to the update manager which adds the `-U` option when calling `pip`, thus it isn't necessary to pin the package to a specific version to force an update.

Icenowy:
Sounds reasonable, changed.

Arksine:
Thanks!

Issue #993, opened 2025-05-29
MQTT power device reports old state after command


mvn23:
### What happened

I am controlling my printer power using a Zigbee smart plug (Ikea Inspelning) via Zigbee2MQTT. Control is working, but when I try to bind the `klipper` service to the power device it is giving me all kinds of trouble.
Sometimes the service starts or stops as expected when using the power device. Most of the time however, the service start/stop logic either works in reverse (starting when powering off or vice versa) or not at all.

### Client

Mainsail

### Browser

Firefox

### How to reproduce

* Use an Ikea Inspelning smart plug to control printer power
* Add it as a power device in `moonraker.conf`, using `command_topic` and `state_topic`
* Bind service `klipper` to the power device
* Try to use it and monitor the `klipper` service state

### Additional information

I have done some debugging on this issue and am working around it at the moment.

I have the following power device config in `moonraker.conf`
```
[power printer]
type: mqtt
locked_while_printing: True
on_when_job_queued: True
bound_services: klipper
qos: 1
command_topic: zigbee/Elzar/set
command_payload:
   {% if command == "on" %}
     ON
   {% else %}
     OFF
   {% endif %}
state_topic: zigbee/Elzar
state_response_template:
   {% set resp = payload|fromjson %}
   {resp["state"]}
query_topic: zigbee/Elzar/get
query_payload:
  {% set my_payload = {"state": ""} %}
  { my_payload|tojson }
```

It seems that when Zigbee2MQTT receives a state change command, it reports the initial state before publishing the new state:
```
$ mosquitto_sub -h 192.168.22.254 -u <<redacted>> -P <<redacted>> -t 'zigbee/Elzar/#' -v
zigbee/Elzar/availability {"state":"online"}
zigbee/Elzar/set ON
zigbee/Elzar {"current":0,"energy":6.71,"identify":null,"last_seen":1748510188250,"linkquality":129,"power":4.8,"power_on_behavior":"previous","state":"OFF","update":{"installed_version":33816645,"latest_version":33816645,"state":"idle"},"voltage":233}
zigbee/Elzar {"current":0,"energy":6.71,"identify":null,"last_seen":1748510188250,"linkquality":129,"power":4.8,"power_on_behavior":"previous","state":"ON","update":{"installed_version":33816645,"latest_version":33816645,"state":"idle"},"voltage":233}

```
This seems to be standard Zigbee2MQTT behavior as my Zigbee wall dimmers show a similar sequence.

[...]

PR #994, opened 2025-05-31
power: Fix MQTT PowerDevice set_power()

mvn23:
Wait for power to reach desired state in set_power()

When setting a power state, the MQTT power device only waited for the next message on the MQTT state topic. This is not guaranteed to give the desired result (see #993). We fix this by waiting for the requested state instead.


Arksine:
Thanks.  This appears to be a workaround for undesirable behavior in a particular MQTT device rather than a fix to Moonraker itself.  When Moonraker issues a request to change device power it is reasonable to expect the next response to be the updated power state.

It appears to me that the proposed workaround could introduce a regression in some scenarios.  For example, if a device experiences an internal error when attempting to power on or off it would be valid to respond with its current state.  If the device requires a query, the request would get stuck in the loop until the issue resolved itself.

That said, I do think that Moonraker should accommodate devices that don't respond as we would expect.  Its possible to extend the context passed to the `state_response_template` with some additional information that would allow you to discard a known bad response for specific devices.  I'll work on an implementation and update the attached issue when it is complete.

mvn23:
Thank you for your response. I have to say that I was surprised to see the old state first as well, but when you think about it it could be any other MQTT message on the same topic, e.g. a change in the `last_seen` or `linkquality` properties. I quickly checked and this actually seems to be where the first message could be coming from.
With that in mind, waiting for the expected state with a timeout seemed like a more sensible approach to me. I have to admit that I'm not familiar enough with the code base to consider the effect this has in certain other situations that you mentioned, so maybe my approach is a bit naive in that regard. Either way I'm glad to see that it's being worked on!

Arksine:
I pushed a commit that should allow users to accommodate device behavior such as the Zigbee in the `state_response_template`.  Two new variables are available, `last_request` and `response_count`.  The `last_request` will be `refresh`, `power_on`, `power_off`, or an empty string.  Most are self explanatory, the empty string will be present if the device publishes a state update to the topic outside of a Moonraker request.  The `response_count` is the number of responses received for the current request (or the number of consecutive updates received not requested by Moonraker).  Additionally, the template can now render a `discard` response in addition to `on` and `off`.

You should be able add something like the following as the `state_response_template`:
```
{% if response_count == 1 and last_request.startswith("power_") %}
  discard
{% else %}
  {% set resp = payload|fromjson %}
  {resp["state"]}
{% endif %}
```

PR #1001, opened 2025-07-05
Improve SimplyPrint progess calculation mimicing what's being done in Mainsail and Fluidd

jbatonnet:
Printers with multiple Klipper versions / implementations are using different fields to provide print progress.
Current implementation only relies on `display_status.progress`. For example, the Anycubic Kobra 3/S1 printers do not use this field.

In order to maintain some consistency with other clients implementation, I reimplemented the same algorithm as Mainsail and Fluidd:
- Mainsail: https://github.com/mainsail-crew/mainsail/blob/6f0f562a0b09ab242f35533b71fa7f99a75e2025/src/store/printer/getters.ts#L37
- Fluidd: https://github.com/fluidd-core/fluidd/blob/d01112461ac49962ca32c8bd4c12d6eec619ac1c/src/store/printer/getters.ts#L123

I also kept current implementation adding a test to make sure all the fields are correctly provided.
This should now be more robust with other printers.

### DIFF
--- a/moonraker/components/simplyprint.py
+++ b/moonraker/components/simplyprint.py
@@ -535,6 +535,7 @@ class SimplyPrint(APITransport):
         query = await self.klippy_apis.query_objects({"heaters": None}, None)
         sub_objs = {
             "display_status": ["progress"],
+            "virtual_sdcard": ["file_position", "progress"],
             "bed_mesh": ["mesh_matrix", "mesh_min", "mesh_max"],
             "toolhead": ["extruder"],
             "gcode_move": ["gcode_position"]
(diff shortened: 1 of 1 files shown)


Arksine:
I'm not sure I agree that this is necessary.  Any printer, regardless of whether it has a display, can enable `display_status` via the configuration, and in fact Moonraker's [documentation](https://moonraker.readthedocs.io/en/latest/installation/#klipper-configuration-requirements) instructs the user to do so.

The `[display_status]` module enables support for M73 status, which should be the most accurate method of indicating print progress.  When M73 progress is not available `display_status` will fall back to `virtual_sdcard.progress`.

jbatonnet:
Some printers are using different Klipper implementations. This is the case for the Anycubic Kobra 3/S1 series.
They do not support display_status and are just emitting None for the progress that crashes the current implementation.
This PRbrings a better implementation in every aspect than the code before with better compatibility and more checks.

AlbertMN:
@Arksine this PR, that @jbatonnet was nice enough to conjure up in no time following a bug report from a user, ensures that we more seamlessly integrate with more printers, requiring no config-tinkering, which would of course be neat!

To my understanding, Mainsail and Fluidd uses similar behavior, and after a couple of bug reports like;
> "SimplyPrint doesn't show the print progress and time left, though Mainsail and Fluidd does"

For some printers, and not others (from the same brands, even...), we went looking for a way to make sure that the setup experience becomes less of a hassle, and that we don't have to add yet another notice in the setup guide for this specific case, for some printers.

Is there anything that would make this more mergeable? Is the issue added `sub_objs`, or the fact that this shouldn't be a Moonraker-thing to fix (compatibility issues, due to bad printer setups - which I would agree with; in a perfect world they would make better stock firmware 😅).

Let us know - we're very motivated to resolve the problem that the given users are facing. But if there's a better way than this, or this PR can be modified a bit to better align with what you want for Moonraker, @Arksine, let us know!

Thanks for getting on this so swiftly 💪

Arksine:
> Some printers are using different Klipper implementations. This is the case for the Anycubic Kobra 3/S1 series.
They do not support display_status and are just emitting None for the progress that crashes the current implementation.
This PRbrings a better implementation in every aspect than the code before with better compatibility and more checks.

I see, we are talking about a fork of Klipper.  Can you provide a link to the source?

jbatonnet:
Not a fork but a reimplementation in Go: https://github.com/ANYCUBIC-3D/Kobra3/tree/main/klipper-go

Arksine:
> Not a fork but a reimplementation in Go: https://github.com/ANYCUBIC-3D/Kobra3/tree/main/klipper-go

Thanks.  At least it appears to be mostly in compliance with the GPL, as although it isn't a fork it is clearly a derivative.  It seems like they don't correctly handle subscriptions to objects that don't exist.

I am going to merge this, however I do want to explain my initial hesitation, as the code itself is fine.  There are two primary reasons:

1) The `display_status` object in Klipper is intended to provide a consistent progress indicator that is shared between displays and API consumers.  It also provides support for M73.  It doesn't currently support the `gcode_start_byte` and `gcode_end_byte`, however that support could be added.  This would be the ideal location for this fix, however I recognize that this would not help the specific scenario outlined in this PR.

2) As a general policy Moonraker officially supports "official" Klipper,  avoiding workarounds for unofficial versions of Klipper. That said,  I will make an exception here due to the nature of the `simplyprint` component, which is unique relative to the rest of Moonraker.

PR #1005, opened 2025-07-24
fix typos in git_deploy.py

thijstriemstra:
found this typo while using fluidd:

Screenshot 2025-07-24 at 21 16 55


thijstriemstra:
there's another similar "typo" here: https://github.com/Arksine/moonraker/blob/72b89d905ee66b621ea3c4c633541e181822b448/moonraker/components/machine.py#L856 but since it's code-related I decided to keep it as is, as not to break stuff.

PR #1007, opened 2025-07-28
git_deploy: Support relative paths in gitfile

bozzzzo:
When a git project managed by moonraker is a submodule of a superproject, gitfile
contains a relative path.

Resolve contents of the gitfile relative to the directory containing the gitfile.

Why the superproject and submodules? Poor man 'restore point', so I can roll back software and configuration to a known-working state.


Arksine:
Looks good, thanks.

PR #1016, opened 2025-09-08
fix: non-strict JSON decode for compat w/ klipper JSON output

SanaaHamel:
Klipper uses the std lib JSON module when `msgspec` is not installed.
This encoder produces non-strict JSON, notably supporting inf/NaN.

Run the `msgspec` JSON decoder in non-strict mode for out-of-the-box compatibility.
This also fixes divergence in behaviour between different settings for `MOONRAKER_ENABLE_MSGSPEC`. If disabled, then the std lib JSON module is used, which non-strict.

See [here](https://jcristharif.com/msgspec/supported-types.html) for details regarding what `strict=False` enables.

### DIFF
--- a/moonraker/utils/json_wrapper.py
+++ b/moonraker/utils/json_wrapper.py
@@ -22,12 +22,16 @@ if _msgspc_var in ["y", "yes", "true"]:
         encoder = msgspec.json.Encoder()
         decoder = msgspec.json.Decoder()
         dumps = encoder.encode  # noqa: F811
-        loads = decoder.decode  # noqa: F811
+
+        def loads(data: str | bytes | bytearray) -> Any:  # noqa: F811
+            return decoder.decode(data, strict=False)
+
         MSGSPEC_ENABLED = True
+
 if not MSGSPEC_ENABLED:
     import json
     from json import JSONDecodeError  # type: ignore # noqa: F401,F811
-    loads = json.loads  # type: ignore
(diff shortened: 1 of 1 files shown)


Arksine:
Thanks.  After reviewing the flag and testing, the `strict` argument allows for type coercion, but doesn't support the non-spec json generated by the standard json module.  For example, the following will fail:

```python
msgspec.json.decode(b'{"key": Infinity}', strict=False)
```

I don't believe msgspec provides an option to support non-finite floats outside of coercing strings.   This would appear to leave you with two options:

1) Require users of your Klipper mod to disable msgspec in Moonraker (and Klipper if they are using it, as non-finite values are encoded as `null` by msgspec)
2) Modify your module so that it does not export non-finite values.

I'm not aware of any other Klipper object or webhooks response that includes a non-finite value.  It should be okay to add `allow_nan=False` in Klipper's json encoder if we want to guarantee compatibility between msgspec and json.

FWIW, if you choose option 1, its entirely possible that it could break a Moonraker client if its using strict json decoding.

SanaaHamel:
The inf value is present in the Klipper configuration JSON as a default value, which is sent to moonraker as part of the setup call by Klipper. Rather rude of them, and outside my control. At this point it's either add hacks to work around the issue w/ Klipper, or get Klipper to fix it their serialization...

Apologies for not verifying/catching that msgspec only does str coercion when non-strict, rather than recognizing new lexemes.

PR #1017, opened 2025-09-09
metadata: adds JPG and QOI thumbnails support

pedrolamas:
This PR add support for JPG and QOI thumbnails, as Prusa Slicer (and derivatives) have support for these also and not only PNG.

image

I have used these [samples files](https://github.com/user-attachments/files/22239701/samples.zip) to test these changes.

### DIFF
--- a/moonraker/components/file_manager/metadata.py
+++ b/moonraker/components/file_manager/metadata.py
@@ -20,6 +20,7 @@ import uuid
 import logging
 import shlex
 import subprocess
+import qoi
 from PIL import Image
 
 # Annotation imports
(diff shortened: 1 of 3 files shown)


Arksine:
Thanks.  I implemented a similar version of this in a slightly different way in commit fa90202eb960ff57e7f0898d15b080f39ef6fbb0 with a few goals:

1) Avoid new dependencies
2) Make it easily extensible for any format supported by Pillow
3) Log formats not currently supported

I also had some other work that touched this code that I went ahead and pushed...specifically I wanted to better handle files with variable length unicode characters.  When doing so I noticed that small files were generating duplicate thumbnail entries because the thumbnail data was being parsed twice.  That should now be corrected.

pedrolamas:
Thanks @Arksine, I just tested your latest changes and they worked perfectly with my sample files, so will deprecate and close this PR! 🙂

PR #1021, opened 2025-09-19
td1: Adding support for interfacing with TD1 devices

jimmyjon711:
Feature was requested by AJAX3D(designer of TD1) to add the ability into AFC-Klipper-Add-On to have the ability to automatically collect data from TD1 that are connected inline with filament paths.

So this PR adds a new TD1 module which can automatically connect and interface with multiple TD1s that are connected to a printer. The purpose of this is to be used with filament changers(or standalone) so that color and TD values can be grabbed for each filament that's loaded. This has initially been integrated into a [development branch](https://github.com/jimmyjon711/AFC-Klipper-Add-On/tree/td1) of AFC-Klipper-Add-On and tested by multiple testers. 

Signed-off-by: Jim Madill <jcmadill1@gmail.com>

### DIFF
--- a/moonraker/components/td1.py
+++ b/moonraker/components/td1.py
@@ -296,7 +296,7 @@ class TD1:
                 self.logger.error(
                     f"Failed to open port:{port} for TD1 serial {serial_number}")
 
-        except serial.serialutil.SerialException as e:
+        except serial.SerialException as e:
             self.logger.error(f"Failed to open serial port {port}: {e}")
             raise
         except asyncio.CancelledError as e:
--- a/moonraker/components/td1.py
+++ b/moonraker/components/td1.py
@@ -10,32 +10,40 @@ import serial
 import serial.tools.list_ports
 import logging
 import datetime
+import contextlib
 
 from ..common import RequestType
 from ..common import WebRequest
 from ..utils import async_serial
+from ..confighelper import ConfigHelper
+from typing import Optional, Dict, Any
+from logging import Logger
 
-class TD1Protocol(asyncio.Protocol):
-    def __init__(self, config, serial_number, port, logger):
+
+class TD1Protocol():
+    def __init__(self, config: ConfigHelper, serial_number: str,
+                 port: str, logger: Logger) -> None:
         self.config = config
         self.server = config.get_server()
+        self.port = port
         self.serial_number = serial_number
-        # Setting port in config file
-        self.config.source.config.set('td1', 'serial', port)
+        self.baudrate = config.getint("baudrate", 115200)
+        self.name = f"{config.get_name()}-{self.serial_number}"
         self.logger = logger
-        self.buffer = ""
-        self.error_state = False
-        self.serial = async_serial.AsyncSerialConnection(config,
-                                                         config.get("baudrate", 115200))
-        self.serial_task = None
-        self.enabled = True
+        self.error_state: bool = False
+        self.enabled: bool = True
+        self.serial = async_serial.AsyncSerialConnection(self.server,
+                                                         self.name,
+                                                         self.port,
+                                                         self.baudrate)
+        self.serial_task: asyncio.Task
 
         # Values to return back for td1_data endpoint
-        self._td = None
-        self._color = None
-        self._scan_time = datetime.datetime.utcnow().isoformat() + "Z"
-        self._error = None
-        self.done_initializing = False
+        self._td: Optional[float] = None
+        self._color: Optional[str] = None
+        self._scan_time: str = datetime.datetime.utcnow().isoformat() + "Z"
+        self._error: Optional[str] = None
+        self.done_initializing: bool = False
 
     async def initialize(self) -> None:
         self.serial_task = self.server.get_event_loop().create_task(self.run_serial())
@@ -45,9 +53,14 @@ class TD1Protocol(asyncio.Protocol):
             if self.serial.connected:
                 break
 
+    async def readline_with_timeout(self, timeout: float = 5.) -> Optional[bytes]:
+        with contextlib.suppress(asyncio.TimeoutError):
+            return await asyncio.wait_for(self.serial.reader.readline(), timeout)
+        return None
+
     async def run_serial(self) -> None:
         try:
-            self.serial.open(timeout=10)
+            self.serial.open()
         except (self.serial.error, OSError):
             self.error_state = True
             self.logger.exception(
@@ -59,20 +72,12 @@ class TD1Protocol(asyncio.Protocol):
 
         self.done_initializing = True
 
-        while self.enabled:
-            async for line in self.serial.reader:
-                self.buffer += line.decode("utf-8")
-                while "\n" in self.buffer:
-                    line, self.buffer = self.buffer.split("\n", 1)
-                    line = line.strip()
-                    if line:
-                        self._parse_line(line)
-                        self.logger.debug(f"Serial: {self.serial_number} Line: {line}")
+        async for read_line in self.serial.reader:
+            line = read_line.decode("utf-8").rstrip()
+            self._parse_line(line)
+            self.logger.debug(f"{self.serial_number}: {line}")
 
-            if self.enabled:
-                await asyncio.sleep(0.5)
-
-    def _parse_line(self, line):
+    def _parse_line(self, line: str) -> None:
         try:
             parts = line.split(",")
             if len(parts) >= 6:
@@ -82,7 +87,7 @@ class TD1Protocol(asyncio.Protocol):
         except Exception as e:
             self.logger.error(f"Parse error from {self.serial_number}: {e}")
 
-    async def check_error_file(self):
+    async def check_error_file(self) -> None:
         """
         Requests error file from TD1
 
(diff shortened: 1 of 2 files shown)


Arksine:
I believe this is a useful addition to Moonraker.  I had some comments about the implementation above.  Thanks.

jimmyjon711:
@Arksine Code has been updated to use `async_serial` and fixed the other things you pointed out. I also moved a lot of the serial logic into `TD1Protocol` class. I probably  made some other mistakes, but please let me know.

I did retest to verify all logic still works like before.

moonraker/components/td1.py
```diff
@@ -0,0 +1,255 @@
...
+        self._vid = 0xE4B2
+        self._pid = 0x0045
+
+    async def initialize(self):
+        self._server.register_endpoint(
+            "/machine/td1_data", RequestType.GET, self._handle_get_data
+        )
+        self._server.register_endpoint(
+            "/machine/td1_reboot", RequestType.POST, self._handle_reboot
+        )
```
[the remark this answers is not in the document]
jimmyjon711: Fixed in latest commit

Arksine:
I pushed some changes to the `async_serial` wrapper.  It can now be instantiated without a config and its possible to use a callback in place of the StreamReader.   I notice you have made some changes, if possible I will attempt to do a review later today. 

At a glance one thing I noticed that I missed last time is that most class methods and functions do not include annotations.  All methods need to be fully annotated (arguments, except `self`, and return value) for the type checker to work correctly.  It isn't necessary to annotate every variable, the type checker will let you know which ones need annotations.

jimmyjon711:
> I pushed some changes to the `async_serial` wrapper. It can now be instantiated without a config and its possible to use a callback in place of the StreamReader. I notice you have made some changes, if possible I will attempt to do a review later today.
> 
> At a glance one thing I noticed that I missed last time is that most class methods and functions do not include annotations. All methods need to be fully annotated (arguments, except `self`, and return value) for the type checker to work correctly. It isn't necessary to annotate every variable, the type checker will let you know which ones need annotations.

I will go back and do this, something that I glanced over when looking over your contribution docs. I did run `mypy` and it didn't bring up anything that I had to annotate.

Arksine:
> I will go back and do this, something that I glanced over when looking over your contribution docs. I did run mypy and it didn't bring up anything that I had to annotate.

FWIW, if a method is not annotated `mypy` will skip type checking on all source code within the method.

moonraker/components/td1.py
```diff
@@ -0,0 +1,316 @@
...
+        self.config = config
+        self.server = config.get_server()
+        self.serial_number = serial_number
+        # Setting port in config file
+        self.config.source.config.set('td1', 'serial', port)
+        self.logger = logger
+        self.buffer = ""
+        self.error_state = False
+        self.serial = async_serial.AsyncSerialConnection(config,
+                                                         config.get("baudrate", 115200))
```
Arksine, line 29: It is no longer necessary to pass a config object to AsyncSerial connection.  It will need a server object, a unique name, the port and the baudrate.

moonraker/components/td1.py
```diff
@@ -0,0 +1,316 @@
...
+                await ret
+
+            if not self.serial_tasks[serial_number].error_state:
+                self.logger.info(f"TD1 listening on {port} (Serial: {serial_number})")
+                self._known_serials.add(serial_number)
+            else:
+                self.logger.error(
+                    f"Failed to open port:{port} for TD1 serial {serial_number}")
+
+        except Exception as e:
```
Arksine, line 278: We may not want to handle a bare exception here.  At the very least its necessary to re-raise `asyncio.CancelledError`.

moonraker/components/td1.py
```diff
@@ -0,0 +1,316 @@
...
+
+            # Find currently connected devices
+            current_serials = set()
+            for port, serial_number in self._find_td1_devices():
+                current_serials.add(serial_number)
+                if serial_number not in self._known_serials:
+                    self.logger.info(f"Hot-plugged TD1 detected: {serial_number}")
+                    try:
+                        await self._start_serial_task(port, serial_number)
+                    except Exception:
```
Arksine, line 241: As with line 278, it is necessary to re-raise `asyncio.CancelledError`.

moonraker/components/td1.py
```diff
@@ -0,0 +1,316 @@
...
+        try:
+            if self.serial_task is not None or not self.serial_task.done():
+                await self.serial_task
+        except asyncio.CancelledError:
+            self.logger.debug(f"TD1 {self.serial_number} task was already canceled")
+
+    async def reboot(self):
+        self.logger.info(f"Rebooting {self.serial_number}")
+        self.enabled = False
+        self.serial_task.cancel()
```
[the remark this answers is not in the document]
jimmyjon711: When doing a reset the RP2040 closes the serial port so this is why I am exiting the task. Then when it reconnects `_watch_for_new_devices` will pick it back up and start a new serial task. Then we need to recheck for the error file as a user might not have fixed the original issue.

moonraker/components/td1.py
```diff
@@ -0,0 +1,316 @@
...
+            self.logger.exception(
+                f"Error trying to open TD1 serial: {self.serial_number}"
+            )
+            return
+
+        await self.check_error_file()
+
+        self.done_initializing = True
+
+        while self.enabled:
```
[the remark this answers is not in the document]
jimmyjon711: This is more me following `wled.py` file as an example and not being as familiar with how async works

jimmyjon711:
@Arksine All comments have been addressed except https://github.com/Arksine/moonraker/pull/1021#discussion_r2366363531. I also removed the commit for `async_serial.py` file. Not sure if you like others to resolve comments but I will keep them unresolved as I would rather you close them to verify that I fixed them correctly. 

Also big thanks for taking your time in helping me through this as I normally don't work with async code like this.

moonraker/components/td1.py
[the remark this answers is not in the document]
Arksine: Ok.  I pushed a small change to `async_serial` that will allow you to gracefully exit the serial task so we don't risk the iterator consuming the `ack` from the `change settings` request.  Something like the following should work:

```python
self.serial.set_read_callback(None, force=True)
with contextlib.suppress(asyncio.TimeoutError):
     await asyncio.wait_for(self.serial_task, 1.)
```

The above will feed an EOF to the stream reader, causing the iterator to exit.  A new reader is then instantiated which will be guaranteed to receive the response to the "change settings" command.

moonraker/components/td1.py
```diff
@@ -0,0 +1,337 @@
...
+            if not self.serial_tasks[serial_number].error_state:
+                self.logger.info(f"TD1 listening on {port} (Serial: {serial_number})")
+                self._known_serials.add(serial_number)
+            else:
+                self.logger.error(
+                    f"Failed to open port:{port} for TD1 serial {serial_number}")
+
+        except serial.serialutil.SerialException as e:
+            self.logger.error(f"Failed to open serial port {port}: {e}")
+            raise serial.serialutil.SerialException
```
Arksine, line 297: This should be `raise` or `raise e` if you want to re-raise an exception.  It isn't valid to raise the type.

moonraker/components/td1.py
```diff
@@ -0,0 +1,337 @@
...
+        except asyncio.CancelledError as e:
+            self.logger.error(f"Task canceled error {e}")
+            raise asyncio.CancelledError
```
Arksine, line 300: Same as line 297.

moonraker/components/td1.py
```diff
@@ -0,0 +1,337 @@
...
+
+            # Find currently connected devices
+            current_serials = set()
+            for port, serial_number in self._find_td1_devices():
+                current_serials.add(serial_number)
+                if serial_number not in self._known_serials:
+                    self.logger.info(f"Hot-plugged TD1 detected: {serial_number}")
+                    try:
+                        await self._start_serial_task(port, serial_number)
+                    except serial.serialutil.SerialException:
```
Arksine, line 256: There is probably no need for a `try` block here since no exception was handled.  Previously I believe that this was a bare exception and I indicated that `asyncio.CanceledError` needs to be re-raised.  So it would need to be something like:

```python
try:
    await self._start_serial_task(port, serial_number)
except asyncio.CancelledError:
    raise
except Exception:
    # log error here if you would like
    continue
```

moonraker/components/td1.py
```diff
@@ -0,0 +1,337 @@
...
+        )
+        self._error = block_content
+
+    async def close(self) -> None:
+        self.enabled = False
+        with contextlib.suppress(asyncio.TimeoutError):
+            await asyncio.wait_for(self.serial.close(), 1)
+        if self.serial_task is not None or not self.serial_task.done():
+            with contextlib.suppress(asyncio.CancelledError):
+                with contextlib.suppress(asyncio.TimeoutError):
```
Arksine, line 155: It isn't necessary to use nested calls to `contextlib.suppress` if you want to suppress multiple exceptions.  You can do the following:

```python
with contextlib.suppress(asyncio.CancelledError, asyncio.TimeoutError):
    await asyncio.wait_for(self.serial_task, 1)
```

moonraker/components/td1.py
```diff
@@ -0,0 +1,341 @@
...
+                await ret
+
+            if not self.serial_tasks[serial_number].error_state:
+                self.logger.info(f"TD1 listening on {port} (Serial: {serial_number})")
+                self._known_serials.add(serial_number)
+            else:
+                self.logger.error(
+                    f"Failed to open port:{port} for TD1 serial {serial_number}")
+
+        except serial.serialutil.SerialException as e:
```
Arksine, line 299: I missed this before, however it appears that `serialutil` is not a valid attribute of `serial`.  This should be `serial.SerialException`.

Arksine:
The above comment is the only functional issue that appears to be outstanding.  There are some annotations that need to be cleaned up, however I can take care of that after merging.

After fixing the above I'm prepared to merge once you confirm that the device works as expected.

jimmyjon711:
> The above comment is the only functional issue that appears to be outstanding. There are some annotations that need to be cleaned up, however I can take care of that after merging.
> 
> After fixing the above I'm prepared to merge once you confirm that the device works as expected.

@Arksine Made change and retested, everything looks to be working correctly.

Endpoint results:
image

Error when turning on TD1:
image

Arksine:
Thanks.  I squashed the commits and merged them in e22da356d13aa4194d691fe2b178ec72469dbf1b and b11a791566c93ade67870b274352c90cfaf00d60.

PR #1025, opened 2025-09-30
metadata: allow generating thumbnails for unknown slicer

Matszwe02:
This will allow for less common or newly created slicers to attach thumbnails into gcode, without a need to report as being a different slicer.

As for example, OrcaSlicer was one of them in the past:
https://github.com/SoftFever/OrcaSlicer/issues/840#issuecomment-1527428054

I think this feature will be helpful, as `BaseSlicer` already contains all the universal logic for thumbnails, and PrusaSlicer forks won't stop being created.

### DIFF
--- a/moonraker/components/file_manager/metadata.py
+++ b/moonraker/components/file_manager/metadata.py
@@ -358,9 +358,6 @@ class UnknownSlicer(BaseSlicer):
     def parse_chamber_temp(self) -> Optional[float]:
         return regex_find_float(r"M191 S(%F)", self.header_data)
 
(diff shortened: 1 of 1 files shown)


Arksine:
Seems reasonable...that said the lack of thumbnails gives the user a visual indicator that something isn't right.  I suppose there are other indicators they can rely on.

I may look into seeing if its possible to identify PrusaSlicer (and its forks) in a more generic fashion.  It should be possible if there is a pattern shared by all that isn't shared with other slicers.

PR #1032, opened 2025-11-14
simplyprint: increase connect timeout from 5.0 to 15.0 seconds

cjavad:
Increase the connection timeout to match the OctoPrint implementation, not sure if there was any historical reason for it to be 5 seconds, which we have had reports on not being sufficient in some cases / places on earth.

TBT, needs testing before being merged in, these two places should be the only "hot" paths, but perhaps somehow we can augment the clients default connect timeout, that would be a preferred solution.

### DIFF
--- a/moonraker/components/simplyprint.py
+++ b/moonraker/components/simplyprint.py
@@ -200,7 +200,7 @@ class SimplyPrint(APITransport):
                 log_connect = False
             try:
                 self.ws = await tornado.websocket.websocket_connect(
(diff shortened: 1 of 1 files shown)


cjavad:
@AlbertMN for review, tested, addresses Qidi Q2 issues (underpowered hardware/network capabilties)

AlbertMN:
LGTM

Arksine:
Thanks.

PR #1033, opened 2025-11-21
Adds OIDC support

### LINKED ISSUE #1008: OIDC Auth for moonraker

Would be great to use oidc to authenticate to moonraker. LDAP and AD have support already, would be nice to add something more accessable as an authentication method.

Impliment OIDC Authentication alternative

_No response_

### Additional information

_No response_

pedrolamas:
Adds OIDC support to Moonraker.

I have successfully tested this with both Github and Google, expect others will work too!

To test, you will need to create a new OAuth/OIDC configuration in some provider, though GitHub is rather easy:
- https://github.com/settings/developers
- create a new OAuth app
- fill the details with something like this (make sure the URL's match your Moonraker server):
  image
- click Generate new Client Secret and copy that value
- you will also need the Client Id

On moonraker.conf, add this config:

```ini
[oidc github]
client_id: XXXXXXX
client_secret: XXXXXXX
authorization_url: https://github.com/login/oauth/authorize
token_url: https://github.com/login/oauth/access_token
userinfo_url: https://api.github.com/user
jwks_uri: https://token.actions.githubusercontent.com/.well-known/jwks
scope: openid profile email
redirect_uri: http://localhost:7125/access/oidc/callback
```

This NodeJs script will help testing: https://gist.github.com/pedrolamas/9ebfbd835c10ffe2d626de2b0ffdac22

Just save the script somewhere and run with `node oidcTest.mjs` then follow the script instructions.

Resolves #1008

### DIFF
--- a/moonraker/common.py
+++ b/moonraker/common.py
@@ -196,6 +196,29 @@ class UserInfo:
     def as_dict(self) -> Dict[str, Any]:
         return dataclasses.asdict(self)
 
+@dataclasses.dataclass
+class OIDCLoginRecord:
+    login_id: str
+    state: str
+    provider: str
+    next_url: Optional[str]
+    expires_at: float
+
+    def as_dict(self) -> Dict[str, Any]:
+        return dataclasses.asdict(self)
+
+@dataclasses.dataclass
+class OIDCTokenRecord:
+    login_id: str
+    username: str
+    access_token: str
+    refresh_token: str
+    source: str
+    expires_at: float
+
+    def as_dict(self) -> Dict[str, Any]:
+        return dataclasses.asdict(self)
+
 class Redirect:
     def __init__(
         self,
--- a/moonraker/components/authorization.py
+++ b/moonraker/components/authorization.py
@@ -20,7 +20,13 @@ import logging
 from tornado.web import HTTPError
 from libnacl.sign import Signer, Verifier
 from ..utils import json_wrapper as jsonw
-from ..common import RequestType, TransportType, SqlTableDefinition, UserInfo
+from ..common import (
+    RequestType,
+    TransportType,
+    SqlTableDefinition,
+    UserInfo,
+    Redirect
+)
 
 # Annotation imports
 from typing import (
@@ -41,6 +47,7 @@ if TYPE_CHECKING:
     from .database import MoonrakerDatabase as DBComp
     from .database import DBProviderWrapper
     from .ldap import MoonrakerLDAP
+    from .oidc import MoonrakerOIDC
     IPAddr = Union[ipaddress.IPv4Address, ipaddress.IPv6Address]
     IPNetwork = Union[ipaddress.IPv4Network, ipaddress.IPv6Network]
     OneshotToken = Tuple[IPAddr, Optional[UserInfo], asyncio.Handle]
@@ -146,6 +153,9 @@ class Authorization:
                 "[authorization]: Option 'default_source' set to 'ldap',"
                 " however [ldap] section failed to load or not configured"
             )
+        self.oidc: Optional[MoonrakerOIDC] = None
+        if config.get_prefix_sections('oidc '):
+            self.oidc = self.server.load_component(config, "oidc", None)
         database: DBComp = self.server.lookup_component('database')
         self.user_table = database.register_table(UserSqlDefinition())
         self.users: Dict[str, UserInfo] = {}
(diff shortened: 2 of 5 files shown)


pedrolamas:
~@Arksine I too a dependency on PyJWT but I think that can go if we just use the `decode_jwt` from `authorization.py`... but I don't think I can call that from `oidc.py`... what do you suggest is the best approach here?~

I stand corrected, we need a custom approach here, which at the moment is bia PyJWT (though we might be able to do it manually is we don't want to take a dependency, but this should work as-is right now)

Arksine:
Thanks for working on this.  I'll try to dig into the changes and test them within the next few days.

I do think that requiring every user to register a fork of Moonraker as an oauth app may be a show stopper.  I don't think GitHub intends applications to use OAuth in this manner, and as we previously discussed it would seem to limit users to a single instance of Moonraker.

I did take the time to re-familiarize myself with GitHub's oauth documentation, and I suspect we are going to have to implement the device flow.  This would allow Moonraker to register a single app, and the client ID would be included in the source.  There are still rate limits and other issues I need to work out, but its something I'm willing to try to tackle.  I do suspect that some of the code in this PR will be applicable though.

pedrolamas:
> I did take the time to re-familiarize myself with GitHub's oauth documentation, and I suspect we are going to have to implement the device flow. This would allow Moonraker to register a single app, and the client ID would be included in the source. There are still rate limits and other issues I need to work out, but its something I'm willing to try to tackle. I do suspect that some of the code in this PR will be applicable though.

To be honest, I actually thought the whole device flow was not a standard, I see now it is https://datatracker.ietf.org/doc/html/rfc8628

So yeah, I fully agree with you, this might be the best course of action.

I am fine with having pre-configured Client ID on the code base to make it easier for users, but I do believe we should also have that configurable for those who want to use a custom provider (I've seen a lot of self-hosting solutions on this by the way!)

I might leave this one as is and take a look at how to implement the above RFC now... 😁

shxshxshxshx:
@Arksine 
> I do think that requiring every user to register a fork of Moonraker as an oauth app may be a show stopper. I don't think GitHub intends applications to use OAuth in this manner [...]

'Application' refers to individual *deployments* in the context of OAuth, and every user registering their own application is indeed how Github (and other OAuth providers) intend their service to be used.

> [...] as we previously discussed it would seem to limit users to a single instance of Moonraker.

The Name/Description fields only exist for user convenience - users running multiple instances of Moonraker are expected to register separate applications for each.

@pedrolamas

This should implement a configurable group check if you intend it to be used with public OAuth providers - having access to *any* Github/Google/etc account would currently be enough to get access.

Though we arguably shouldn't, using public OAuth providers to secure private applications is hard to do safely for individual users and tends to require provider-specific implementations.

-- 

As a more general authentication note, we should strongly consider replacing the client-facing token flow with cookie-based sessions for OAuth - giving access tokens to a SPA (ie Fluidd/Mainsail) turns any XSS vulnerability it might have into a token compromise.

The same issue applies to the existing authentication methods as well, though it'd represent a extensive breaking change for those.

Arksine:
> 'Application' refers to individual deployments in the context of OAuth, and every user registering their own application is indeed how Github (and other OAuth providers) intend their service to be used.

In general I agree that the application refers to deployments, at least with regard to the web/code flow.  That said, its general use case is for web deployments, where users are not  expected to register the application with their OAuth provider.

The best analog I can give right now is VSCode.   You can log in to GitHub through VSCode, presumably using an implementation of OAuth.  It is not necessary for the user register their own instances of VSCode on GitHub.   If users were required to do so then I suspect very limited subset would bother.

In our use case the Device Flow solves this.  Authentication can occur using a primary client ID, however it doesn't prohibit users from creating and registering their own if they wish.  For example, a print farm may prefer to use their own client ID.

> As a more general authentication note, we should strongly consider replacing the client-facing token flow with cookie-based sessions for OAuth - giving access tokens to a SPA (ie Fluidd/Mainsail) turns any XSS vulnerability it might have into a token compromise.

The front end should not need OAuth tokens.  Authentication should be strictly between Moonraker and the OAuth Provider, the front end only needs the verification uri and the user code.  Once the user is authenticated Moonraker would generate its own JWTs to be used by the front end for authorization.   These tokens are relatively short lived unless the attacker manages to get a refresh token.  Even in that scenario the damage is limited vs a compromised OAuth token.

We need to avoid depending on cookies as we cannot assume a client is a web app running in a browser.  This is another reason why the OAuth 2 device flow is the likely best fit for Moonraker.

PR #1035, opened 2025-11-22
authorization: add OAuth2 Device Flow support

pedrolamas:
Adds OAuth2 Device Authorization Flow support as specified on https://datatracker.ietf.org/doc/html/rfc8628

To test, you will need to create a new OAuth configuration in some provider, though GitHub is rather easy:
- https://github.com/settings/developers
- create a new OAuth app
- fill the details with something like this (I don't think the URL's matter for device auth flow but ensure "Enable Device Flow" is ticked):
  image
- copy the Client Id (github does not require Client Secret, but other providers might - Google does!)

On moonraker.conf, add this config:

```ini
[oauth2 github]
client_id: XXXXXX
# GitHub device flow doesn't require client_secret for public apps
# client_secret: XXXXX
device_authorization_endpoint: https://github.com/login/device/code
token_endpoint: https://github.com/login/oauth/access_token
userinfo_endpoint: https://api.github.com/user
# GitHub doesn't provide JWKS for device flow, so we'll use userinfo endpoint
jwks_uri:
scope: read:user user:email
```

This NodeJs script will help testing: https://gist.github.com/pedrolamas/593c918a3c6f748ff920e999eecbecdf

Just save the script somewhere and run with `node oauth2DeviceFlowTest.mjs` then follow the script instructions.

### DIFF
--- a/moonraker/components/authorization.py
+++ b/moonraker/components/authorization.py
@@ -41,6 +41,7 @@ if TYPE_CHECKING:
     from .database import MoonrakerDatabase as DBComp
     from .database import DBProviderWrapper
     from .ldap import MoonrakerLDAP
+    from .oauth2 import MoonrakerOAuth2
     IPAddr = Union[ipaddress.IPv4Address, ipaddress.IPv6Address]
     IPNetwork = Union[ipaddress.IPv4Network, ipaddress.IPv6Network]
     OneshotToken = Tuple[IPAddr, Optional[UserInfo], asyncio.Handle]
@@ -145,6 +146,11 @@ class Authorization:
             self.server.add_warning(
                 "[authorization]: Option 'default_source' set to 'ldap',"
                 " however [ldap] section failed to load or not configured"
+            )
+        self.oauth2: Optional[MoonrakerOAuth2] = None
+        if config.get_prefix_sections('oauth2 '):
+            self.oauth2 = self.server.load_component(
+                config, "oauth2", None
             )
         database: DBComp = self.server.lookup_component('database')
         self.user_table = database.register_table(UserSqlDefinition())
@@ -265,6 +271,18 @@ class Authorization:
         )
         self.server.register_endpoint(
             "/access/info", RequestType.GET, self._handle_info_request,
+            transports=TransportType.HTTP | TransportType.WEBSOCKET,
+            auth_required=False
+        )
+        self.server.register_endpoint(
+            "/access/oauth2/device/start", RequestType.POST,
+            self._handle_oauth2_start,
+            transports=TransportType.HTTP | TransportType.WEBSOCKET,
+            auth_required=False
+        )
+        self.server.register_endpoint(
+            "/access/oauth2/device/poll", RequestType.POST,
+            self._handle_oauth2_poll,
             transports=TransportType.HTTP | TransportType.WEBSOCKET,
             auth_required=False
         )
@@ -402,9 +420,13 @@ class Authorization:
             request_trusted = True
         elif req_ip is not None:
             request_trusted = await self._check_authorized_ip(req_ip)
+        oauth2_providers: List[str] = []
+        if self.oauth2 is not None:
+            oauth2_providers = self.oauth2.get_provider_names()
         return {
             "default_source": self.default_source,
             "available_sources": sources,
+            "oauth2_providers": oauth2_providers,
             "login_required": login_req,
             "trusted": request_trusted
         }
@@ -560,6 +582,30 @@ class Authorization:
             action = "user_logged_in"
             if hashed_pass != user_info.password:
                 raise self.server.error("Invalid Password")
+        token, refresh_token = await self._generate_user_tokens(
+            user_info
+        )
+        conn = web_request.get_client_connection()
+        if create:
+            event_loop = self.server.get_event_loop()
+            event_loop.delay_callback(
+                .005, self.server.send_event,
+                "authorization:user_created",
+                {'username': username})
+        elif conn is not None:
+            conn.user_info = user_info
+        return {
+            'username': username,
+            'token': token,
+            'source': user_info.source,
+            'refresh_token': refresh_token,
+            'action': action
+        }
+
+    async def _generate_user_tokens(
+        self, user_info: UserInfo
+    ) -> Tuple[str, str]:
+        username: str = user_info.username
         jwt_secret_hex: Optional[str] = user_info.jwt_secret
         if jwt_secret_hex is None:
             private_key = Signer()
@@ -578,22 +624,7 @@ class Authorization:
         refresh_token = self._generate_jwt(
             username, jwk_id, private_key, token_type="refresh",
             exp_time=datetime.timedelta(days=self.login_timeout))
-        conn = web_request.get_client_connection()
-        if create:
-            event_loop = self.server.get_event_loop()
-            event_loop.delay_callback(
-                .005, self.server.send_event,
-                "authorization:user_created",
-                {'username': username})
-        elif conn is not None:
-            conn.user_info = user_info
-        return {
-            'username': username,
-            'token': token,
-            'source': user_info.source,
-            'refresh_token': refresh_token,
-            'action': action
-        }
+        return token, refresh_token
 
     async def _delete_jwt_user(self, web_request: WebRequest) -> Dict[str, str]:
         username: str = web_request.get_str('username')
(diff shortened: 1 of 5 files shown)


moonraker/components/authorization.py
```diff
@@ -933,6 +964,71 @@ def get_api_key(self) -> Optional[str]:
...
+        error, provider, username = await self.oauth2_device_flow.poll_token(
+            device_code
+        )
+
+        if error:
+            return {
+                'error': error
+            }
+
+        if username not in self.users:
```
pedrolamas: Another one: right now we are creating a new user when someone passes the oauth2 flow, which means virtually _anyone_ would be able to access this instance with any Google account...

If using a custom realm or custom OAuth2 provider, then that would be fine I guess, as they could just control who can access that realm directly on the provider!

Arksine:
Thanks.  I have several high level thoughts and observations, but before I get to that I think we should discuss two potential showstoppers.  The first is the one you brought up in a comment, anyone with an account with a provider can log in.   We need to resolve this before merging, as it would completely break authentication.  Fortunately I think there is a relatively easy solution, and that is we only allow new oauth accounts to be created locally through an authenticated request.  This is how local account creation is handled.

The second is a bit more difficult. When reading GitHub's oath2 docs, its clear that they want to avoid applications from repeatedly requesting new tokens.  This could be a problem for us, as a user may want to monitor prints through multiple devices, and each of these devices could require a log in.  In addition, if the user has multiple printers, each instance of Moonraker would require a log in.   Users may also wish to log out to prevent unauthorized access when they are not using their printers.

I think the only solution here is documentation.  We need to warn users that some 3rd party OAuth providers may not be appropriate for users that want to monitor multiple printers, frequently log out, etc.  These users may find a provider that doesn't mind it, or I suppose it would be possible for them to host their own provider.

With that out of the way, I'll begin addressing the implementation.  The first thing that stands out is the new dependency on PyJWT.  Unfortunately we can't add it due to its dependency on `cryptography`.   The crypto package requires rust to build, and installing rustup on low resource SBCs is difficult.  It isn't an issue for Raspberry Pi's since they have prebuilt wheels, but it is commonly an issue with other SBCs.  This is the main reason why I didn't use pyjwt when adding JWT/JWK support for Moonraker authentication.

Below are some other observations and recommendations.  I'm completely open to discussion on these:
- I think we can shorten the name of the component to `oauth2.py`, and simplify the source names by prefixing `oauth2:`.
- Loading prefix modules and registering `oath2` specific endpoints should be moved to `oauth2.py`.  The authorization module can then just call `server.load_component("oauth2")`.
- The login flow in Moonraker should probably be similar to `LDAP`.  Something like the following:
  - Oauth user logs in through `/access/login`
  - Moonraker requests the user code.  It emits the code, verification uri, etc through a notificaton.  An endpoint returning this data can also be provided, which front ends can poll for HTTP exclusive access.
  - Moonraker polls the access token.  Once received it logs in the user and returns the `/access/login` request
  - The first `oauth` login attempt should be done through an authenticated `POST /access/user` request.

pedrolamas:
> Fortunately I think there is a relatively easy solution, and that is we only allow new oauth accounts to be created locally through an authenticated request. This is how local account creation is handled.

Not sure I understand this correctly; right now we are creating a new account for any LDAP authenticated user (which does make some sense to me as LDAP normally is locally controlled anyway).

Anyone going through the above will be authenticated in terms of having valid OAuth2 account (on GitHub, Google, etc.) and as it stands, would have an account created on Moonraker - something I feel is absolutely wrong!

I would argue that users should first create the local accounts for OAuth2 accounts they would want to use (so one would login with a local Moonraker user account, create a new user account for the OAuth2 email, and then be able to login via OAuth2)

Ideally, we could have that as an opt-out setting (so users using custom OAuth2 realms could allow for automatic account creation in Moonraker).

> The second is a bit more difficult. When reading GitHub's oath2 docs, its clear that they want to avoid applications from repeatedly requesting new tokens. This could be a problem for us, as a user may want to monitor prints through multiple devices, and each of these devices could require a log in. In addition, if the user has multiple printers, each instance of Moonraker would require a log in. Users may also wish to log out to prevent unauthorized access when they are not using their printers.
>
> I think the only solution here is documentation. We need to warn users that some 3rd party OAuth providers may not be appropriate for users that want to monitor multiple printers, frequently log out, etc. These users may find a provider that doesn't mind it, or I suppose it would be possible for them to host their own provider.

This issue only applies if we have a shared Moonraker clientId, and indeed we can have that but indicate on the docs that users are recommended to create their own OAuth2 app and use their clientId to avoid such issues - this is quite standard in apps that have OAuth2 login support.

> The first thing that stands out is the new dependency on PyJWT. Unfortunately we can't add it due to its dependency on `cryptography`. The crypto package requires rust to build, and installing rustup on low resource SBCs is difficult. It isn't an issue for Raspberry Pi's since they have prebuilt wheels, but it is commonly an issue with other SBCs. This is the main reason why I didn't use pyjwt when adding JWT/JWK support for Moonraker authentication.

Fair enough, I wasn't aware of this and truth be told, we can drop the whole JWT verification as it currently stands!

I will see if we can do it manually and not depend on a 3rd party module, if not we can drop this - GitHub for example, doesn't even support it, and in a world where everything is under HTTPS, I don't see not having JWT verification as a significant problem anyway.

> The login flow in Moonraker should probably be similar to `LDAP`. Something like the following:
> * Oauth user logs in through `/access/login`
> * Moonraker requests the user code.  It emits the code, verification uri, etc through a notificaton.  An endpoint returning this data can also be provided, which front ends can poll for HTTP exclusive access.
> * Moonraker polls the access token.  Once received it logs in the user and returns the `/access/login` request
> * The first `oauth` login attempt should be done through an authenticated `POST /access/user` request.

Looking at that flow, seems to me we would then support OAuth2 authentication via socket only, as the `/access/login` as a webrequest would be stuck waiting to return and could eventually timeout - is this something we actually want to do? Also, how would we do to get the url and code in a webrequest scenario?

LDAP auth returns really fast and doesn't require user intervention, so works fine in socket or webrequest - OAuth2 is a different matter, I'd reckon! 😁

pedrolamas:
@Arksine regarding the `pyjwt[crypto]`, I am trying to replace it with `python-jose`, do you think any of [these backends](https://github.com/mpdavis/python-jose#cryptographic-backends) would work for us?

Arksine:
> Not sure I understand this correctly; right now we are creating a new account for any LDAP authenticated user (which does make some sense to me as LDAP normally is locally controlled anyway).

Correct, filters are applied to LDAP users to avoid unauthorized logins. Thus we can safely create new local accounts through `/access/login`.

> Anyone going through the above will be authenticated in terms of having valid OAuth2 account (on GitHub, Google, etc.) and as it stands, would have an account created on Moonraker - something I feel is absolutely wrong!

> I would argue that users should first create the local accounts for OAuth2 accounts they would want to use (so one would login with a local Moonraker user account, create a new user account for the OAuth2 email, and then be able to login via OAuth2)

I agree.  My suggestion is to use `POST /access/user` to log in to OAuth accounts that haven't previously been created locally.  That request must come from an authenticated source (either a currently logged in user or a trusted client).

> This issue only applies if we have a shared Moonraker clientId, and indeed we can have that but indicate on the docs that users are recommended to create their own OAuth2 app and use their clientId to avoid such issues - this is quite standard in apps that have OAuth2 login support.

Perhaps I'm unaware, but most applications that support OAuth2 are web apps.  Even though Moonraker uses web technologies it is not intended to be web hosted and a users commonly run multiple instances of Moonraker.   Therefore even if a user registers their own application with a 3rd party provider they could hit rate limiting issues.  Regardless I think we are on the same page here.

> Looking at that flow, seems to me we would then support OAuth2 authentication via socket only, as the /access/login as a webrequest would be stuck waiting to return and could eventually timeout - is this something we actually want to do? Also, how would we do to get the url and code in a webrequest scenario?

Yes, the request would block until the user logs in and could timeout.  I don't see an issue with that, in fact I will probably need to add a lock later to prevent re-entrancy (we wouldn't want multiple login requests occurring concurrently).

My suggestion provides two ways for front-ends to get the code.  Websocket clients will receive a notification.  Other clients can poll an endpoint (which we will have to implement) that returns the code, uri, etc.  We could implement the login with another endpoint, but I think the flow should be the same.  Note that if we do add an oauth specifc login endpoint then that endpoint would need to be added to the authorization allowlist, otherwise it would be impossible to log in without already being authenticated.

> @Arksine regarding the pyjwt[crypto], I am trying to replace it with python-jose, do you think any of [these backends](https://github.com/mpdavis/python-jose#cryptographic-backends) would work for us?

Each one of those backends raises concerns.  Cryptography we mentioned.  PyCryptodome has a bunch of c extensions, I would need to know the build time on MIPS and armv6 before I would be comfortable adding it.  The `Python-RSA` package has been archived and will receive no further updates, which will complicate things with future versions of  Python.

Since Token validation is an OIDC spec rather than OAuth2 (and optional at that) I think its fine to omit it for now.   I'm willing to look into adding it later.  I'm confident that we can validate access tokens with no additional deps, but validating ID tokens against the jwk will require additional deps.

PR #1036, opened 2025-11-22
power: support HTTP digest authentication for Shelly Gen 2+ devices

### LINKED ISSUE #774: Authentication failure Shelly Plug Plus S
### What happened

moonraker failes authenication with shelly plug plus s. 
https://www.shelly.com/en/products/shop/shelly-plus-plug-s

This results in the following log:
```
2023-12-17 21:14:05,815 [power.py:init_state()] - Device Init Error: shelly_plug
tornado.httpclient.HTTPClientError: HTTP 401: Unauthorized

The above exception was the direct cause of the following exception:

  File "/home/user/moonraker/moonraker/components/power.py", line 451, in init_state
    state = await self._send_status_request()
  File "/home/user/moonraker/moonraker/components/power.py", line 1020, in _send_status_request
    res = await self._send_shelly_command("info")
  File "/home/user/moonraker/moonraker/components/power.py", line 1017, in _send_shelly_command
    return await self._send_http_command(url, command)
  File "/home/user/moonraker/moonraker/components/power.py", line 484, in _send_http_command
    response.raise_for_status(
  File "/home/user/moonraker/moonraker/components/http_client.py", line 363, in raise_for_status
    raise ServerError(msg, code) from self._error
moonraker.utils.ServerError: Error sending 'shelly' command: info
```

### Client

Mainsail

### Browser

Chrome

## config:
```c++
[power printer]
type: shelly
address: shelly_plug
password: example_password
off_when_shutdown: True
locked_while_printing: True
on_when_job_queued: True
restart_klipper_when_powered: True
restart_delay: 2
bound_service: klipper
```

### Additional information

I have had a the simmular issue with OctoPrint PSU control pluging. I managed to solve it with the following thread.
https://github.com/edekeijzer/OctoPrint-PSUControl-Shelly/issues/16

The issue seems to be related to the authentication method used by the plugin. As the new plug use a different method (HTTPDigestAuth). The provided solution has the following changes to handle this change in authentication method.

```python
if self.config['ng_device']:
    auth = HTTPDigestAuth(self.config['username'], self.config['password'])
else:
    auth = HTTPBasicAuth(self.config['username'], self.config['password'])
```

rschaeuble:
Starting with Gen 2, Shelly devices switched from Basic Auth to Digest Auth. This is now supported for all HTTPDevice subclasses.

To support Digest Auth, we have to switch from simple_httpclient to curl_httpclient.

I could successfully test this with my Shelly Plug. I don't have any devices that require Basic Auth, so I could only use log output to test that Basic Auth still works. Would be great if someone with a suitable device could verify that.

Fixes #774.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1073,6 +1073,10 @@ Currently only Gen 1 Shelly devices support Authentication
 address:
 #   A valid ip address or hostname for the shelly device.  This parameter
 #   must be provided.
+auth_type:
+#   The type of HTTP authentication to use (if username and password are specified).
+#   Defaults to "basic" (for Gen 1 Shelly devices). For Gen 2 and later, "digest"
+#   has to be used.
 user:
 #   A user name to use for request authentication.  This option accepts
 #   Jinja2 Templates, see the [secrets] section for details.  If no password
--- a/moonraker/components/http_client.py
+++ b/moonraker/components/http_client.py
@@ -33,10 +33,9 @@ if TYPE_CHECKING:
     from io import BufferedWriter
     StrOrPath = Union[str, pathlib.Path]
 
-MAX_BODY_SIZE = 512 * 1024 * 1024
 AsyncHTTPClient.configure(
-    None, defaults=dict(user_agent="Moonraker"),
-    max_body_size=MAX_BODY_SIZE
+    "tornado.curl_httpclient.CurlAsyncHTTPClient",
+    defaults=dict(user_agent="Moonraker")
 )
 
 GITHUB_PREFIX = "https://api.github.com/"
@@ -82,7 +81,9 @@ class HttpClient:
         send_etag: bool = True,
         send_if_modified_since: bool = True,
         basic_auth_user: Optional[str] = None,
-        basic_auth_pass: Optional[str] = None
+        basic_auth_pass: Optional[str] = None,
+        digest_auth_user: Optional[str] = None,
+        digest_auth_pass: Optional[str] = None
     ) -> HttpResponse:
         cache_key = url.split("?", 1)[0]
         method = method.upper()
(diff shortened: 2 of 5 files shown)


Arksine:
Thanks.  Unfortunately we cannot switch the `curl` backend.  I attempted this at one time and quickly ran into the issue described in [Tornado's Documentation](https://www.tornadoweb.org/en/branch4.5/httpclient.html).

> Note that if you are using curl_httpclient, it is highly recommended that you use a recent version of libcurl and pycurl. Currently the minimum supported version of libcurl is 7.22.0, and the minimum version of pycurl is 7.18.2. It is highly recommended that your libcurl installation is built with asynchronous DNS resolver (threaded or c-ares), otherwise you may encounter various problems with request timeouts (for more information, see http://curl.haxx.se/libcurl/c/curl_easy_setopt.html#CURLOPTCONNECTTIMEOUTMS and comments in curl_httpclient.py).

Some distros package `libcurl` without asynchronous DNS support, resulting in timeout errors leading to regressions throughout Moonraker.

At a higher level, I don't think it is necessary to add native Shelly Gen 2+ support to Moonraker, as there are alternative ways to control these switches (MQTT, Home Assistant).  If a significant number of users would find native support useful I would consider it, but it would likely require the use of an external client such as `httpx`.

rschaeuble:
Ah, that's unfortunate that we can't use the curl client.

Using home assistant as a workaround is what I'm doing right now. The beauty of controlling the Shelly directly is that the overall system is less complex. So while it's not strictly necessary, I think it would be really nice.

On a first glace `httpx` looks good. But I have absolutely no feeling for how easy or hard it would be to replace the old http client with `httpx`, so I'm not sure if I want to try it ;)

Issue #1037, opened 2025-11-23
[power] configured for MQTT uses stale state


cachamber:
### What happened

When the [power] component is configured to use MQTT - turning on/off the printer via mainsail utilizes stale state.  On becomes off and off becomes on. 

### Client

Mainsail

### Browser

Brave

### How to reproduce

Moonraker configuration:
```
[power printer]
type: mqtt
initial_state: off
command_topic: zigbee2mqtt/ZB_Plug_E/set
command_payload: {command}
retain_command_state: True
state_timeout: 5
state_topic: zigbee2mqtt/ZB_Plug_E
state_response_template:
    {% set resp = payload|fromjson %}
    { resp["state"] }
query_topic: zigbee2mqtt/ZB_Plug_E/get/state
off_when_shutdown: True
off_when_shutdown_delay: 1
query_after_command: True
locked_while_printing: True
on_when_job_queued: True
restrict_action_processing: False
bound_services: 
     klipper
```
Use the power control to turn the MQTT connected device off and on.  

### Additional information

Versions:
<img width="479" height="544" alt="Image" src="https://github.com/user-attachments/assets/8b8b7925-1a03-447a-994d-d2c5e8970bfd" />

Initial State: OFF
Clicked the **Power On** button - power on event sent via MQTT - and printer turned on, but UI still has OFF state, and tries to stop klipper due to the bound service.
<img width="1203" height="332" alt="Image" src="https://github.com/user-attachments/assets/fdc42798-ef8c-4f51-8f8a-3b0a161a59ce" />
From log:
`2025-11-22 23:15:38,566 [power.py:process_bound_services()] - Power Device printer: Performing stop action on bound service klipper
`

State: ON
Printer was on - used the slider to turn off the printer, MQTT sent OFF command - printer turned off, UI doesn't detect - in fact - logging shows an ON command and attempts to start klipper because of the wrong state:
<img width="1199" height="446" alt="Image" src="https://github.com/user-attachments/assets/06224b2b-3ff0-4dcb-b86c-49bec0a5cfc4" />
From log:
```
2025-11-22 23:16:40,777 [power.py:process_bound_services()] - Power Device printer: Performing start action on bound service klipper
2025-11-22 23:16:50,324 [power.py:_power_off_on_shutdown()] - Powering off device 'printer' due to klippy shutdown
```

I added additional logging to power.py to track what was going on - but no indication of printer actual 
[...]

PR #1038, opened 2025-11-23
power: wait for matching MQTT state response in set_power

cachamber:
fixes: #1037

Avoid acting on stale retained MQTT state messages. After publishing a command this change waits up to `state_timeout` for a state update that matches the requested state (e.g. `on`) before accepting success. This avoids transient transitions caused by retained/stale messages.


cachamber:
NA

PR #1041, opened 2025-12-05
power: support TP-Link Tapo Smart Plug

satoshi-ohba:
Hello,

I developed this feature primarily for my own needs, so I integrated the TP-Link Tapo smart plug support into moonraker. What do you think?
I read past discussions and know moonraker doesn't accept TP-Link Tapo support that depends on external libraries.
I submitted it anyway, just in case.
I added code to `power.py` that calls the Unofficial Tapo API Client created by Mihai Dinculescu.
[GitHub](https://github.com/mihai-dinculescu/tapo) / [PyPI](https://pypi.org/project/tapo/).
- The client library is under the MIT license.
- The client library itself has no further dependencies on additional external third-party libraries or Python packages.
- The client library itself is written in Rust, and I didn't feel the need to touch Rust directly during the integration process at all.
- PyPI wheels are available for a wide range of architectures. (Supporting x86_64, arm64, aarch64, armv7l, i686, ppc64le, and s390x)
- All client API calls are asynchronous, which should prevent interference with moonraker's overall operation.
- Supported smart plugs: (P100, P105, P110, P110M, P115), power strips (P300, P304M, P316M)
  I have confirmed that it works with the P110M. We need to check with the other plugs and power strips.
- Since the client library supports Python 3.11+, to avoid conflicts with Moonraker running on 3.7<= and <3.11, I initially implemented it using dynamic imports to ensure it only activates for those who really need it. I understand this approach depends on design philosophy; it could be rewritten for static imports or made installable solely via `[project.optional-dependencies]` in pyproject.toml.

I appreciate you taking the time to review this. Any feedback on the approach or integration method is welcome, even if the feature ultimately isn't accepted.

Thank you.

### DIFF
--- a/THIRD-PARTY-LICENSES
+++ b/THIRD-PARTY-LICENSES
@@ -1,0 +1,25 @@
+The following component is licensed under the MIT License:
+
+Component: tapo
+
+MIT License
+
+Copyright (c) 2022-2025 Mihai Dinculescu
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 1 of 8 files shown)


Arksine:
Thanks.  Because the power module has become a bit unwieldy I am no longer accepting device specific implementations.   Most http devices are supported through the generic `http` implementation, and those that cannot be configured this way are generally supported by Home Assistant and/or MQTT.     I try to minimize 3rd party dependencies as in my experience library developers may stop supporting their project.  This can result in unfixed bugs or incompatibilities when new versions of Python are released.

That said, one option you have if you would like direct integration with Moonraker is to deliver the module as a unofficial 3rd party component, similar to [timelapse](https://github.com/mainsail-crew/moonraker-timelapse).  For example, you could name your component `tapo_switch.py`, and move your `PowerDevice` implementation to this module.  When this component is loaded it should first load the `power` component (`power = server.load_component(config, "power")`), then it may register tapo switch instances with the power module `power.add_device(name, dev_instance)`.

satoshi-ohba:
Thanks for the detailed explanation.

I reviewed the structure of both the [timelapse component](https://github.com/mainsail-crew/moonraker-timelapse) and the PrinterPower.add_device() implementation in [power.py](https://github.com/moonraker/component/power.py).
I understand now how third-party components can integrate with the built-in power module. I will investigate the loading mechanism for 3rd-party components and try implementing a tapo_switch component when I have time.

Thank you for your guidance.

satoshi-ohba:
Hi Arksine,

I originally thought I could integrate Tapo support with just a few lines — something like calling `power = load_component(config, "power")` and then `power.add_device(dev_name, dev_instance)`.
But once I actually tried it, I realized `add_device()` is `async`, so the process turned out to be trickier than I expected.

Still, I managed to get the power control working again on my setup, so I'll keep refining it a bit more.

Just wanted to share a quick update — no requests or anything.
Thanks as always.

PR #1048, opened 2026-01-21
Initial work to add OpenRC/Alpine Linux support.

RealDeuce:
Not sure how to test some of this stuff, but at least poweroff and reboot work, and it's not complaining.


RealDeuce:
There, services and poweroff/reboot now working with appropriate sudo "stuff".

PR #1049, opened 2026-01-27
metadata: add JusPrin slicer support

kennethjiang:
## Summary
- Add JusPrin to the PrusaSlicer-compatible slicer aliases for metadata extraction

JusPrin is a fork of OrcaSlicer that uses the same gcode metadata format. Adding it to the PrusaSlicer aliases allows proper metadata extraction for files sliced with JusPrin.

## Test plan
- Sliced a file with JusPrin 1.4.0
- Before: metadata fields (layer height, nozzle diameter, filament) showed `--`
- After: all metadata fields correctly populated

### DIFF
--- a/moonraker/components/file_manager/metadata.py
+++ b/moonraker/components/file_manager/metadata.py
@@ -369,6 +369,7 @@ class PrusaSlicer(BaseSlicer):
             'BambuStudio': r"BambuStudio[^ ]*\s(.*)\n",
             'A3dp-Slicer': r"A3dp-Slicer\s(.*)\son",
             'QIDISlicer': r"QIDISlicer\s(.*)\son",
+            'JusPrin': r"JusPrin\s(.*)\son",
         }
         for name, expr in aliases.items():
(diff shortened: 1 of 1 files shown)


kennethjiang:
Thank you @Arksine 🙏

Arksine:
Hi Kenneth, sorry about the delay.  I had some work on metadata.py I wanted to complete before responding, as it directly impacts this request.  Included in a recent series of patches is the ability for the metadata processor to detect all forks of PrusaSlicer that adhere to the same metadata format.  So as long as the identifier looks like the following the gcode file will be processed:

```
; generated by <slicer_name> <slicer_version> on <date (yyyy-mm-dd)> at <time (hh:mm:ss)>
```

For example: 

```
; generated by PrusaSlicer 2.7.0-rc1+win64 on 2023-11-19 at 21:45:58 UTC
```



If you could, pull the latest version of Moonraker and see if it will process gcode files generated by `JusPrin`.  If not I can add it to the fallback detection.  Thanks.

PR #1055, opened 2026-03-06
fix: remove password hash check for LDAP authentication

shxshxshxshx:
LDAP authentication currently fails if a user has changed their password since first login, Moonraker already validates the credentials by rebinding in `_perform_ldap_auth`.

### DIFF
--- a/moonraker/components/authorization.py
+++ b/moonraker/components/authorization.py
@@ -558,7 +558,7 @@ class Authorization:
             hashed_pass = hashlib.pbkdf2_hmac(
                 'sha256', password.encode(), salt, HASH_ITER).hex()
             action = "user_logged_in"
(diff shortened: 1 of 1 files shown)


Arksine:
Thanks for the report, I agree that Moonraker does not need to re-authenticate the password for ldap users, or even store a hashed version of the password locally.  The fix looks good, however it is likely that Moonraker will add additional authentication sources.  I pushed commit fb257f8ab80adec276f623077858208366b77b8f which refactors the `_login_jwt_method` user a bit and skips password authentication users registered with non-local sources.

shxshxshxshx:
Agreed, refactoring the flow to not save a unnecessary hash to begin with is the better solution.

PR #1059, opened 2026-03-17
spoolman: add per-tool spool tracking for multi-tool printers

### LINKED ISSUE #1057: Add per-tool spool tracking for multi-tool printers

Moonraker's Spoolman integration currently supports only a single active spool per printer instance. The active spool is stored as a single value (spoolman.spool_id in the database) and the API endpoints (POST /server/spoolman/spool_id, GET /server/spoolman/spool_id) have no concept of tool/extruder index.

This is a problem for multi-tool printers (tool changers, IDEX, and MMU/AFC setups) where multiple spools are loaded simultaneously — one  per tool. Today, the only workaround is for frontends or Klipper macros to call spoolman_set_active_spool on every tool change, swapping  the single global spool ID back and forth. This has several drawbacks:

  - Filament usage tracking can misattribute extrusion to the wrong spool during the window between a tool change and the spool swap call
  - Print history only records a flat list of spool IDs used, with no information about which tool used which spool
  - Pre-print validation cannot check that all tools have appropriate spools assigned
  - The recent MMU E-position watermark fix (commit b3f9566) mitigates accounting errors but doesn't solve the fundamental single-spool
  limitation

Multi-tool/multi-material printing is increasingly common (Happy Hare, AFC, ERCF, tool changers, IDEX), and proper per-tool spool tracking  would benefit a large portion of the community.

Extend Moonraker's Spoolman integration to support a per-tool spool mapping. Specifically:

1. Data model: Replace the single spool_id with a mapping of {tool_index: spool_id}, while maintaining backward compatibility (tool index  0 or a null key acts as the legacy single-spool behavior for single-extruder printers).
2. API endpoints: Extend the existing endpoints to accept an optional tool parameter:
    - POST /server/spoolman/spool_id — accepts {"spool_id": N, "tool": 0} (omitting tool preserves current behavior)
    - GET /server/spoolman/spool_id — accepts optional ?tool=0 query param; without it, returns the full mapping
    - Remote method: spoolman_set_active_spool(spool_id, tool=None)
3. Filament tracking: Track extrusion per-tool independently, reporting usage to the correct spool based on the currently active extruder (toolhead.extruder) and its mapped spool ID.
4. History integration: Record the tool-to-spool mapping in print history (e.g., {"spool_ids": {0: 5, 1: 12}}) instead of the current flat list.
5. Notifications: Include tool index in notify_active_spool_set events.

- Status quo (frontend-driven spool swapping): Mainsail currently works around this by storing spool IDs in Klipper gcode macro variables
  and calling SET_ACTIVE_SPOOL on tool changes. This works but is fragile, loses tracking accuracy, and puts the burden on every frontend
  independently.
- Klipper-side tracking: Having Klipper macros manage everything. This moves complexity into gcode macros and doesn't integrate with Spoolman's usage tracking or print history.
- Spoolman-server-side multi-spool: Having Spoolman itself support multiple active spools per printer. This would require changes to Spoolman's API as well, and Moonraker would still need to be the intermediary.

### Additional information

I'm willing to implement this and submit a PR. I'd appreciate feedback on the proposed API design before starting, particularly around backward compatibility. The corresponding Mainsail UI changes would be submitted as a separate coordinated PR to mainsail-crew/mainsail.

goeland86:
Replace the single `spool_id` with a tool-indexed map to support StealthChanger, IDEX, and MMU setups. Each tool gets independent spool assignment and filament usage tracking.

**Changes:**
- Replace `spool_id` attribute with `_tool_spool_map` dict + backward-compatible `spool_id` property (returns tool 0)
- Add toolchanger discovery on `klippy_ready` with extruder name fallback
- Per-tool `_highest_epos` watermarks for correct extrusion attribution across tool changes
- API endpoints (`GET/POST /server/spoolman/spool_id`) accept optional `tool` parameter (default 0)
- `GET /server/spoolman/status` response includes `tool_spool_map`
- `spoolman:active_spool_set` notification includes `tool` field
- History tracking via `tool_spool_map` auxiliary field
- Spool deletion iterates full tool map (clears all tools using deleted spool)
- DB migration from legacy single `spoolman.spool_id` key to `spoolman.tool_spool_map`; both keys maintained for rollback safety

**Backward compatibility:**
- All changes are opt-in via the `tool` parameter; existing single-tool setups work identically
- Legacy DB key continues to be written alongside the new format
- `spool_id` property returns tool 0's spool for code that reads it directly

**Tests:**
- 54 new unit tests covering: tool-spool map CRUD, backward-compat property, extrusion attribution across tool changes, extruder discovery parsing, spool deletion handling, GET/POST API endpoints, DB persistence with string key serialization, legacy migration, and historyreset callbacks — all for both single-tool and multi-tool scenarios

Fixes #1057

Signed-off-by: Jon Charnas <goeland86@gmail.com>

### DIFF
--- a/moonraker/components/spoolman.py
+++ b/moonraker/components/spoolman.py
@@ -532,10 +532,11 @@ class SpoolManager:
             spool_id = web_request.get_int("spool_id", None)
             tool = web_request.get_int("tool", 0)
             self.set_active_spool(spool_id, tool=tool)
-        else:
-            tool = web_request.get_int("tool", None)
-        # For GET requests (or after POST), return spool info
-        if tool is not None:
+            return {"spool_id": self._tool_spool_map.get(tool)}
+        # GET request
+        tool_val = web_request.get("tool", None)
+        if tool_val is not None:
+            tool = int(tool_val)
             return {"spool_id": self._tool_spool_map.get(tool)}
         # No tool specified: legacy response (tool 0)
         return {"spool_id": self.spool_id}
--- a/moonraker/components/spoolman.py
+++ b/moonraker/components/spoolman.py
@@ -35,6 +35,7 @@ if TYPE_CHECKING:
 
 DB_NAMESPACE = "moonraker"
 ACTIVE_SPOOL_KEY = "spoolman.spool_id"
+TOOL_SPOOL_MAP_KEY = "spoolman.tool_spool_map"
 
 class SpoolManager:
     def __init__(self, config: ConfigHelper):
@@ -50,17 +51,27 @@ class SpoolManager:
         self.ws_connected: bool = False
         self.reconnect_delay: float = 2.
         self.is_closing: bool = False
-        self.spool_id: Optional[int] = None
         self._error_logged: bool = False
-        self._highest_epos: float = 0
+        # Per-tool spool mapping: tool_index -> spool_id
+        self._tool_spool_map: Dict[int, Optional[int]] = {}
+        # Per-tool extrusion tracking watermarks
+        self._highest_epos: Dict[int, float] = {}
         self._last_epos: float = 0
         self._current_extruder: str = "extruder"
+        # Extruder name -> tool index mapping
+        self._extruder_to_tool: Dict[str, int] = {}
+        self._current_tool: int = 0
         self.spool_history = HistoryFieldData(
             "spool_ids", "spoolman", "Spool IDs used", "collect",
             reset_callback=self._on_history_reset
         )
+        self.tool_spool_history = HistoryFieldData(
+            "tool_spool_map", "spoolman", "Tool-spool assignments", "collect",
+            reset_callback=self._on_tool_spool_history_reset
+        )
         history: History = self.server.lookup_component("history")
         history.register_auxiliary_field(self.spool_history)
+        history.register_auxiliary_field(self.tool_spool_history)
         self.klippy_apis: APIComp = self.server.lookup_component("klippy_apis")
         self.http_client: HttpClient = self.server.lookup_component("http_client")
         self.database: MoonrakerDatabase = self.server.lookup_component("database")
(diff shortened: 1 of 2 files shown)


Arksine:
Thanks.  Before I get into a detailed code review there are a couple of high level issues that we need to work out.

#### Maintaining compatibility with existing workflows

It appears to me that this could potentially break existing flows for filament changes in multi-extruder setups.  If the user issues `set_active_spool(<spool_id>)` to change the current spool ID then only tool 0 will ever be changed, resulting in incorrect tracking for any other extruder.  That said, I believe we could overcome this by setting all tracked spools to the requested spool ID when the tool is omitted rather than just tool 0.

#### Moonraker can't depend on APIs or printer objects that are not included in the official repo

This may be the bigger issue.  We can't refer to objects or APIs on forks and/or modifications to Klipper.  The potential for breakage and/or maintenance issues is too high.  I can't reliably be sure that the modification will be maintained in sync with the current Klipper repo and I can't be sure that Klipper won't at some point in the future add a printer object with the same name.  I also don't want to open up the can of worms that would exist when authors of other mods want to integrate directly with Moonraker.

If its possible to rewrite this with only dependencies on current hardcoded Klipper objects I will be open to the changes.  Otherwise this could be revisited if MMU/Toolchanger support is merged in the official repo.

Edit:  In addition, I just noticed that this PR includes AI generated code.  I'm not sure yet what my policy will be on this, as we just started having these discussions.  I may limit it to tests, analysis, and docs only.   I may open it up to Moonraker's source with the caveat that only prior contributors that have demonstrated a working knowledge of the code base can depend on generative AI tools.   There are potential licensing issues with AI generated code that need to be discussed, and ultimately I would like Moonraker (and Katapult) to have a policy that is consistent with Klipper.

goeland86:
So I thought I'd double checked single tool use-case, but I hadn't looked in detail at a multiplexing system. My main goal was for tool changer/IDEX support. So I'll see what tests I can run since I don't have an AMS-like system yet.

Regarding the unofficial printer objects, I'll try to refactor it as best I can. Is there a way to create moonraker add-ons that aren't independent services? If it's better to rework it in that way I'm happy to do so if refactoring proves impractical.

Regarding AI-generated code, I fully understand your concerns. I write C++ for a living, and I've worked a long time on the Recore's embedded image project as well. Grappling with licensing is something I'm quite familiar with, and I won't push you one way or another on that stance, but what I will say is that it's a lot less of a legal hurdle for open source projects than closed-source. It becomes primarily a matter of attribution, which the LLM in question may be able to properly trace. 

If the code generated passes muster, both in real-world and unit-test cases, I see value in the output. I have been checking and re-directing Claude on a lot of points, regarding functionality as well as architecture, even if I'm not the most competent python programmer.

PR #1065, opened 2026-03-25
feat: added Odrive S1 BLDC motor control support via can bus

arthurtumanyan:
Added support to control Odrive S1 motor controller via REST API . Motor controller should be connected to Klipper supported device  via CAN bus. 

Successfully tested on manta-m8p device.
Example of request:
 curl -X POST http://<moonraker_ip>:7125/machine/odrive/command   -H "Content-Type: application/json"   -d '{"action": "H0", "distance": 0}'
 curl -X POST http://moonraker_ip:7125/machine/odrive/command   -H "Content-Type: application/json"   -d '{"action": "G0", "distance": 100}'
 curl -X POST http://moonraker_ip:7125/machine/odrive/command   -H "Content-Type: application/json"   -d '{"action": "RESET"}'

### DIFF
--- a/moonraker/components/odrive_can.py
+++ b/moonraker/components/odrive_can.py
@@ -1,0 +1,957 @@
+"""
+odrive_can.py  -  Moonraker Component for ODrive S1 CAN Bus Control
+Mission-critical hardened version.
+
+Place at: ~/moonraker/moonraker/components/odrive_can.py
+
+Enable in moonraker.conf:
+  [odrive_can]
+  can_interface: can0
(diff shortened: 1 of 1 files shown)


Arksine:
Thanks.  One of the design principles of Moonraker is that direct robot control is strictly in Klipper's domain and Moonraker should not encroach on it.  This would apply to steppers, servos, heaters, and various sensors.  From Klipper it is possible to register endpoints that are proxied via Moonraker that expose direct control if necessary, although it would likely be more efficient to simply register GCode Commands.

PR #1070, opened 2026-04-22
scripts: add opensuse installation support

ldevulder:
Automate installation for openSUSE Linux based systems. Supports Tumbleweed, Slowroll and Leap 16+

Tested on my modified Sidewinder X1 with an arm64 board installed with openSUSE Tumbleweed on it.

### DIFF
--- a/scripts/install-moonraker.sh
+++ b/scripts/install-moonraker.sh
@@ -242,6 +242,21 @@ system_deps = {
         "python3-libcamera; vendor == 'raspberry-pi' and distro_version >= '11'",
         "curl", "build-essential"
     ],
+    "opensuse-leap": [
+        "python3-virtualenv", "python3-devel", "libopenjp2-7", "libsodium-devel",
+        "zlib-ng-compat-devel", "libjpeg8-devel", "PackageKit", "wireless-tools",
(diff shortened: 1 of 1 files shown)


scripts/install-moonraker.sh
```diff
@@ -242,6 +242,21 @@ system_deps = {
...
+    "opensuse-slowroll": [
+        "python3-virtualenv", "python3-devel", "libopenjp2-7", "libsodium-devel",
+        "zlib-ng-compat-devel", "libjpeg8-devel", "PackageKit", "wireless-tools",
+        "python3-libcamera", "curl", "unzip", "libsodium26",
+    ],
+    "opensuse-tumbleweed": [
+        "python3-virtualenv", "python3-devel", "libopenjp2-7", "libsodium-devel",
+        "zlib-ng-compat-devel", "libjpeg8-devel", "PackageKit", "wireless-tools",
+        "python3-libcamera", "curl", "unzip", "libsodium26",
+    ],
```
ldevulder, line 259: For now all openSUSE variants have the same declared packages, but this will not be always the case, as Tumbleweed ans Slowroll evolve more quickly,

PR #1078, opened 2026-06-08
git_deploy: handle "(no branch)" detached HEAD and tag-pinned repos

lexfrei:
## Problem

Pinning a managed `git_repo` (for example Klipper) to a release tag leaves the working copy in a detached HEAD. git renders that state as either `(HEAD detached at <ref>)` or `(no branch)`. The update manager only recognised the first form; `(no branch)` was treated as a literal branch name, so `_find_current_branch` ran `git config branch.(no branch).remote` and got `error: invalid key: branch.(no`. The tracking remote was lost and the whole refresh failed, leaving the repo reported as fully invalid with every field `?`, even though `git remote get-url origin` and `git describe --tags` both worked.

## What this changes

Two independent commits plus a docs note.

1. `git_deploy: handle "(no branch)" detached HEAD` — treat any current branch starting with `(` as detached so the invalid `branch.<name>.remote` lookup is never built. When the detached ref encodes no remote (a bare tag or commit), fall back to `origin`, or the sole remote if there is exactly one, so the repo url, fetch and tag-based version detection still resolve. This alone turns the broken state into the normal detached-HEAD state the update manager already supports.

2. `git_deploy: treat tag-pinned detached HEAD as valid for stable/beta` — the `stable` and `beta` channels track release tags, so a checkout sitting detached exactly on a qualifying tag is their natural state. `has_recoverable_errors()` previously flagged every detached HEAD as an error, which forced `is_valid()` to false and disabled updates. It now accepts a detached HEAD that is exactly on a tag matching the channel (a final release for `stable`, a non-alpha tag for `beta`), reusing the same tag-eligibility predicates as the upstream-version filter. The detached update path is also corrected: `checkout()` assumed a tracking branch and built `<remote>/<branch>`, which is wrong when the branch is unknown or stale after pinning; it now checks out the resolved target commit directly and raises a clear error instead of running `git checkout ?` when no target resolves.

## Design note

The `stable`/`beta` model is "stay on the primary branch, but reset the branch pointer to the latest qualifying tag commit". A genuinely detached checkout is a less common shape, so the two commits are kept independent: the first is a self-contained crash fix that does not depend on the second, and the second makes the detached-on-tag shape a first-class valid state for the tag-tracking channels.

## Testing

Verified:

- Added `tests/test_git_deploy.py` covering branch parsing (normal branch, `(HEAD detached at origin/master)`, detached on a tag, `(no branch)`, sole-remote and ambiguous-remote fallback, stale tracking after pinning), `has_recoverable_errors` (qualifying `stable`/`beta` tags are valid; `dev_count > 0`, dirty, diverged, alpha-for-beta and the `dev` channel stay recoverable), and `checkout()` target resolution. The tests are red against the pre-fix code and green after.
- flake8 6.1.0 and mypy 1.5.1 pass on the changed source with the project's CI options.

Reproduction for end-to-end confirmation:

- Build a tag-only clone that renders `* (no branch)` (`git clone --no-checkout`, set `remote.origin.fetch` to a single `+refs/tags/<tag>:refs/tags/<tag>`, `git checkout <tag-commit>`), configure an `[update_manager]` entry with `channel: stable`, and call `GET /machine/update/status?refresh=true`. Before the fix the entry is invalid with `git_messages` containing `error: invalid key: branch.(no`; after the fix the entry is valid with a populated `version` and `remote_url`.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -2179,7 +2179,10 @@ type: git_repo
 #   This parameter must be provided.
 channel: dev
 #   The update channel.  May be set to stable, beta, or dev.
-#   The default is dev.
+#   The default is dev.  A detached HEAD, such as a checkout pinned
+#   directly to a tag, is reported as not updatable; return the repo to a
+#   branch (attached) to update it.  The exception is debug mode, where a
+#   detached HEAD can update when its ref is a remote branch.
 refresh_interval:
 #   This overrides the refresh_interval set in the primary [update_manager]
 #   section.


Arksine:
Thanks.  Beta and stable updates necessarily require that the repo be in a valid attached state.  The way stable/beta updates work is to stay on the current branch and pull to the appropriate tagged commit.  The update manager would generate a pull command like the following:

```
git pull <remote> <tagged commit hash>
```

To be more specific, we must know what the current branch the repo is on to reliably determine which tags are reachable from the current commit.  Without that information we might attempt to update to a tag on a completely different remote and/or branch from what we are on.  One thing the update manager tries to stay away from is inferences.  It inspects the current git repo, compares it to the configuration, warns the user of discrepancies, and otherwise allows updates.  It won't use the configuration to infer the remote, branch, etc if that data is not detected.
 The UM *will* restore some previously detected state when detached and it will allow fallbacks to a previously valid state, both of which are generally safe. 

Detached update support is specifically intended for developers and testers.  It updates from the current head to the remote head and nothing else.  In the situation you present it appears that repo was cloned, then a `git checkout` was performed to move to the desired tag. The correct procedure would be to clone the repo, reset the current branch to the desired tag, then allow the update manager to update.  It may be worth documenting that beta and stable updates can only be performed in an attached state.

As a side note, I'm currently debating a Generative AI policy, but I'm leaning toward something similar to the [proposed revision]( https://www.phoronix.com/news/QEMU-Patch-Allows-Some-AI) of QEMU's current AI policy.  That is, there won't be a complete ban on AI tools, but acceptance would be limited to small impact items like tests, docs, etc.

lexfrei:
Fair enough — I won't argue the update mechanism; you're right that resolving reachable tags needs the attached branch, so I'll drop the auto-update-while-detached part.

Two things are still worth separating from the update policy, though:

`(no branch)` isn't "no update", it's a parse bug. It's treated as a literal branch name, so `_find_current_branch` runs `git config branch.(no branch).remote` and git errors with `invalid key: branch.(no`. That wipes detection — `remote_url`/`version`/`owner` all go `?` and the error lands in `git_messages` — and it means your "restore previously-detected state when detached" path never gets a chance to run. Treating `(no branch)` as detached, like `(HEAD detached at ...)`, is just correctness, no inference.

Separately: a clean checkout on a release tag is a valid state I picked on purpose (tag-only clone, to save space). I'm fine with it not auto-updating, but reporting it as *invalid* is a false positive. Could detection/status be decoupled from updatability — shown as pinned/manual rather than invalid?

Happy to cut this PR down to the parse fix plus the doc note you suggested, and handle the decoupling separately. And noted on the AI policy: this was AI-assisted, so I'm fine keeping it to the minimal fix + docs.

Arksine:
> (no branch) isn't "no update", it's a parse bug. It's treated as a literal branch name, so _find_current_branch runs git config branch.(no branch).remote and git errors with invalid key: branch.(no. That wipes detection — remote_url/version/owner all go ? and the error lands in git_messages — and it means your "restore previously-detected state when detached" path never gets a chance to run. Treating (no branch) as detached, like (HEAD detached at ...), is just correctness, no inference.

I agree, `(no branch)` should be handled as detached.  That said, unless I am mistaken `(no branch)` can indicate that the repo is in a state where it can't be updated, such as during a rebase in progress.  I'll need to do some additional research, but for the time being I am ok with a fix that sets the head as detached and applies the previous branch and remote if they are available.  I do not think its correct to infer the remote if no previously stored remote is available.

> Separately: a clean checkout on a release tag is a valid state I picked on purpose (tag-only clone, to save space). I'm fine with it not auto-updating, but reporting it as invalid is a false positive. Could detection/status be decoupled from updatability — shown as pinned/manual rather than invalid?

When the update manager reports an item as valid it is specifically referring its ability to perform an update (this applies to all AppDeploy subclasses, not just git repos).  So while from git's perspective checking out a tag (or any ref) is certainly valid, from the update manager's perspective it is not unless the named ref is a branch.  Similarly, a dirty repo is "valid" as far as git is concerned, but it can't be safely updated due to potential conflicts.  Front ends use the `is_valid` value to determine whether or not an item can be updated, so we can't decouple git repo state from the ability to update without breaking the API.

There are a few alternatives:
1) I presume the primary focus here is on Klipper and/or Moonraker, as any other repo could just go unconfigured.  It would be possible to add an option to disable one or both so their status won't be reported by the update manager.  That said, this could potentially break front ends, as to date they could reliably be sure that both will report some state.  Since the default would be to enable status reporting this isn't a blocker, just something front end devs have to deal with.

2) We could report an additional field such as `git_repo_detected` which reflects the repo's state.

3)  We could add the `info_tags` option to the list of overrides for Moonraker and Klipper then allow frontends to use a tag to identify a repo as "manual" or something similar.

Options 2 and 3 would require front end developers to implement solutions, whereas option 1 might work with some or all front ends depending on how they implement update management.

lexfrei:
Done — pushed a trimmed version: `(no branch)` is now handled as detached, keeping the previously-tracked branch/remote when stored and not inferring one otherwise, plus a doc note that a detached repo is reported as not updatable. Dropped the validity change.

On status: I'd still want the repo visible in the update manager — I rely on seeing the current hash/version at a glance — so disabling reporting (option 1) isn't it. Any fix to this surface needs front-end changes regardless, so "needs front-end work" isn't really a downside unique to the other options. The backend just has to stay backward-compatible — existing fields and `is_valid` semantics untouched — while exposing new data that front ends can adopt in a future release. That points to option 2: an additive field like `git_repo_detected`. Glad to do it as a separate PR.

Arksine:
Thanks.  I merged the first commit that fixes "no branch" detection.  I held the second back, as the additional documentation isn't 100% accurate.  When debug mode is enabled Moonraker can update a detached head if the named ref is a remote branch.

Its no problem to add a `git_repo_detected` field, however I am considering instead flag instead.  I'll give it a bit of thought and make that addition soon.

lexfrei:
Thanks for merging the parse fix. I updated the doc note to cover the debug-mode exception — a detached HEAD can update when its ref is a remote branch — so it should be accurate now. On the field vs flag: happy to implement whichever shape you'd prefer as a separate PR — just say which and I'll put it together.

PR #1079, opened 2026-06-16
spoolman: support HTTP Basic Auth credentials embedded in server URL

TyanColte:
## Summary
- Allows `[spoolman]`'s `server` option to be specified as `scheme://user:pass@host`
- Credentials are stripped from the URL before use and applied as HTTP Basic Auth to both the websocket connection and the HTTP client requests (get spool, use spool, generic proxy request)

## Motivation
Useful for deployments where Spoolman sits behind a reverse-proxy/auth layer that requires HTTP Basic Auth for non-browser/API clients (e.g. a forward-auth gateway with a separate Basic-Auth bypass route for service accounts), while browser traffic to the same host goes through a different auth flow.

## Test plan
- [x] Running in production against a Spoolman instance behind such a proxy; websocket connects and spool sync/use-length updates succeed with credentials in the URL
- [x] Verified `[spoolman]` `server` without embedded credentials still works unchanged (no behavior change when no credentials present)

### DIFF
--- a/moonraker/components/spoolman.py
+++ b/moonraker/components/spoolman.py
@@ -9,8 +9,10 @@ import asyncio
 import logging
 import re
 import contextlib
+import urllib.parse
 import tornado.websocket as tornado_ws
 from tornado import version_info as tornado_version
+from tornado.httpclient import HTTPRequest as TornadoHTTPRequest
 from ..common import RequestType, HistoryFieldData
 from ..utils import json_wrapper as jsonw
 from typing import (
@@ -83,6 +85,20 @@ class SpoolManager:
         scheme = url_match["scheme"] or "http"
         host = url_match["host"].rstrip("/")
         ws_scheme = "wss" if scheme == "https" else "ws"
+        # Extract Basic Auth credentials from URL if present (e.g. user:pass@host)
+        parsed = urllib.parse.urlsplit(f"{scheme}://{host}")
+        self._basic_auth_user: Optional[str] = parsed.username or None
+        self._basic_auth_pass: Optional[str] = parsed.password or None
+        if self._basic_auth_user:
+            # Rebuild host without credentials
+            clean_host = parsed.hostname
+            if parsed.port:
+                clean_host = f"{clean_host}:{parsed.port}"
+            host = clean_host
+        logging.info(
+            f"Spoolman basic_auth_user={self._basic_auth_user!r}, "
+            f"host={host}"
+        )
         self.spoolman_url = f"{scheme}://{host}/api"
         self.ws_url = f"{ws_scheme}://{host}/api/v1/spool"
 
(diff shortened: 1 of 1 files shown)


TyanColte:
Any reason why this didn't commit?

Arksine:
> Any reason why this didn't commit?

I apologize, until now I have not had an opportunity to review the PR.  I have done so and will provide feedback.  To start, I would recommend reviewing Moonraker's [contribution guidelines](http://moonraker.readthedocs.io/en/latest/contributing).  Commits must be signed off to confirm acceptance of the DCO.  In addition, I'm am unsure of the impact this change would have.  Generally speaking I wouldn't expect Spoolman to be installed behind a proxy separate from Moonraker.  In addition, if authentication is desirable I think it would be better to implement it directly in to Spoolman rather then rely on Basic Auth behind a proxy.

All of the above aside, the contribution itself appears well implemented.  I do think it would make more sense to use configuration options for the username and password, we don't really need to put credentials in the URL since we have that option.  When doing so we should use the template pattern so credentials can be stored in the secrets file if users desire to keep them out of their configs/logs.

TyanColte:
> > Any reason why this didn't commit?
> 
> I apologize, until now I have not had an opportunity to review the PR. I have done so and will provide feedback. To start, I would recommend reviewing Moonraker's [contribution guidelines](http://moonraker.readthedocs.io/en/latest/contributing). Commits must be signed off to confirm acceptance of the DCO. In addition, I'm am unsure of the impact this change would have. Generally speaking I wouldn't expect Spoolman to be installed behind a proxy separate from Moonraker. In addition, if authentication is desirable I think it would be better to implement it directly in to Spoolman rather then rely on Basic Auth behind a proxy.
> 
> All of the above aside, the contribution itself appears well implemented. I do think it would make more sense to use configuration options for the username and password, we don't really need to put credentials in the URL since we have that option. When doing so we should use the template pattern so credentials can be stored in the secrets file if users desire to keep them out of their configs/logs.

No worries, I appreciate your reply and review. This PR was submitted to fix an issue with the way Moonraker's spoolman python script handles the headers for spoolman when behind an authentication wall.

Prior to this PR, the spoolman.py and http.py stripped out any extraneous headers form the spoolman url. For users who run their spoolman behind a reverse proxy (like myself) and configure moonraker to connect to the external URL instead of the internal IP:Port structure. This allows users' to inject basic auth credentials via their reverse proxy manager (Nginx, Caddy, etc) to secure access to their spoolman instance. (Since no auth support from spoolman devs is forthcoming) and configure Moonraker to read the basic auth url and pass those credentials through to spoolman in the form of `scheme://<user>:<password>@<spoolman-url>`.

I will say, since this PR was submitted, I have moved to a different authentication system (Authentik) and all my publically accessible URLs are only available behind an Authentik auth wall. However, moonraker still connects to spoolman using basic auth credentials that are set up in my authentik configuration for spoolman for internal connections. Without this change. moonraker would not be able to connect to spoolman behind my auth wall and I would be unable to see spools in Fluidd, Mainsail, KlipperScreen, or OctoApp/MobileRaker on my Android device.

I have been running this repo fork on my own system (Heavily customized Longer LK5 Pro, Kalico, Moonraker, Fluidd (Docker), Mainsail (Docker), KlipperScreen (Running via wireless ADB on an old Pixel 6a using Xserver-XSDL), Obico (For visual ML print monitoring), OctoApp (Android), and MobileRaker (Android) as a dirty repo since the PR was submitted with zero issues. Hopefully this can give you the confidence to merge the PR to the public repo.

I appreciate your consideration.

EDIT: I want to make it clear, this does not change the existing functionality of the connection between spoolman and monraker. Normal users who call to spoolman locally via IP:PORT will not be affected in any way. It only allows users who use a reverse proxy to access their spoolman instance remotely to include basic auth credentials to either mitigate unauthorized access. Or to create a bypass layer around a more robust auth solution for moonraker to talk to spoolman's api.

EDIT 2: Fixing typos and weird wording.

Issue #893, opened 2024-08-03
Metadata extract fails with 'Input data was truncated' some of the time.


Cellivar:
### What happened

Most of the time, but not _all_ of the time, Moonraker fails to extract metadata from gcode files. The file prints correctly.

The logs show [this line loading the json](https://github.com/Arksine/moonraker/blob/b40751ba9d7f05e7da81636e3e4a5735ef85bb27/moonraker/components/file_manager/file_manager.py#L2562) as the problem. It doesn't look like there's a lot of logging right around there so I'm uncertain if more verbose logging will help.

This is the error I see in my logs related to metadata parsing:

```
[shell_command.py:pipe_data_received()] - INFO:metadata:Object Processing is enabled

[shell_command.py:pipe_data_received()] - INFO:metadata:File 'test_file.gcode' currently supports cancellation, processing aborted

[shell_command.py:_check_proc_success()] - Command (/home/moonraker/venv/bin/python /home/moonraker/moonraker/moonraker/components/file_manager/metadata.py -p /home/moonraker/printer_data/gcodes -f "test_file.gcode" --check-objects) successfully finished

[file_manager.py:_process_metadata_update()] - Error running extract_metadata.py
Traceback (most recent call last):

  File "/home/moonraker/moonraker/moonraker/components/file_manager/file_manager.py", line 2520, in _process_metadata_update
    await self._run_extract_metadata(fname, ufp_path)

  File "/home/moonraker/moonraker/moonraker/components/file_manager/file_manager.py", line 2564, in _run_extract_metadata
    decoded_resp: Dict[str, Any] = jsonw.loads(result.strip())
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^
msgspec.DecodeError: Input data was truncated
```
<details>
  <summary>Additional context logs, though not particularly helpful</summary>
  
```
[application.py:prepare()] - Upload Request Received from 192.168.1.44
Content-Type: 
[application.py:prepare()] - Upload Request Received from 192.168.1.44
Content-Type: multipart/form-data; boundary=----WebKitFormBoundaryFov2ibiAVBTX5PEA
[application.py:post()] - Processing Uploaded File: test_file.gcode
[shell_command.py:pipe_data_received()] - INFO:metadata:Object Processing is enabled

[shell_command.py:pipe_data_received()] - INFO:metadata:File 'test_file.gcode' currently supports cancellation, processing aborted

[...]

Issue #795, opened 2024-01-21
Loading gcode files causes fluidd to lose connection to moonraker


samwiseg0:
### What happened

On the latest version it appears there is some timeout that is triggered that causes fluidd to lose communication to moonraker. 

This happens when trying to load the list of gcode files. When opening a folder where the gocde files are the preview image may load on a few gcode files and then the interface hangs and becomes unresponsive. 

To test I rolled back to commit, https://github.com/Arksine/moonraker/commit/6d8cb762ffca1bd151d98e0cec5b66b966965014 and everything works as expected. 

### Client

Fluidd

### Browser

Chrome, Safari, Other or N/A

### How to reproduce

Run the latest version. I am using authz but I am unsure if that is related to this issue. Navigate to jobs and open a folder containing gcode files and the interface will become unresponsive. 

### Additional information

[moonraker-5.log](https://github.com/Arksine/moonraker/files/14002257/moonraker-5.log)

Seeing this repeatedly in the logs. 
```
2024-01-21 12:08:31,111 [proc_stats.py:_watchdog_callback()] - EVENT LOOP BLOCKED: 20.31 seconds, total blocked count: 21
```

Arksine:
Thanks.  I'm working on tracking this down but thus far have been unable to reproduce.  Can you add the following line to `~/printer_data/systemd/moonraker.env`?

```
MOONRAKER_VERBOSE_LOGGING="y"
```

Then restart Moonraker, repro the issue, and upload the log.  This will create large logs, but will hopefully help me track down where the blocking issue is occurring (and how its related to authorization, it really shouldn't be).

Arksine:
Update:

Even though I can't reproduce I think I have found the source of the problem.  There is a potential blocking call to `socket.getfqdn` in the `authorization` module that was generally bypassed previously.  Its a call that is very rarely reached and normally it returns immediately.  I'll work on a fix.  You can hold off on the verbose logging right now, as I am reasonably confident that this is the problem.

samwiseg0:
> Update:
> 
> 
> 
> Even though I can't reproduce I think I have found the source of the problem.  There is a potential blocking call to `socket.getfqdn` in the `authorization` module that was generally bypassed previously.  Its a call that is very rarely reached and normally it returns immediately.  I'll work on a fix.  You can hold off on the verbose logging right now, as I am reasonably confident that this is the problem.

[...]

PR #1096, opened 2026-08-19
file_manager: performance improvements

zackgomez:
File listings run synchronously on the event loop. On SBC-class hardware with a
large gcodes directory each listing blocks the loop for several seconds,
stalling every other request — the `EVENT LOOP BLOCKED` symptom in #795, also
visible in the logs of #893 — and the same per-file work runs in `scan_node()`
at startup, before the server begins listening (~10s on the machine below).

Two costs stack in the per-file work: 5–7 path-resolution syscalls per file
where a single `stat` suffices, and `get_path_info` walking every reserved
path's ancestor chain for every file — O(reserved paths × files) pure Python,
about 2ms/file on a CB1 with 16 reserved paths.

The commits, in order:

1. **fix symlink loop detection in scan_node** — the visited check compared an
   `os.stat_result` against a set of `(st_dev, st_ino)` tuples, so it never
   matched and a directory-symlink cycle recursed to the recursion limit.
2. **run file list requests in a thread** — `_list_directory` and
   `get_file_list` are pure filesystem/stat work; run them via
   `event_loop.run_in_thread()`, the pattern this module already uses for its
   `shutil` operations. (A Moonraker-initiated delete can now interleave with an
   in-flight listing request; listings already raced external writers.)
3. **reduce per-file syscalls when building file lists** — iterate with
   `os.scandir()`, deriving file type from cached `d_type`/stat data with a
   single `stat` per file.
4. **compute directory containment once per listing** — resolve the parent
   directory's reserved-path containment once per listing; plain entries share
   it, symlinked entries keep the full per-entry ancestor walk. Listing output
   is unchanged.

Results on a BTT CB1 (Allwinner H616, SD storage, 874-file gcodes dir,
16 reserved paths, warm):

| Metric | before | after |
|---|---|---|
| `GET /server/files/directory` | 2.77s | 0.073s |
| `GET /server/files/directory?extended=true` | 3.5s | 0.68s |
| `GET /server/files/list` (gcodes) | 2.62s | 0.125s |
| Startup event-loop block (`scan_node`, pre-listen) | 8.9–11.8s | 0.37s |
| `/server/info` during a listing | queued behind it | ~10ms |
| `proc_stats` EVENT LOOP BLOCKED events | constant | none |

Testing: the containment change was equivalence-tested against `get_path_info`
(symlinks including links into reserved trees, broken links, reserved children,
the reserved root itself, mixed read permissions) with byte-identical output;
all listing endpoints were captured and diffed before/after each commit on two
live printers (the CB1 above and a Pi 4), identical; both printers run this
branch in daily use. `flake8==6.1.0` and `mypy==1.5.1` pass with the CI flags.


thijstriemstra:
+1, these results look awesome, any thoughts on these changes @Arksine?

zackgomez:
Yeah I mean one might say "what kind of idiot stores 874 loose gcodes" but yeah that's me.  I like having the full history and I don't want to organize this folder.

If you don't want the whole stack (the directory containment one deserves scrutiny).  the first 3 give a performance improvement, but the full stack takes my printer from a slog to snappy.  I have a similar KlipperScreen print screen improvement in the pipe when this is landed.

